PR #2, opened 2023-09-07
chore: Copied sofie-server code from core repository.

KvelaGorrrrnio:
Taken from commit c2b17cfd0ce7b3d59a494d5604fbbe8e326aea14.


LindvedKrvang (approved):

RasmusAlbrektsen (dismissed):
Since my changes were the most recent ones, i have looked through and approved on the basis that i found these changes and that the code base is as recent as it should be.

PR #8, opened 2023-09-19
Adds tests for repository get methods.

andr9528:
Adds tests to ensure the get methods of the 4 main repositories works as intended.
Also add `EntityFactory` for use to create entities for test.

Sof-1527


src/model/entities/test/entity-factory.ts
```diff
@@ -0,0 +1,52 @@
+import { Rundown, RundownInterface } from '../rundown'
+import { Segment, SegmentInterface } from '../segment'
+import { Part, PartInterface } from '../part'
+import { Piece, PieceInterface } from '../piece'
+
+export class EntityFactory {
```
LindvedKrvang, line 6: Is this only meant for testing purporses? (that fact you set ids to `Math.random()` indicates that).

If yes, it would be nice if the name reflected that so a developer won't accidentally use this for produciton code.

LindvedKrvang (commented):
We have talked about your reintroduction of `TesteeParams` and `populateAndCreateTestee()` so I'm not writing those comments. Let me know if you want me to write them anyway :)

src/data-access/repositories/test/mongo-part-repository.spec.ts
```diff
@@ -262,20 +263,118 @@ describe(`${MongoPartRepository.name}`, () => {
...
+      const testee: PartRepository = createTestee({
+        mongoConverter: mongoConverter,
+      })
+
+      await testee.getParts(nonExistingId)
+
+      const [capturedMongoParts] =  capture(mongoConverter.convertParts).first()
+      expect(capturedMongoParts.length).toBe(0)
+      expect((await db.collection(COLLECTION_NAME).find().toArray()).length).toBe(1)
+    })
```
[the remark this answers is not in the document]
andr9528: As quickly mention before lunch, by changing the test to the suggested, including the change to `createTestee()`, the test should never fail, as the expect is checking the length of the mocked return value.

Propably need a longer chat about this one.

src/data-access/repositories/test/mongo-rundown-repository.spec.ts
```diff
@@ -183,23 +185,101 @@ describe(`${MongoRundownRepository.name}`, () => {
...
+  describe(`${MongoRundownRepository.prototype.getRundown.name}`, () => {
+    it('throws exception, when nonexistent rundownId is given', async () => {
+      const nonExistingId: string = 'nonExistingId'
+      const mongoRundown: MongoRundown = createMongoRundown()
+      await testDatabase.populateDatabaseWithInactiveRundowns([mongoRundown])
+
+      const testee: MongoRundownRepository = createTestee({})
+      const action: () => Promise<Rundown> = async () => testee.getRundown(nonExistingId)
+
+      await expect(action).rejects.toThrow(NotFoundException)
```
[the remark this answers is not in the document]
andr9528: Have updated a few such lines as suggested.

src/data-access/repositories/test/mongo-rundown-repository.spec.ts
```diff
@@ -183,23 +185,101 @@ describe(`${MongoRundownRepository.name}`, () => {
...
+    it('returns rundown, when existing rundownId is given', async () => {
+      const rundown: Rundown = EntityFactory.createRundown()
+      const testee: RundownRepository = await populateAndCreateTestee({}, {rundown: rundown})
+
+      const result: Rundown = await testee.getRundown(rundown.id)
+
+      expect(result).not.toBeNull()
+    })
+
+    it('converts from mongo rundown to our rundown entity, when existing rundownId is given', async () => {
```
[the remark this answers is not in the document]
andr9528: Tests like the one commented on are gone.

src/data-access/repositories/test/mongo-part-repository.spec.ts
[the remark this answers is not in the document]
LindvedKrvang: Your implementation fails if we change the implementation, but not the functionality. I.e. if we decided to no longer use the MongoEntityConverter and just instantiate the entities directly, then this test would fail even though it still returns zero entities on a non-existing id.

And as long as the correct functionality is there, the test shouldn't fail.

src/data-access/repositories/test/mongo-rundown-repository.spec.ts
```diff
@@ -70,9 +72,8 @@ describe(`${MongoRundownRepository.name}`, () => {
       const db: Db = testDatabase.getDatabase()
 
       const testee: MongoRundownRepository = createTestee({})
-      const action: () => Promise<void> = async () => testee.deleteRundown(nonExistingId)
 
-      await expect(action).rejects.toThrow(NotFoundException)
+      await expect(() => testee.deleteRundown(nonExistingId)).rejects.toThrow(NotFoundException)
```
KvelaGorrrrnio, line 76: Is it necessary to wrap this in a function?
If it returns a promise, that should be sufficient to pass to expect.

src/model/entities/test/test-entity-factory.ts
```diff
@@ -0,0 +1,52 @@
+import { Rundown, RundownInterface } from '../rundown'
+import { Segment, SegmentInterface } from '../segment'
+import { Part, PartInterface } from '../part'
+import { Piece, PieceInterface } from '../piece'
+
+export class TestEntityFactory {
+  public static createRundown(rundownInterface?: Partial<RundownInterface>): Rundown {
+    if (!rundownInterface) {
+      rundownInterface = {} as RundownInterface
+    }
```
KvelaGorrrrnio, line 10: You can get around this by changing the parameter to having a default value:
```typescript
  public static createRundown(rundownInterface: Partial<RundownInterface> = {}): Rundown {
```

src/model/entities/test/test-entity-factory.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    return new Rundown({
+      id: rundownInterface.id ?? 'rundownId' + Math.floor(Math.random()*1000),
+      name: rundownInterface.name ?? 'rundownName',
+      segments: rundownInterface.segments ?? []
+    } as RundownInterface)
```
KvelaGorrrrnio, line 16: Isn't the purpose of this factory to fill in default values for non-provided values for the `XInterface` objects?

andr9528: Yes, but i chose to only add for the properties i make use at the moment, planning for people to add default values as they need them. Don't know if that is the correct approch to take, but at the same time having a value for all properties doesn't matter, if no test makes use of it.

src/model/entities/test/test-entity-factory.ts
andr9528: Updated

src/data-access/repositories/test/mongo-rundown-repository.spec.ts
andr9528: TIL

They do indead not need to be wrapped in a function. 
Don't know if that is because it returns a promise, or if that is just a general thing for expecting to throw.

src/data-access/repositories/test/mongo-part-repository.spec.ts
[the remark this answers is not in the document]
andr9528: Since my last comment, i have also come to a similar conclusion.

I will be updating the 3 test wiith the signature as the one commented on, to the suggested signature.
Will also add a comment stating that while we have and mock the `MongoEntityConverter`, the tests should never fail.

src/data-access/repositories/test/mongo-rundown-repository.spec.ts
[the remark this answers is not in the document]
andr9528: Has been updated further after one of Anders' comments.

src/data-access/repositories/test/mongo-part-repository.spec.ts
```diff
@@ -240,23 +240,18 @@ describe(`${MongoPartRepository.name}`, () => {
   })
 
   describe(`${MongoPartRepository.prototype.getParts.name}`, () => {
+    // TODO: Remove below comment, ones 'MongoEntityConverter' is gone.
+    // The below test should never fail, while we have the 'MongoEntityConverter', as it is the mocked value that is asserted.
```
LindvedKrvang, line 244: No need for this...

LindvedKrvang: Goes for the others as well

src/data-access/repositories/test/mongo-rundown-repository.spec.ts
KvelaGorrrrnio: It is just because it is a promise/data object that you return. Throwing calls should be wrapped in lambdas.

src/data-access/repositories/test/mongo-part-repository.spec.ts
andr9528: Updated.

src/data-access/repositories/test/mongo-part-repository.spec.ts
```diff
@@ -86,16 +87,12 @@ describe(`${MongoPartRepository.name}`, () => {
...
       const nonExistingId: string = 'nonExistingId'
       const mongoPart: MongoPart = createMongoPart({})
       await testDatabase.populateDatabaseWithParts([mongoPart])
       const db = testDatabase.getDatabase()
 
-      when(mongoConverter.convertParts(anything())).thenReturn([])
-      const testee = createTestee({
-        mongoConverter: mongoConverter,
-      })
+      const testee = createTestee({ })
```
KvelaGorrrrnio, line 95: If you set the parameter of `createTestee` to have the default value `{}`, then you don't have to pass in an empty object, when using the default state.

src/model/entities/test/test-entity-factory.ts
KvelaGorrrrnio: I do get that, but the effort needed to put in some default values is quite small.
Also if you use the spread operator to overwrite the default values, you will get more readable methods:
```typescript
  public static createRundown(rundownInterface: Partial<RundownInterface> = {}): Rundown {
    return new Rundown({
      id: 'rundownId' + Math.floor(Math.random()*1000),
      name: 'rundownName',
      segments: [],
      ...rundownInterface,
    } as RundownInterface)
  }

KvelaGorrrrnio (approved):
A few comments for optimizations.

LindvedKrvang (approved):

src/model/entities/test/test-entity-factory.ts
andr9528: Have changed to use the suggested method signature, and also added default values for a few more properties on each entity. Still not for all properties, as i frankly don't know what would be an ok default for a great deal of the newer properties introduced by Lindved.

Ps. Hope its okay that i had a small amount of fun with some of the default values :P
It is only used for tests anyways. If not, then i will begrudgingly change them.

src/data-access/repositories/test/mongo-part-repository.spec.ts
andr9528: Done that now.

KvelaGorrrrnio (approved):

LindvedKrvang (approved):

PR #10, opened 2023-09-20
Sof 1539 Support AdLibs

LindvedKrvang:
SofieServer now supports "AdLibbing"/inserting Parts and Pieces

Implemented in this PR:
- Blueprints creates "InsertPartActions" for Cameras.
- Blueprint creates "StopSoundBed" PieceAction.
- Blueprint creates "MixTransitionForNextPart" PieceAction.
- SofieServer supports "AdLibbing"/inserting Parts as next.
- SofieServer supports "AdLibbing"/inserting Parts as the active Part.
- SofieServer supports "AdLibbing"/inserting Pieces into the active Part.
- SofieServer supports "AdLibbing"/inserting Pieces onto the next Part.
- SofieServer supports "mutating" action with "runtime" data before executing them.
- Rundown no longer unmarks and marks Segments as next when the "new" next Segment is the same Segment as the "old" next Segment.
- Deleted "AdLibPieces" and corresponding files since we no longer use this concept

Going forward the term "AdLib" will no longer exist in the code since it's a poorly defined term that has different meanings depending on who you ask.
In Sofie it basically means that we insert a Part or a Piece, so this is what it will be called going forward.


src/blueprints/tv2/factories/tv2-transition-factory.ts
```diff
@@ -0,0 +1,110 @@
...
+import { AtemMeTimelineObject, AtemTransition, AtemType } from '../../timeline-state-resolver-types/atem-types'
+import { DeviceType } from '../../../model/enums/device-type'
+import { Tv2BlueprintTimelineObject } from '../value-objects/tv2-meta-data'
+import { TimelineObject } from '../../../model/entities/timeline-object'
+
+const FRAME_RATE: number = 25
+
+const FRAMES_FOR_MIX_TRANSITION: number = 25 // TODO: Get from configuration.
+
+const MIX_TRANSITION_ID: string = 'nextTakeHasMixTransitionAction'
```
andr9528, line 17: Due to how this is used in `isTransitionAction(...)`, i get the sense that there might well exist other Mix transition ids in the future.

You might say YAGNI, but would it hurt to have this as an enum already?
The method i mentioned should then be able to easily be updated to get an array of elements in the enum, when the need arise in the future, or already now, if you want to make the change already.

If there isn't the possibility for more Mix transition ids, then why is the value put into an array in the method?

src/blueprints/tv2/factories/tv2-transition-factory.ts
```diff
@@ -0,0 +1,110 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  private updateAtemMeInput(action: Action, plannedPiece: Piece): Action {
+    const timelineObject: TimelineObject | undefined = plannedPiece.timelineObjects.find(timelineObject => timelineObject.layer === Tv2AtemLayer.PROGRAM)
+    if (!timelineObject) {
+      console.log(`Can't update Atem Me Input. No TimelineObject for '${Tv2AtemLayer.PROGRAM}' found on Piece '${plannedPiece.id}'.`)
```
andr9528, line 93: Small thing. When we update the log statements to use Tv2's logging NPM package, should this be an info log or a warning log?

If it should be a warning log, maybe consider using `console.warn(...)` for it. That way it easy to see waht logging type is intended to be used, for when the update to the NPM package.

Same goes for the log statement in the next if block, and perhabs any other log statements elsewhere.

src/blueprints/tv2/helpers/tv2-sisyfos-persistent-layer-finder.ts
```diff
@@ -47,14 +47,14 @@ export class Tv2SisyfosPersistentLayerFinder {
...
-      (previous: Piece | undefined, current: Piece) => !previous || previous.start <= current.start ? current : previous,
+      (previous: Piece | undefined, current: Piece) => !previous || previous.getStart() <= current.getStart() ? current : previous,
       undefined
     )
   }
 
   private isPiecePlaying(piece: Piece, partExecutedAt: number, time: number): boolean{
     const hasPieceStoppedPlaying: boolean =
-				piece.duration > 0 && piece.start + piece.duration + partExecutedAt <= time
+				piece.duration > 0 && piece.getStart() + piece.duration + partExecutedAt <= time
```
andr9528, line 57: We need to take a moment to add/update linting rules... This is indentet more than it should, unless i am mistaken.

src/blueprints/tv2/value-objects/tv2-layers.ts
```diff
@@ -1,51 +1,25 @@
-export enum Tv2CasparCgLayer {
-  CasparCGDVELoop = 'casparcg_dve_loop',
-  CasparCGLYD = 'casparcg_audio_lyd',
-  CasparPlayerClipPending = 'casparcg_player_clip_pending',
-  CasparPlayerJingle = 'casparcg_player_jingle',
+// The enum values in this file is taking from Blueprints and still has to match those values until we control ingest.
```
andr9528, line 1: "...is **taken** from..."

src/blueprints/tv2/value-objects/tv2-meta-data.ts
andr9528, line 1: Do we use "meta-data" or "metadata"?

src/model/value-objects/rundown-event.ts
```diff
@@ -37,16 +38,70 @@ export interface PartSetAsNextEvent extends PartEvent {
...
-export interface RundownAdLibPieceInsertedEvent extends RundownEvent {
-  type: RundownEventType.AD_LIB_PIECE_INSERTED,
-  segmentId: string
-  partId: string
-  adLibPiece: {
-    id: string
-    name: string
-    layer: string
+// TODO: Find a better way to type Parts and Piece for Inserted events.
+export interface PartInsertedAsOnAirEvent extends RundownEvent {
```
andr9528, line 42: The type defined for `part` in this and `PartInsertedAsNextEvent` appears to be 100% duplicates. Might not completely solve the TODO, but pulling those to an interface to eliminate the duplication is a good first step. 

That said i don't know of a good name for that interface, which might also be why you created the TODO.

src/blueprints/tv2/value-objects/tv2-meta-data.ts
RasmusAlbrektsen: "metadata" is the correct spelling so it should reflect that in the files name.

src/model/value-objects/rundown-event.ts
LindvedKrvang: It's deliberately implemented as is for now, so we don't try to come up with a half baked solution for this and just stick with.
Anders is going to take a look at this while he implements the frontend part of inserting Parts/Pieces.

src/blueprints/tv2/factories/tv2-transition-factory.ts
LindvedKrvang: Currently, there aren't any techincal limitations for using enum here. 
However, I am bit hesistant to use Enums to store ids.

> The method i mentioned should then be able to easily be updated to get an array of elements in the enum

The method already is easily updatable ;) 

> If there isn't the possibility for more Mix transition ids, then why is the value put into an array in the method?

You could argue YAGNI here, but there are going to be more transition ids. We still need to implement `dip` and `wipe` transitions. And we also might need to have a id for each transtion that is for next Part and one for active Part.

src/model/value-objects/rundown-event.ts
andr9528: As long as we agree that the duplication is gone before release, i am happy with it.
Fell free to mark resolved, ones read.

src/blueprints/tv2/factories/tv2-transition-factory.ts
LindvedKrvang: Honestly, the only I thing I thought about here was that I thought it was a bit too agressive to throw an error, hence the log. 

In general, I'm more in favour of throwing errors then just logging statement when something goes wrong. 
Personally, I think we should determine if this scenario shouldn't be possible, and hence throw an error. If the scenario is possible, then it's a valid use case and we shouldn't log at all.

src/data-access/repositories/mongo/mongo-action-repository.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  public async getAction(actionId: string): Promise<Action> {
+    return await this.getCollection().findOne({id: actionId}) as unknown as Action
+  }
+
+  public async saveActions(actions: Action[]): Promise<void> {
+    for (const action of actions) {
+      await this.getCollection().updateOne({ _id: action.id }, { $set: action }, { upsert: true })
+    }
```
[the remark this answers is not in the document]
LindvedKrvang: Yes, the `Array.map()` has some funky behavior when calling async methods

src/blueprints/tv2/value-objects/tv2-meta-data.ts
LindvedKrvang: As Rasmus points out, the correct spelling is medata. The file name is a leftover since I initally named it all "metaData" and didn't catch that one in my refactor.

It's been updated.

src/blueprints/tv2/factories/tv2-transition-factory.ts
andr9528: What i am trying to get at, is that if the Id was stored in an enum, and the moethod was taking the values of the enum to check for, then the method would already handle adding new id, without having to make any changes to it.

I am fine with keeping it as it. Just my mind wanting to prepare for the possibility of more ids being added, ahead of time.

src/data-access/repositories/mongo/mongo-action-repository.ts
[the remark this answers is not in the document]
andr9528: Hmm, something very similar to what i suggested is being used in `mongo-test-database.ts`, and i havn't seen any issues during any of the tests that use it.
The file i mention does call `insertOne(...)` instead of the `.updateOne(...)`, but as both return a promise, and the issue you mention sound to be related to that, i don't see what that difference should make.

Can you give some more details as to what it does weirdly?

src/blueprints/tv2/factories/tv2-transition-factory.ts
andr9528: I also favour throwing errors, but as you said, that is sometimes too aggresive.
When throwing an error is too aggresive, logging a warning, would be my go-to way of indicating that something isn't quite right, but not critically so.

If in the end it is decided to not throw an error, but it still isn't a valid use case, then i'd very much like for it to be updated to a `console.warn(...)`.

src/data-access/repositories/mongo/mongo-action-repository.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: @LindvedKrvang What funky behaviour are you referring to?
I don't see why the Promise.all approach, as @andr9528 suggested, shouldn't work. (Though the async and parentheses are unnecessary for the lambda :wink:)

src/model/entities/action.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface Action {
+  id: string
+  name: string
+  type: ActionType
+  data: unknown
+}
+
+export interface PartAction extends Action {
+  type: ActionType.INSERT_PART_AS_ON_AIR | ActionType.INSERT_PART_AS_NEXT
```
KvelaGorrrrnio, line 13: I don't know if I am a fan of this or not.
In my head, there would ideally be a `PieceActionType` and a `PartActionType`. Then `Action` could be the union of those or `Action.type` could be something more generic like `string`.

src/model/entities/test/part.spec.ts
```diff
@@ -1,10 +1,10 @@
 import { Part, PartInterface } from '../part'
 import { EntityMockFactory } from './entity-mock-factory'
 import { PartTimings } from '../../value-objects/part-timings'
-import { Piece } from '../piece'
+import { Piece, PieceInterface } from '../piece'
 
-describe('Part', () => {
-  describe('getTimings', () => {
+describe(`${Part.name}`, () => {
```
KvelaGorrrrnio, line 6: We the string interpolation?
You can just pass in the attribute value, like so:
```typescript
describe(Part.name, () => {
```

src/model/value-objects/rundown-event.ts
KvelaGorrrrnio: The plan is to move these types to the presentation layer and use the DTO definitions for the entity content.

LindvedKrvang: @KvelaGorrrrnio But they are used on in the bussiness-logic-layer, which is not allowed to know about the presentation layer. How do you plan to handle that?

src/data-access/repositories/mongo/mongo-action-repository.ts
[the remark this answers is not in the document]
LindvedKrvang: @KvelaGorrrrnio Something that you once told me, but I can't remember it. But if there is no issue, then I will update it

src/blueprints/timeline-state-resolver-types/atem-types.ts
```diff
@@ -24,3 +35,26 @@ export enum AtemType {
...
+    rate: number
+  }
+  dip?: {
+    rate: number,
+    input: number
+  },
+  wipe?: {
+    //...
+  }
+}
```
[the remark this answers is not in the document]
LindvedKrvang: No (or it wouldn't make logical sense).
These refer to the different MixEffects that Atem is able to do when changing source.
It can't both do a Mix and a Dip effect at the same time.

src/blueprints/tv2/tv2-blueprints-facade.ts
```diff
@@ -3,12 +3,21 @@ import { Tv2Blueprint } from './tv2-blueprint'
...
 export class Tv2BlueprintsFacade {
   public static createBlueprint(): Blueprint {
     return new Tv2Blueprint(
       new Tv2EndStateForPartService(new Tv2SisyfosPersistentLayerFinder()),
-      new Tv2OnTimelineGenerateService(new Tv2SisyfosPersistentLayerFinder())
+      new Tv2OnTimelineGenerateService(new Tv2SisyfosPersistentLayerFinder()),
+      new Tv2ActionsService(
+        new Tv2CameraFactory(),
+        new Tv2TransitionFactory(),
+        new Tv2AudioFactory()
```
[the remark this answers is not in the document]
LindvedKrvang: I'm open for naming suggestions

src/business-logic/services/execute-action-service.ts
```diff
@@ -0,0 +1,112 @@
...
+      return action
+    }
+    const mutateActionMethods: MutateActionMethods | undefined = this.blueprint.getMutateActionMethods(action)
+    if (!mutateActionMethods) {
+      return action
+    }
+    const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
+    const plannedPiece: Piece | undefined = rundown.getNextPart().getPieces().find(mutateActionMethods.plannedPiecePredicate)
+    if (plannedPiece) {
+      action = mutateActionMethods.updateActionWithPlannedPieceData(action, plannedPiece)
```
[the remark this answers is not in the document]
LindvedKrvang: Yes, if it mutates it from a PieceAction to a PartAction then I would say you are misusing the Actions system and should just have created a PartAction from the beginning.

src/business-logic/services/interfaces/rundown-event-builder.ts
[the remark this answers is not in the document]
LindvedKrvang: But that forces you to move the interface `RundownEventEmitter` to the model layer, so we still can reference it from the bussiness logic layer (which might be okay)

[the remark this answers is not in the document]
LindvedKrvang: That forces you to move some interfaces to the model layer, so we still can reference them by the bussiness logic layer. Be mindfull of that

src/model/entities/action.ts
LindvedKrvang: Then you are introducing two "values" to tell what kind of action is and when you need to figure out which kind of Action you are dealing with you need to do something like:
```
switch(action.type) {
  case ActionType.PIECE: {
    switch(action.subType) {
       case ActionSubType.INSER_AS_ON_AIR: {
         ...
       }
       case ActionSubType.INSER_AS_NEXT: {
         ...
       }
    }
  }
  case: ActionType.PART: {
    switch(action.subType): {
      case ActionSubType.INSER_AS_ON_AIR: {
        ...
      }
      case ActionSubType.INSER_AS_NEXT: {
        ...
      }
    }
  }
}
```
I would very much like us to avoid that...

src/business-logic/services/execute-action-service.ts
```diff
@@ -0,0 +1,112 @@
...
+    private readonly configurationRepository: ConfigurationRepository,
+    private readonly actionRepository: ActionRepository,
+    private readonly rundownService: RundownService,
+    private readonly rundownRepository: RundownRepository,
+    private readonly blueprint: Blueprint
+  ) {}
+
+  public async getActions(): Promise<Action[]> {
+    const configuration: Configuration = await this.configurationRepository.getConfiguration()
+    // TODO: These should be generated on ingest. Move them once we control ingest.
```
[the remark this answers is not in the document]
LindvedKrvang: Updated to say `The Actions` instead of `These`

src/business-logic/services/execute-action-service.ts
```diff
@@ -0,0 +1,112 @@
...
+    const mutateActionMethods: MutateActionMethods | undefined = this.blueprint.getMutateActionMethods(action)
+    if (!mutateActionMethods) {
+      return action
+    }
+    const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
+    const plannedPiece: Piece | undefined = rundown.getNextPart().getPieces().find(mutateActionMethods.plannedPiecePredicate)
+    if (plannedPiece) {
+      action = mutateActionMethods.updateActionWithPlannedPieceData(action, plannedPiece)
+    }
+    return action
```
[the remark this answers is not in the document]
LindvedKrvang: Negated the if-statement.

src/business-logic/services/execute-action-service.ts
```diff
@@ -0,0 +1,112 @@
...
+  private async insertPartAsOnAir(partAction: PartAction, rundownId: string): Promise<void> {
+    const part: Part = this.createPartFromAction(partAction)
+    await this.rundownService.insertPartAsOnAir(rundownId, part)
+  }
+
+  private createPartFromAction(partAction: PartAction): Part {
+    const pieces: Piece[] = partAction.data.pieceInterfaces.map(pieceInterface => new Piece(pieceInterface))
+
+    const partInterface: PartInterface = partAction.data.partInterface
+    partInterface.id = `${partInterface.id}_${Date.now()}`
```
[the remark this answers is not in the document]
LindvedKrvang: Extracted to a method called `makeUnique()`, so the name is not implementation dependent and the usage becomes: `partInterface.id = this.makeUnique(partInterface.id)`

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -40,21 +38,25 @@ export class RundownTimelineService implements RundownService {
...
     this.emitAddInfinitePieces(rundown, [])
 
-    const activateEvent: RundownEvent = this.rundownEventBuilder.buildActivateEvent(rundown)
-    this.rundownEventEmitter.emitRundownEvent(activateEvent)
-
-    const setNextEvent: RundownEvent = this.rundownEventBuilder.buildSetNextEvent(rundown)
-    this.rundownEventEmitter.emitRundownEvent(setNextEvent)
+    this.sendEvents(rundown, [
+      this.rundownEventBuilder.buildActivateEvent,
+      this.rundownEventBuilder.buildSetNextEvent]
```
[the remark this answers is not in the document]
LindvedKrvang: Agreed and updated

src/model/entities/test/part.spec.ts
LindvedKrvang: I just took what we did somewhere else :p 
Updated

src/blueprints/tv2/factories/tv2-transition-factory.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    return {
+      id: MIX_TRANSITION_ID,
+      name: 'Mix transition on next take',
+      type: ActionType.INSERT_PIECE_AS_NEXT,
+      data: pieceInterface
+    }
+  }
+
+  private getTimeFromFrames(frames: number): number {
```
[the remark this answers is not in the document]
andr9528: As long as the unit is specifed ones during some part of a calculation, it should be expected that the unit doesn't change until otherwise specifed, or easily deducable from the calculations done. Where the line goes for what is easily deducable, and what is not, is the hard part.

A good example of this, is when you are calculation duration. As long as `InMs` is specified in one of the used input values, the result of the calculation is expected to also be `InMs`, without stating so in the name, unless the name specifically states that it is now e.g `InSeconds`. 
Sometimes the change of unit can be deduced easily, with the calculation being done, e.g dividing a value known to be `InMs` by 1000, means now it is in seconds. Because of this, specifying the unit can becomes less important, but will still useful, if further calculation is done with it.

src/blueprints/tv2/helpers/tv2-sisyfos-persistent-layer-finder.ts
LindvedKrvang: Updated

src/blueprints/tv2/factories/tv2-transition-factory.ts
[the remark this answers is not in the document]
LindvedKrvang: Personally, I think it would be nice if the variable name gave an indication of if the value is in milliseconds, seconds, minutes, hours etc. So something like `InMs`, `secondsToWaitForX` etc

I think `epoch` should be used when it's important that the `number` provided is an epoch timestamp and there is no other indications for it. For example the `CallbackScheduler` interface:
```
export interface CallbackScheduler {
  start(epochTimeToExecuteCallback: number, callback: () => void): void
  stop(): void
}
```

I think having a "standard" unit might not be a good option. Everything is case by case dependent and it's easy to forget when we are deviating from the standard. 
But if we are going for some sort of standard, then I suggest to use milliseconds :)

src/data-access/repositories/mongo/mongo-action-repository.ts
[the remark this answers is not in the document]
LindvedKrvang: Updated.

src/model/entities/test/part.spec.ts
andr9528: Great! 

Thats a remnant from some rules in Sofie-Core, where i originally found that string interpolation was needed for it to be happy. It was me who origianlly suggested that way, and since moving to the new repo, we didn't check if the string interpolation was still needed.

Updating the lines in the test files touched by my current In Review task.

src/business-logic/services/test/execute-action-service.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+import { ConfigurationRepository } from '../../../data-access/repositories/interfaces/configuration-repository'
+import { ActionRepository } from '../../../data-access/repositories/interfaces/action-repository'
+import { RundownService } from '../interfaces/rundown-service'
+import { RundownRepository } from '../../../data-access/repositories/interfaces/rundown-repository'
+import { Blueprint } from '../../../model/value-objects/blueprint'
+import { anyOfClass, anyString, capture, instance, mock, verify, when } from '@typestrong/ts-mockito'
+import { Piece, PieceInterface } from '../../../model/entities/piece'
+
+describe(ExecuteActionService.name, () => {
+  describe(`${ExecuteActionService.prototype.executeAction.name}`, () => {
```
andr9528, line 14: Per Anders comment regarding string interpolation, this can also be updated, to not use string interpolation.

andr9528 (approved):

src/presentation/controllers/action-controller.ts
```diff
@@ -0,0 +1,37 @@
...
+
+@RestController('/actions')
+export class ActionController extends BaseController {
+
+  constructor(private readonly actionService: ActionService, private readonly httpErrorHandler: HttpErrorHandler) {
+    super()
+  }
+
+  @GetRequest()
+  public async getActions(_reg: Request, res: Response): Promise<void> {
```
RasmusAlbrektsen, line 17: I'm so confused, i commented on this once before as well i think. Is this not supposed to be `req` rather than `reg`? It's the same in the method below.

RasmusAlbrektsen (approved):

src/presentation/controllers/action-controller.ts
andr9528: Just took a look at all the existing controllers. Seems to be spelled `reg` everywhere, which as Ralle points out is wrong.

andr9528: I have fixed it in the other controllers, and created PR for that (#16).

src/blueprints/timeline-state-resolver-types/atem-types.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: It would be nice to have these as 3 different transition interfaces then unioned as `AtemTransitionSettings`.

src/blueprints/tv2/factories/tv2-transition-factory.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: I am also for having the units. It makes it less confusing. But is it something that we want to introduce on the entities (when we have the freedom to do so)? E.g. `Piece.start` becomes `Piece.startOffsetInMs` or something like that?

src/blueprints/tv2/tv2-blueprints-facade.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: If the outward facing concern is creating actions, then I would insert `Action` between the piece type and the `Factory`. E.g.: `Tv2CameraActionFactory`.

src/business-logic/services/execute-action-service.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: Then in that case, we can use generic types to extend the MutateActionsMethods interface members with Generic types that extend Action in order to ensure that the input and output types correspond.

src/model/entities/action.ts
KvelaGorrrrnio: I might have formulated it badly, but you would not need the subtype attribute. I was thinking something like:

```typescript
type ActionType = PieceActionType | PartActionType

enum PieceActionType {
  INSERT_PIECE_AS_ON_AIR = 'INSERT_PIECE_AS_ON_AIR',
  INSERT_PIECE_AS_NEXT = 'INSERT_PIECE_AS_NEXT',
  // ...
}

enum PartActionType {
  INSERT_PART_AS_ON_AIR = 'INSERT_PART_AS_ON_AIR',
  INSERT_PART_AS_NEXT = 'INSERT_PART_AS_NEXT',
  // ...
}

interface Action {
  type: ActionType
  // ...
}

interface PartAction extends Action {
  type: PartActionType
  // ...
}

interface PieceAction extends Action {
  type: PieceActionType
  // ...
}

```

KvelaGorrrrnio (approved):
I would still like to the previous comments to be addressed prior to merging if possible.

src/blueprints/timeline-state-resolver-types/atem-types.ts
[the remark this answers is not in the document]
LindvedKrvang: I agree, but I think we should make those when we start moving stuff to blueprints.

src/blueprints/tv2/factories/tv2-transition-factory.ts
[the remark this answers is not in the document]
LindvedKrvang: You asks some good questions @KvelaGorrrrnio 

Something in me wants to let it just stay `Piece.start`, but that kinda goes against my previous comment and I don't have any good arguments to let it stay `Piece.start`

src/blueprints/tv2/tv2-blueprints-facade.ts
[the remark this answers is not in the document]
LindvedKrvang: Renamed the factories

src/model/entities/action.ts
LindvedKrvang: I have updated to your suggestion.

PR #12, opened 2023-09-27
SOF-1621 Introduced budgetDuration on Segment.

temk-tv2:


src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -148,13 +149,14 @@ export class MongoEntityConverter {
 
   public convertSegment(mongoSegment: MongoSegment): Segment {
     return new Segment({
+      budgetDuration: mongoSegment.budgetDuration,
```
LindvedKrvang, line 152: Personal preference, but it would be very nice if we could keep the ids and names as some of the first things when setting attributes.
Ideally, we should keep the same order as the attributes have been declared in.

LindvedKrvang (approved):

andr9528 (approved):

src/data-access/repositories/mongo/mongo-entity-converter.ts
andr9528: I will agree with Lindved that keeping the Id and names as the first ones is nice.
I don't have strong preference regarding the order of the rest of the attributes, thought.

PR #13, opened 2023-09-28
SOF-1523 No longer does a Take on activation of Rundown

LindvedKrvang:
This PR is dependent on: https://github.com/tv2/sofie-server/pull/10
It also has breaking changes for the RundownActivatedEvent that has been fixed in: https://github.com/tv2/ng-sofie/pull/24

The Rundown no longer perfoms a Take on activation.
To achieve this, it has been necessary to make the active Part and active Segment on the Rundown optional.

### DIFF
--- a/src/model/entities/test/rundown.spec.ts
+++ b/src/model/entities/test/rundown.spec.ts
@@ -1,4 +1,4 @@
-import { Segment } from '../segment'
+import { Segment, SegmentInterface } from '../segment'
 import { Rundown, RundownInterface } from '../rundown'
 import { Part } from '../part'
 import { Piece } from '../piece'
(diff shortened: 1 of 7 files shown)


src/model/entities/test/rundown.spec.ts
```diff
@@ -9,8 +9,9 @@ import { NotActivatedException } from '../../exceptions/not-activated-exception'
 import { NotFoundException } from '../../exceptions/not-found-exception'
 import { LastPartInSegmentException } from '../../exceptions/last-part-in-segment-exception'
 import { LastPartInRundownException } from '../../exceptions/last-part-in-rundown-exception'
+import { AlreadyActivatedException } from '../../exceptions/already-activated-exception'
 
-describe('Rundown', () => {
+describe(`${Rundown.name}`, () => {
```
andr9528, line 14: As per Anders comment on #10, string interpolation can be removed. 
Same goes for a number of other describes.

andr9528 (approved):

RasmusAlbrektsen (approved):

KvelaGorrrrnio (approved):

PR #14, opened 2023-09-28
SOF-1539 No longer build a NextPart group

LindvedKrvang:
Dependent on https://github.com/tv2/sofie-server/pull/13

We no longer build a "Next Part" group when building the Timeline.
The "Next Part" group was used to "do a Take" for autoNext.
We, however, do an actual Take when we autoNext, so there is no need for the "Next Part" group anymore.


src/business-logic/services/test/superfly-timeline-builder.spec.ts
```diff
@@ -34,8 +33,8 @@ const LOOKAHEAD_PRIORITY: number = 0.1
 const BASELINE_PRIORITY: number = 0
 const LOW_PRIORITY: number = -1
 
-describe('superfly-timeline-builder', () => {
-  describe('buildTimeline', () => {
+describe(`${SuperflyTimelineBuilder.name}`, () => {
```
andr9528, line 36: As per Anders comment on #10, string interpolation can be removed.
Same goes for the describe below.

andr9528 (approved):

KvelaGorrrrnio (approved):

PR #15, opened 2023-09-28
Sof 1574 - expose blueprint configuration

andr9528:
- Adds new endpoint to retrieve all ShowStyle and Studio blueprint configurations.
- Adds new endpoint allowing extraction of ShowStyleVariant from database. 
   - An example of the returned value is shown below. 
   - It contains all the fields in the database, for the requested ShowStyleVariant, even when they are not made available in the backend via the interface.

```json
{
    "_id": "r3Kom8JweCsps8dFr",
    "showStyleBaseId": "show0",
    "name": "DEFAULT",
    "blueprintConfig": {
        "SelectedGfxSetupName": {
            "value": "Nef4dxBukL87LDq1L",
            "label": "SC"
        },
        "GfxDefaults": [
            {
                "DefaultSetupName": {
                    "value": "Nef4dxBukL87LDq1L",
                    "label": "SC"
                },
                "_id": "",
                "DefaultSchema": {
                    "value": "N/A",
                    "label": "N/A"
                },
                "DefaultDesign": {
                    "value": "zZcwo8Z8o2rd4Avxy",
                    "label": "DESIGN_SC"
                }
            }
        ]
    },
    "_rundownVersionHash": "BLwJKeFnTDvEEYoOoRsLumB_l3s_",
    "_rank": 0
}
```


src/presentation/controllers/configuration-controller.ts
```diff
@@ -0,0 +1,29 @@
...
+export class ConfigurationController extends BaseController {
+  constructor(
+    private readonly configurationRepository: ConfigurationRepository,
+    private readonly httpErrorHandler: HttpErrorHandler
+  ) {
+    super()
+  }
+
+  @GetRequest()
+  public async getConfigurations(_req: Request, res: Response): Promise<void> {
```
LindvedKrvang, line 18: We only have one configuration for our system, so I think this should be singular not plural. Especially since you are not returning a list of configurations.

Also, the method is called `getConfiguration`, but you are only returning the Blueprint part of the configuration.
Either the method should be renamed to `getBlueprintConfiguration` or you should return the entire configuration.

andr9528: Has been updated to `getBlueprintConfiguration`. 

I had it in plural, as it is returning the `BlueprintConfiguration` for both `ShowStyle` and `Studio`, thus it being multiple, and in my mind needing to be plural.

LindvedKrvang: You are returning one Configuration that has a Studio and Showstyle subsection. It's still just one configuration

LindvedKrvang: A good indicator for if something should be plural or not: Is it an array or not?

src/model/entities/show-style-variant.ts
```diff
@@ -0,0 +1,6 @@
+export interface ShowStyleVariant {
+  _id: string
```
LindvedKrvang, line 2: We are using `id`, not `_id` for our entities.

src/presentation/controllers/configuration-controller.ts
```diff
@@ -0,0 +1,44 @@
...
+      res.send({
+        showStyleConfiguration: configuration.showStyle.blueprintConfiguration,
+        studioConfiguration: configuration.studio.blueprintConfiguration
+      })
+    } catch (error) {
+      this.httpErrorHandler.handleError(res, error as Exception)
+    }
+  }
+
+  @GetRequest('/:rundownId')
```
LindvedKrvang, line 33: This should be `'/rundowns/:rundownId'`.
If you leave it as `'/:rundownId'`, then the URL will look like this `'configurations/this-is-an-id'` where `this-is-an-id` refers to a Rundown id, but by only looking at the URL, REST says that is is a Configuration id, which it is not.

In REST, you must put the entity before the id in your URL, so if you want to add a Rundown id in your Configuration URL, the URL must look like `'configurations/rundowns/this-is-an-id'`.

The reasons that `RundownController` does not need to specify `/rundowns` in each endpoint is becasue `/rundowns` is specified on the controller level, just as you have speficied `/configurations` on the controller level in here.

src/model/entities/rundown.ts
```diff
@@ -50,10 +51,13 @@ export class Rundown extends BasicRundown {
 
   private persistentState?: RundownPersistentState
 
+  private readonly showStyleVariantId: string
```
LindvedKrvang, line 54: I'm not a fan of `Rundown` being the one that knows which `ShowStyleVariant` has been selected.
The `Rundown` has the responsibility of keeping track of its Segments and Parts and ensuring they are in the right state.

Which ShowStyleVariant that has been selected is out of the scope of what `Rundown` should know.

But it's possible to have different variants selected for different Rundowns, so we would need some way to map them, which might be out-of-scope for this task.

Leave it as is for now :) Anders and I need to have a talk about this.

src/model/entities/show-style-variant.ts
andr9528: Fixed in the upcomming commit.

src/presentation/controllers/configuration-controller.ts
andr9528: Fixed in next commit.

src/model/entities/show-style-variant.ts
andr9528: Has been updated to `id`, and mapping from Mongo version to our entity added to MongoEntityConverter.

KvelaGorrrrnio (approved):

src/presentation/controllers/configuration-controller.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @GetRequest()
+  public async getBlueprintConfiguration(_req: Request, res: Response): Promise<void> {
+    try {
+      const configuration: Configuration = await this.configurationRepository.getConfiguration()
+      res.send({
+        showStyleConfiguration: configuration.showStyle.blueprintConfiguration,
+        studioConfiguration: configuration.studio.blueprintConfiguration
+      })
```
LindvedKrvang, line 27: There is a slight mismatch between the URL and what the endpoint returns.
The URL is `/configurations` so I would expect to get a Configuration for the entire system, but what I am receving is just the BlueprintsConfigurations (which fit with the method name).

Either the URL should be changed to be `/configurations/blueprints` or the method should be updated to return the entire Configuration `res.send(configuration)`

LindvedKrvang: Not to mention that we don't have an id for which configuration we want to return so `/configurations` should by REST standards return a list of all Configurations.
Not sure if we should change that now...

LindvedKrvang (approved):
It looks good.
I would like you to look at the comment about `getBlueprintConfiguration` before merging :)

src/presentation/controllers/configuration-controller.ts
andr9528: Hmm, i don't know if it would be usefull for client to know of the layers in `Studio`. 
For know i will change the endpoint to include `/blueprints`.

PR #18, opened 2023-10-02
Budget duration made optional

KvelaGorrrrnio:
- Segment budget duration is marked as optional and converts `null` values to `undefined`.


src/model/entities/segment.ts
```diff
@@ -54,6 +54,9 @@ export class Segment {
   }
 
   public setAsNext(): void {
+    if (!this.isSegmentOnAir) {
+      this.reset()
+    }
     this.isSegmentNext = true
     if (!this.isSegmentOnAir) {
       this.reset()
```
LindvedKrvang, line 62: I think you should take a closer look at this code ;)

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -148,7 +148,7 @@ export class MongoEntityConverter {
       isOnAir: false,
       isNext: false,
       parts: [],
-      budgetDuration: mongoSegment.budgetDuration,
+      budgetDuration: mongoSegment.budgetDuration ?? undefined, // Ensure that null values are stripped
```
LindvedKrvang, line 151: `undefined` values aren't much better than `null` values...

src/model/entities/segment.ts
LindvedKrvang: One could even say you should do a double take :D

andr9528 (approved):
As long as the woopsy Lindved commented on is fixed, i am happy with it.

src/model/entities/segment.ts
KvelaGorrrrnio: The reset below is from develop and not this branch. I have merged develop into the branch and removed the double reset.

src/data-access/repositories/mongo/mongo-entity-converter.ts
KvelaGorrrrnio: No, but they are better than nullish values :)

RasmusAlbrektsen (approved):

PR #19, opened 2023-10-03
Sof 1546 - move static actions

andr9528:
- Recreates Static actions from Blueprint in the new code structure. Currently reimplemented:
  - ThemeOut
    - [x] Tested and verified that it has the same result in staging
  - OverlayInitialize
    - [x] Double checked that the send TimelineObject matches the original one.
  - ContinueGraphic
    - [x] Tested and verified that it has the same result in staging 
  - ClearGraphic
    - [x] Tested and verified that it has the same result in staging
  - AllOutGraphic
    - [x] Tested and verified that it has the same result in staging
  - MicrophoneUp
    - [x] Tested and verified that it has the same result in staging
  - MicrophoneDown
    - [x] Tested and verified that it has the same result in staging
  - ResynchronizeAudio (originally resyncSisyfos)
    - [x] Tested and verified that it has the same result in staging
  - FadePersistedAudio
    - [x] Tested and verified that it has the same result in staging
  - DownstreamKeyer[0..*]On
    - [x] Tested and verified that it has the same result in staging
  - DownstreamKeyerOff[0..*]Off
    - [x] Tested and verified that it has the same result in staging
- Reworks `Tv2ActionsService.generateActions(...)` for better readability and seperation of concerns.

- [x] Test that the actions can be retrieved and executed on a rundown via the assosiated endpoints, via e.g Postman.


src/blueprints/timeline-state-resolver-types/viz-types.ts
```diff
@@ -0,0 +1,39 @@
...
+    deviceType: DeviceType.VIZMSE
+    type: VizType.LOAD_ALL_ELEMENTS
+  }
+}
+
+export interface VizMseContinueTimelineObject extends TimelineObject {
+  content: {
+    deviceType: DeviceType.VIZMSE
+    type: VizType.CONTINUE
+    direction: -1 | 1 | undefined // Taken directly from Blueprints. Perhaps can/should be changed.
```
LindvedKrvang, line 25: This is most likely the values that the physical Viz device expects, so we should keep that in mind if we change it.
Also, it's the values that TSR expects.

src/blueprints/tv2/factories/tv2-graphic-action-factory.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+    return {
+      id: 'continueGraphicAction',
+      name: 'Continue Graphic',
+      type: PieceActionType.INSERT_PIECE_AS_ON_AIR,
+      data: pieceInterface
+    }
+  }
+
+  // Todo: Question if 'Graphics' should be removed from method name.
```
LindvedKrvang, line 44: In this particular case it would make sense to just call it `createThemeOutAction`.

Sidenote: We should really confirm if it is called "Theme out" in english.

src/blueprints/tv2/factories/tv2-graphic-action-factory.ts
```diff
@@ -0,0 +1,101 @@
...
+      partId: '',
+      type: PieceType.GRAPHIC,
+      layer: Tv2SourceLayer.GRAPHIC_ACTION_COMMAND,
+      transitionType: TransitionType.NO_TRANSITION,
+      pieceLifespan: PieceLifespan.WITHIN_PART,
+      isPlanned: false,
+      start: 0,
+      preRollDuration: 0,
+      postRollDuration: 0,
+    } as PieceInterface
```
LindvedKrvang, line 79: We are not using `as` in production code

src/blueprints/tv2/factories/tv2-graphic-action-factory.ts
andr9528: Name has been updated to `createThemeOutAction`.

src/blueprints/tv2/factories/tv2-audio-action-factory.ts
```diff
@@ -6,30 +6,84 @@ import { PieceLifespan } from '../../../model/enums/piece-lifespan'
...
       type: PieceType.AUDIO,
-      layer: Tv2SourceLayer.AUDIO_BED,
       pieceLifespan: PieceLifespan.WITHIN_PART,
       transitionType: TransitionType.NO_TRANSITION,
       isPlanned: false,
       start: 0,
-      duration,
       preRollDuration: 0,
       postRollDuration: 0,
+    } as PieceInterface
```
LindvedKrvang, line 54: You should `as` when creating/instantiating interfaces/entities as you run the risk of violating the internal state, but simply not given the needed data.

Instead of using `as` here, just write the object out with "default" values. They are going to be overridden anyway when you do:
```
const pieceInterface: PieceInterface = { 
  ...this.createDefaultAudioPieceInterface,
  id: 'idOfAction'
  ...
}
```

src/blueprints/tv2/factories/tv2-audio-action-factory.ts
```diff
@@ -6,30 +6,84 @@ import { PieceLifespan } from '../../../model/enums/piece-lifespan'
...
+  }
+
+  private createMicrophoneDownAction(blueprintConfiguration: Tv2BlueprintConfiguration): PieceAction {
+    const duration: number = 0
```
LindvedKrvang, line 58: You are only using `duration` once, so no need to create the local variable here.

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
```diff
@@ -0,0 +1,37 @@
...
+  public createMicrophoneDownTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration): SisyfosChannelsTimelineObject {
+    return this.createMicrophoneTimelineObject(blueprintConfiguration, 0)
+  }
+
+  public createMicrophoneUpTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration): SisyfosChannelsTimelineObject {
+    return this.createMicrophoneTimelineObject(blueprintConfiguration, 1)
+  }
+
+  // Todo: Figure out what 'isPmg' stands for and write full name instead of abbreviation
+  private createMicrophoneTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration, isPmg:  0 | 1 | 2): SisyfosChannelsTimelineObject {
```
LindvedKrvang, line 17: It's `isPgm` and not `isPmg` :) 

If this is for Sisyfos, then it's whether the audio fader is active or not.

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
```diff
@@ -0,0 +1,37 @@
...
+        start: 0
+      },
+      priority: 10,
+      layer: Tv2SisyfosLayer.STUDIO_MICS,
+      content: {
+        deviceType: DeviceType.SISYFOS,
+        type: SisyfosType.CHANNELS,
+        channels: blueprintConfiguration.studio.StudioMics.map(studioMic => ({
+          mappedLayer: studioMic,
+          isPgm: isPmg
```
LindvedKrvang, line 30: When you fix the spelling mistake in the parameter then you can just write `isPgm`

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
LindvedKrvang: I believe the values means:
- 0 - Off
- 1 - On
- 2 - VoiceOver

LindvedKrvang: Would be nice if was extracted into an enum.
For now that enum can just live privately inside this file

src/blueprints/tv2/factories/tv2-audio-action-factory.ts
andr9528: Has been updated for the commented on file, and the graphic file too.

src/blueprints/tv2/factories/tv2-audio-action-factory.ts
andr9528: Has been moved inline.

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
andr9528: Fixed typo, and changed type to an enum.

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
andr9528: Done.

src/blueprints/tv2/factories/tv2-viz-gfx-timeline-object-factory.ts
```diff
@@ -63,4 +64,31 @@ export class Tv2VizGraphicsTimelineObjectFactory implements Tv2GraphicsTimelineO
...
+        channelsToSendCommands: sendCommands ? ['OVL1', 'FULL1', 'WALL1'] : undefined,
+        showName: blueprintConfiguration.showStyle.selectedGraphicsSetup.OvlShowName ?? ''
+      }
+    }
+  }
+
+  public createGfxAlternativeOutTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration, duration: number): VizMseClearGfxTimelineObject {
+    return this.buildGfxClearTimelineObject(blueprintConfiguration, duration, false)
+
+  }
```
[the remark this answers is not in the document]
LindvedKrvang: If you have trouble creating the TimelineObject without the Content, then you can parse the Content as a parameter:
`return this.createBaseTimelineObject(duration, this.createGfxClearContent(blueprintConfiguration)`

src/blueprints/timeline-state-resolver-types/sisyfos-types.ts
```diff
@@ -21,6 +21,15 @@ export interface SisyfosChannelsTimelineObject extends TimelineObject {
   }
 }
 
+// Todo: Figure out if 'resync' can be changed to `resynchronize`, or if it is the value that Sisyfos expects.
+export interface SisyfosResynchronizeTimelineObject extends TimelineObject {
```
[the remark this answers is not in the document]
andr9528: Might help if i place it at the correct line. It was aimed at the attribute of the content. Has been moved to just above that one.

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
```diff
@@ -1,20 +1,30 @@
...
+  SisyfosChannelsTimelineObject,
+  SisyfosResynchronizeTimelineObject,
+  SisyfosType
+} from '../../timeline-state-resolver-types/sisyfos-types'
 import { Tv2SisyfosLayer } from '../value-objects/tv2-layers'
 import { DeviceType } from '../../../model/enums/device-type'
 import { Tv2BlueprintConfiguration } from '../value-objects/tv2-blueprint-configuration'
+import { EmptyTimelineObject } from '../../timeline-state-resolver-types/abstract-types'
+
+enum SisyfosFaderState {
```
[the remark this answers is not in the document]
andr9528: Uhhm ok, had added that, not that i know what the difference between it being there and not is...

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
```diff
@@ -27,11 +37,43 @@ export class Tv2SisyfosAudioTimelineObjectFactory implements Tv2AudioTimelineObj
...
+      layer: Tv2SisyfosLayer.RESYNCHRONIZE,
+      content: {
+        deviceType: DeviceType.SISYFOS,
+        type: SisyfosType.CHANNEL,
+        resync: true
+      }
+    }
+  }
+
+  public createStopAudioBedTimelineObject(duration: number): EmptyTimelineObject {
```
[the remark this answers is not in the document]
andr9528: Is it the responsibility of this factory to decide how long of an duration it has, or is that a responsibility of the action factory, that uses this factory?

If it isn't an input anylonger, couldn't we end up with a piece defining a duration of 3000, while some timeline factory defines it as 5000. Would that be okay?

src/blueprints/tv2/factories/tv2-viz-gfx-timeline-object-factory.ts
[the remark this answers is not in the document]
andr9528: Updated to your second suggestion. Content current has type `object`, not sure if it can/need to be specifed better.

src/blueprints/timeline-state-resolver-types/sisyfos-types.ts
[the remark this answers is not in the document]
andr9528: With that said, is the name of the attributes in `content`, set in stone, or can they be changed?

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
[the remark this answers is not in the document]
LindvedKrvang: I haven't looked at the blueprint code to see how it does it there.
However, it's the Action itself that decides the duration of the Action. Whatever is in iNews has no effect on this.
So if all callers of this method just parse the same duration to it (and their value is hardcoded), then there is no need for it to be a parameter.

src/blueprints/timeline-state-resolver-types/sisyfos-types.ts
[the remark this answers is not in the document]
LindvedKrvang: > Might help if i place it at the correct line.

Perfect example as to why we want to minimize comments because they become outdated/moved from what they are commenting on extremely quickly. 
This one didn't even last the commit it was introduced in...

src/blueprints/tv2/factories/tv2-audio-action-factory.ts
```diff
@@ -1,44 +1,116 @@
...
+      layer: Tv2SourceLayer.AUDIO_ACTION_COMMAND,
       isPlanned: false,
       start: 0,
-      duration,
       preRollDuration: 0,
       postRollDuration: 0,
       tags: [],
+      timelineObjects: [],
+    }
+  }
```
KvelaGorrrrnio, line 85: This. might just be me, but having an empty id, does not seem like a valid declaration. Instead of creating a default/empty piece, you could make a function that creates an AudioPieceInterface given a subset. By doing so, you can require that some of the attributes needs to be set, like `id`. E.g.
```typescript
export class Tv2AudioActionFactory {
  private createAudioPieceInterface(piece: Pick<PieceInterface, 'id'> & Partial<PieceInterface>): PieceInterface {
    // ...
  }
}
```

src/blueprints/tv2/factories/tv2-gfx-action-factory.ts
```diff
@@ -0,0 +1,176 @@
...
+      ...this.createDownstreamKeyerOnActions(blueprintConfiguration),
+      ...this.createDownstreamKeyerOffActions(blueprintConfiguration)
+    ]
+  }
+
+  private createDownstreamKeyerOffActions(blueprintConfiguration: Tv2BlueprintConfiguration): PieceAction[] {
+    return blueprintConfiguration.studio.SwitcherSource.DSK.map(config => this.createDownStreamKeyerAction(config, 'Off', false))
+  }
+
+  private createDownStreamKeyerAction(downstreamKeyer: Tv2DownstreamKeyer, actionName: string, isOn: boolean): PieceAction {
```
KvelaGorrrrnio, line 33: Can't the `actionName` be inferred from `isOn`?

src/blueprints/tv2/value-objects/tv2-layers.ts
```diff
@@ -1,8 +1,15 @@
 // The enum values in this file is taken from Blueprints and still has to match those values until we control ingest.
+export enum Tv2VizLayer { // Todo: figure out if enum should have 'Viz' in name or not, as technically other graphics might need the same value.
```
KvelaGorrrrnio, line 2: Are there any plans on when these todos should be done?

src/blueprints/tv2/factories/tv2-gfx-action-factory.ts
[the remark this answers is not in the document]
andr9528: Will be happy to change it, again, ones both techleads agree... 
Lindved is the one to say `Gfx`, after i originally made it `Graphics`

src/blueprints/tv2/factories/tv2-audio-action-factory.ts
LindvedKrvang: As long as this isn't propagated to the TimelineObjects

src/blueprints/tv2/value-objects/tv2-show-style-blueprint-configuration.ts
```diff
@@ -1 +1,19 @@
-export interface Tv2ShowStyleBlueprintConfiguration {}
+export interface Tv2ShowStyleBlueprintConfiguration {
+  GfxDefaults: GfxDefault
```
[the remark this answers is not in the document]
LindvedKrvang: Yes, unforutnately these need to match the name in the database which is currently is `GfxDefaults`.
The reason for this is because of the "framework" Core is exposing to Blueprints to do UI where we had to use a "Table".
This means that the values of `GfxDefaults` is saved as an array to the database by Core.
But since we enforce that there can only ever be one `GfxDefaults` we map the array to the a single unit when converting the "unknown" configuration parsed from SofieServer to Blueprints into a Tv2Configuration.

When we control settings, we have the option to change it.

src/blueprints/tv2/factories/tv2-gfx-action-factory.ts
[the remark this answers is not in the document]
LindvedKrvang: We had a task at some point to change all `Graphics` to `Gfx` in Blueprints. I believe it was because Jacob wants the display value to be `Gfx`.

Personnally, I am more fan of `Graphics` since it's not an abbreviation.
I am all for changing it back to `Graphics`, but all display values should still say `Gfx` per Jacobs request (unless you can convice him the value should be `Graphics` as well).

src/blueprints/tv2/value-objects/tv2-studio-blueprint-configuration.ts
```diff
@@ -20,3 +21,20 @@ export interface Tv2SourceMappingWithSound extends Tv2SourceMapping {
...
+
+export interface Tv2DownstreamKeyer {
+  _id: string
+  Number: number
+  Key: number
+  Fill: number
+  DefaultOn: boolean
+  Roles: string[]
+  Clip: number,
+  Gain: number
```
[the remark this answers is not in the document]
LindvedKrvang: You are absolutely correct. Unfortunately...

src/blueprints/timeline-state-resolver-types/sisyfos-types.ts
```diff
@@ -21,6 +21,15 @@ export interface SisyfosChannelsTimelineObject extends TimelineObject {
   }
 }
 
+export interface SisyfosResynchronizeTimelineObject extends TimelineObject {
+  content: {
+    deviceType: DeviceType.SISYFOS;
+    type: SisyfosType.CHANNEL;
+    // Todo: Figure out if 'resync' can be changed to `resynchronize`, or if it is the value that Sisyfos expects.
+    resync: boolean
```
LindvedKrvang, line 29: `resync` is the value used by TSR, so no, this can't be changed.

src/blueprints/timeline-state-resolver-types/viz-types.ts
LindvedKrvang: Which means the comment `Perhaps can/should be changed` should be deleted.

src/blueprints/tv2/factories/tv2-atem-video-switcher-timeline-factory.ts
```diff
@@ -0,0 +1,42 @@
...
+        type: AtemType.DSK,
+        dsk: {
+          onAir,
+          sources: {
+            fillSource: downstreamKeyer.Fill,
+            cutSource: downstreamKeyer.Key
+          },
+          properties: {
+            clip: downstreamKeyer.Clip * 10, // input is percents (0-100), atem uses 1-000
+            gain: downstreamKeyer.Gain * 10, // input is percents (0-100), atem uses 1-000
```
LindvedKrvang, line 29: Instead of the comment you could have a method called something like `convertPercentageToAtemPercentageValue()` so it would be:
```
clip: this.convertPercentageToAtemPercentageValue(downstreamKeyeer.Clip),
gain: this.convertPercentageToAtemPercentageValue(downstreamKeyeer.Gain)
```

Also: I believe you intented to write `1 - 1000` and not `1 - 000`?

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
```diff
@@ -0,0 +1,79 @@
...
+export class Tv2SisyfosAudioTimelineObjectFactory implements Tv2AudioTimelineObjectFactory {
+  public createMicrophoneDownTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration): SisyfosChannelsTimelineObject {
+    return this.buildMicrophoneTimelineObject(blueprintConfiguration, SisyfosFaderState.OFF)
+  }
+
+  public createMicrophoneUpTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration): SisyfosChannelsTimelineObject {
+    return this.buildMicrophoneTimelineObject(blueprintConfiguration, SisyfosFaderState.ON)
+  }
+
+  private buildMicrophoneTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration, isPgm:  SisyfosFaderState): SisyfosChannelsTimelineObject {
```
LindvedKrvang, line 27: Since `isPgm` is of the type `SisyfosFaderState`, then it would be nice the parameter name reflected that.
So it would be `sisyfosFaderState: SisyfosFaderState`.

src/blueprints/tv2/factories/tv2-video-switcher-action-factory.ts
```diff
@@ -0,0 +1,73 @@
...
+
+
+  private createDefaultVideoSwitcherPieceInterface(): PieceInterface {
+    return {
+      id: '',
+      name: '',
+      duration: 0,
+      partId: '',
+      type: PieceType.GRAPHIC, // Todo: Is this still correct, now that it is in a 'VideoSwitcher' factory?
+      layer: Tv2SourceLayer.GRAPHIC_ACTION_COMMAND, // Todo: Is this still correct, now that it is in a 'VideoSwitcher' factory?
```
LindvedKrvang, line 58: The method is called `createDefaultVideoSwitcherPieceInterface` which should give you a pretty clear indication that types in here should not be graphics

src/blueprints/tv2/factories/tv2-gfx-action-factory.ts
andr9528: Yes it can. Doing so means that instead of the boolean being a pass through argument, it changes what the method does, meaning it is a flag. 

I had created it originally like suggested, and then asked for Lindveds input, after which we reworked it to this.

src/blueprints/tv2/factories/tv2-viz-gfx-timeline-object-factory.ts
```diff
@@ -0,0 +1,114 @@
...
+        reference: Tv2VizLayer.GRAPHIC_PILOT     
+      }
+    }
+  }
+
+  public createGfxClearTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration, duration: number): VizMseClearGfxTimelineObject {
+    return (this.createBaseTimelineObjectWithContent(
+      duration,
+      this.createGfxClearContent(blueprintConfiguration)
+    ) as VizMseClearGfxTimelineObject) // Todo: figure out if this `as` usage is okay.
```
LindvedKrvang, line 72: Just ask us instead of comitting TODOs.

If you just follow the interface you defined in `Tv2GfxTimelineObjectFactory` then you don't need to cast at all.
All you need to do is change the return type from `VizMseClearGfxTimelineObject` to `TimelineOjbect` (which is what the factory interface returns) and you can get rid of the `as`.

src/blueprints/tv2/factories/tv2-viz-gfx-timeline-object-factory.ts
```diff
@@ -0,0 +1,114 @@
...
+      layer: Tv2VizLayer.GFX_ACTIONS,
+      content: content
+    }
+  }
+
+  public createGfxAllOutTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration, duration: number): VizMseClearGfxTimelineObject {
+    return (this.createBaseTimelineObjectWithContent(
+      duration,
+      this.createGfxAllOutContent(blueprintConfiguration)
+    ) as VizMseClearGfxTimelineObject) // Todo: figure out if this `as` usage is okay.
```
LindvedKrvang, line 101: Same comment as to the other one of this TODO.

src/blueprints/tv2/value-objects/tv2-layers.ts
andr9528: Yes. it was meant to help me remeber to ask you and Lindved, which it now has done.
Should it have `Viz` as part of the name or not?

src/business-logic/services/execute-action-service.ts
```diff
@@ -100,6 +100,7 @@ export class ExecuteActionService implements ActionService {
 
   private async insertPieceAsOnAir(pieceAction: PieceAction, rundownId: string): Promise<void> {
     const piece: Piece = this.createPieceFromAction(pieceAction)
+    piece.setExecutedAt(Date.now())
```
LindvedKrvang, line 103: A test ensuring that we don't remove this line at some point would be nice :)

LindvedKrvang (commented):
You have put all your interfaces for the TimelineObject factories under the folder `value-objects`.
The interfaces for the factories differ in quite important way: The factories are _not_ value objects (e.i. they are not objects that you instantiate and parse around like an Action), they are a form of service.

So they shouldn't be in the folder `value-objects`.
I am not really sure where the best place to place them is, but right now I'm leaning towards an `interface` folder within the `factory` folder.
@KvelaGorrrrnio any thoughs on this?

src/blueprints/tv2/value-objects/tv2-studio-blueprint-configuration.ts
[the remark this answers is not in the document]
andr9528: `DefaultOn` is not being used anywhere, at the moment, should the field remain or be deleted?

The Dsk in the database also has a `Toggle` field, which controlls if actions in the original Blueprint should be generated for the toggled DSK.
I have chosen to not pull that value out, yet, and simply generate action for all the DSK's at all times.
In the new frontside, wouldn't it be better to just use that toggle, to decide if the action can be fired or not?

src/blueprints/tv2/factories/tv2-atem-video-switcher-timeline-factory.ts
andr9528: Created and changed to use the suggested method.

The comment is dirrectly copied from Blueprint, and i think originally written by Krzy.
Should i add the missing 1?

src/blueprints/tv2/factories/tv2-sisyfos-audio-timeline-object-factory.ts
andr9528: Has been renamed.

src/blueprints/tv2/factories/tv2-viz-gfx-timeline-object-factory.ts
andr9528: Extracted `content` type of the TimelineObject that was cast to, to its own interface, and used that as return type for `createGfxClearContent`, instead of `object`.

Also changed the return type as suggested, and removed the casting.

src/blueprints/tv2/factories/tv2-viz-gfx-timeline-object-factory.ts
andr9528: Same response.

src/blueprints/tv2/factories/tv2-video-switcher-action-factory.ts
andr9528: New enum entries has been added, and this location changed to use them.

src/blueprints/tv2/factories/tv2-gfx-action-factory.ts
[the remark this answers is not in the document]
andr9528: Have changed mentions of `Gfx` back to `Graphic`, where it isn't linked to a database field, or potentially shown to users.

src/blueprints/tv2/value-objects/tv2-layers.ts
andr9528: `Viz` has been changed to `Graphic`.

src/business-logic/services/execute-action-service.ts
andr9528: Test added :)

andr9528:
The new `Microphone Up` & `Microphone Down` actions doesn't function 100% as sofie-core.

In sofie-core, they are toggle buttons. 
Ones either is clicked, cliking either will turn off the one clicked, i.e clicking `Up` after `Down` will turn off the `Down` button.
Clicking `Down` will turn off microphones as expected. Clicking it again, will turn on microphones, as `Down` is turned off.

With the new action, executing `Down` twice, like in the above example, doesn't turn it back on, which is to be expected in my mind atleast.

I think this is a matter of how the frontend wishes to handle it.
Don't know if this needs to be verified with Jacob.

src/blueprints/tv2/factories/tv2-audio-action-factory.ts
andr9528: Have done as suggested by Anders in action factories.

src/blueprints/timeline-state-resolver-types/sisyfos-types.ts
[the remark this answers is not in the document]
andr9528: Todo has been removed.

src/blueprints/timeline-state-resolver-types/viz-types.ts
andr9528: Part of comment is gone.

src/blueprints/tv2/factories/tv2-atem-video-switcher-timeline-factory.ts
LindvedKrvang: > Should i add the missing 1?

Preferable we wouldn't need the comment at all.

src/blueprints/tv2/factories/tv2-atem-video-switcher-timeline-factory.ts
```diff
@@ -7,14 +7,15 @@ import { Tv2AtemLayer } from '../value-objects/tv2-layers'
...
+  public createDownstreamKeyerTimelineObject(downstreamKeyer: Tv2DownstreamKeyer, onAir: boolean): AtemDownstreamKeyerTimelineObject {
+    const downstreamKeyerNumber: string = String(downstreamKeyer.Number + 1)
     return {
       id: '',
       enable: {
         while: 1
       },
       priority: 10,
-      layer,
+      layer: `${this.getDownstreamKeyerLayerPrefix()}_${downstreamKeyerNumber}`, // Taken from Blueprints.
```
LindvedKrvang, line 18: Don't need to add `Taken from blueprints` so a many times. All of this is taken from Blueprints ;)

src/blueprints/tv2/factories/tv2-atem-video-switcher-timeline-factory.ts
```diff
@@ -25,8 +26,8 @@ export class Tv2AtemVideoSwitcherTimelineFactory implements Tv2VideoSwitcherTime
             cutSource: downstreamKeyer.Key
           },
           properties: {
-            clip: downstreamKeyer.Clip * 10, // input is percents (0-100), atem uses 1-000
-            gain: downstreamKeyer.Gain * 10, // input is percents (0-100), atem uses 1-000
+            clip: this.convertPercentageToAtemPercentageValue(downstreamKeyer.Clip), // input is percents (0-100), atem uses 1-000
```
LindvedKrvang, line 29: The point of making this method was so you could remove the comment....

LindvedKrvang: Or atleast move the comment to the new method, so we only have to have it once...

src/model/enums/piece-type.ts
```diff
@@ -5,6 +5,7 @@ export enum PieceType {
   GRAPHIC = 'GRAPHIC',
   SERVER = 'SERVER',
   AUDIO = 'AUDIO',
-  TRANSITION = 'TRANSITION'
+  TRANSITION = 'TRANSITION',
+  VIDEO_SWITCHER = 'VIDEO_SWITCHER'
```
LindvedKrvang, line 9: Would be nice if we could call it `VIDEO_MIXER` instead of switcher. Switcher is the American term.

LindvedKrvang (dismissed):

src/blueprints/tv2/factories/tv2-atem-video-switcher-timeline-factory.ts
andr9528: Moved, and reworded, to a TsDoc Return on the method.

src/blueprints/tv2/factories/tv2-atem-video-switcher-timeline-factory.ts
andr9528: Have removed the comment.

src/model/enums/piece-type.ts
andr9528: Have renamed all `Video Switcher` to `Video Mixer`.

src/blueprints/tv2/factories/tv2-atem-video-mixer-timeline-factory.ts
```diff
@@ -0,0 +1,48 @@
+import { Tv2VideoMixerTimelineObjectFactory } from '../value-objects/factories/tv2-video-mixer-timeline-object-factory'
+import { Tv2DownstreamKeyer } from '../value-objects/tv2-studio-blueprint-configuration'
+import { AtemDownstreamKeyerTimelineObject, AtemType } from '../../timeline-state-resolver-types/atem-types'
+import { Tv2AtemLayer } from '../value-objects/tv2-layers'
+import { DeviceType } from '../../../model/enums/device-type'
+
+export class Tv2AtemVideoMixerTimelineFactory implements Tv2VideoMixerTimelineObjectFactory {
+  public createDownstreamKeyerTimelineObject(downstreamKeyer: Tv2DownstreamKeyer, onAir: boolean): AtemDownstreamKeyerTimelineObject {
+    const downstreamKeyerNumber: string = String(downstreamKeyer.Number + 1)
```
KvelaGorrrrnio, line 9: Converting it to a string is unnecessary.

src/blueprints/tv2/factories/tv2-graphic-action-factory.ts
```diff
@@ -0,0 +1,138 @@
...
+import { Tv2GraphicTimelineObjectFactory } from '../value-objects/factories/tv2-graphic-timeline-object-factory'
+import { PieceLifespan } from '../../../model/enums/piece-lifespan'
+import { PieceActionType } from '../../../model/enums/action-type'
+
+export class Tv2GraphicActionFactory {
+  constructor(
+    private readonly graphicTimelineObjectFactory: Tv2GraphicTimelineObjectFactory
+  ) { }
+
+  public createGraphicActions(blueprintConfiguration: Tv2BlueprintConfiguration): Action[] {
```
KvelaGorrrrnio, line 16: I am pretty sure that it should be `Graphics` and not `Graphic`.

src/business-logic/services/test/execute-action-service.spec.ts
```diff
@@ -91,6 +91,18 @@ describe(ExecuteActionService.name, () => {
...
+      it('updates Piece ExecutedAt to be set', async () => {
+        const action: PieceAction = createPieceAction(PieceActionType.INSERT_PIECE_AS_ON_AIR)
+        const rundownServiceMock: RundownService = mock<RundownService>()
+
+        const testee: ExecuteActionService = createTestee({rundownService: rundownServiceMock}, {action})
+        await testee.executeAction(action.id, 'rundownId')
+
+        const [, executedPiece] = capture(rundownServiceMock.insertPieceAsOnAir).first()
+
+        expect(executedPiece.getExecutedAt()).not.toBe(undefined)
```
KvelaGorrrrnio, line 104: You could also use the fake timers of jest to check that it is Date.now that is used for the executed at value.

src/model/enums/piece-type.ts
```diff
@@ -5,6 +5,7 @@ export enum PieceType {
   GRAPHIC = 'GRAPHIC',
   SERVER = 'SERVER',
   AUDIO = 'AUDIO',
-  TRANSITION = 'TRANSITION'
+  TRANSITION = 'TRANSITION',
+  VIDEO_MIXER = 'VIDEO_MIXER'
```
KvelaGorrrrnio, line 9: It seems weird that a Piece has the type of Video Mixer. The Piece type should reflect what it content it controls, not the which playout device.

KvelaGorrrrnio (changes requested):

src/blueprints/tv2/factories/tv2-atem-video-mixer-timeline-factory.ts
andr9528: Forgot to remove the convertion, and update type, when what caused the need was removed.
Fixed now.

src/business-logic/services/test/execute-action-service.spec.ts
andr9528: Not 100% sure that the change i did was the way you intended, but it works, and the test failes as it should, when the line being tested is missing.

src/model/enums/piece-type.ts
andr9528: Not sure if changes are needed for this comment. 

Lindved was the one who said that a `Video_Mixer` PieceType was more correct than the originally used `Graphic` PieceType. in the `Tv2AtemVideoMixerTimelineFactory`.

src/blueprints/tv2/factories/tv2-graphic-action-factory.ts
andr9528: Unless i am mistaken, then each action only creates/alters one Graphic element.
The method in question creates multiple actions.

Thus it creates multiple actions (plural), that each creates/alters one graphic element (singular).

src/model/enums/piece-type.ts
LindvedKrvang: > Not sure if changes are needed for this comment.

They are. We don't have Pieces that are VideoMixers. The PieceTypes that are send to a VideoMixer are types like CAMERA and LIVE.
The PieceType reflects what kind of Piece it is. E.g. if it's a Camera or if it's a grahic or an Replay. It does not represent what hardware device is being sent to.

To sum up: There should not be a `VIDEO_MIXER` attribute on `PieceType`.

src/blueprints/tv2/factories/tv2-graphic-action-factory.ts
LindvedKrvang: I'm in favour of `Graphic` instead of `Graphics` as André mentioned it's multiple actions being created that each does a single thing to a Graphic.

src/blueprints/tv2/factories/tv2-video-mixer-action-factory.ts
```diff
@@ -11,11 +9,11 @@ import { PieceType } from '../../../model/enums/piece-type'
 import { Tv2SourceLayer } from '../value-objects/tv2-layers'
 import { TransitionType } from '../../../model/enums/transition-type'
 
-export class Tv2VideoSwitcherActionFactory {
-  constructor(private readonly videoSwitcherTimelineObjectFactory: Tv2VideoSwitcherTimelineObjectFactory) {
+export class Tv2VideoMixerActionFactory {
```
LindvedKrvang, line 12: Something feels wrong about having a `VideoMixerActionFactory`. It's too broad and it doesn't reflect any particular type of Piece or Action.
Case and point: `VideoMixerActionFactory` indicates that all Actions that needs to send data to a VideoMixer should be created in here, but all our TransitionActions are being created by our `TransitionActionFactory`. So we already have "VideoMixerActions" created elsewhere than our `VideoMixerActionFactory`.

We should either rename our `VideoMixerActionFactory` something else or move the transition from `TransitionActionFactory` in here.

src/model/enums/piece-type.ts
LindvedKrvang: You are currently using this attribute for creating the DskActions. These Actions are a bit special compared to our other Actions because they themselves don't do anything visual. They aren't a grahic or a camera that we can put on air. 
Instead, they just send a command to a device. In this case I think it's fine to use `PieceType.UNKNOWN`.

src/business-logic/services/test/execute-action-service.spec.ts
KvelaGorrrrnio: It was not what I intended :D Instead of manually overwriting `Date.now` which is now fixed to that mock for the remainder of the tests, you can use jests fake timers: https://jestjs.io/docs/timer-mocks

So you would, before the test, call `jest.useFakeTimers()` and after the test call `jest.useRealTimers()`. Remember that this should be outside the the scope of the test function. If it is done within the scope and the function throws, then the fake timers are leaking to other tests. You don't have to specify a Date, but just use the date that is locked by jest.

src/blueprints/tv2/factories/tv2-video-mixer-action-factory.ts
KvelaGorrrrnio: I don't think that VideoMixerActionFactory is way off. What we want to create actions for is essentially for configuring the video mixer. So it could be renamed to `Tv2VideoMixerConfigurationActionFactory`.

src/model/enums/piece-type.ts
andr9528: Removed `VIDEO_MIXER`, and changed used PieceType for Dsk to `UNKNOWN`.

src/business-logic/services/test/execute-action-service.spec.ts
andr9528: Fixed

src/blueprints/tv2/factories/tv2-video-mixer-action-factory.ts
andr9528: Has been renamed to `Tv2VideoMixerConfigurationActionFactory`.

src/blueprints/tv2/factories/tv2-graphic-action-factory.ts
andr9528: Has been renamed to be with 's'.

KvelaGorrrrnio (approved):
Good job 💪

LindvedKrvang (approved):

PR #22, opened 2023-10-11
Sof 1596/run mongodb as docker

KvelaGorrrrnio:
- `yarn start-mongo`  start a mongo instance on port 3002.
- `yarn stop-mongo` stops the instance.
- `yarn watch` calls `yarn start-mongo` prior to starting watching.
- Connects to database on port 3002 with directConnection set to true.

Cause of changing it to draft: There seems to be some instability in how it can be used. The PR should be merged when it is stable.

### DIFF
--- a/mongodb/compose.yml
+++ b/mongodb/compose.yml
@@ -1,0 +1,15 @@
+services:
+  mongo:
+    container_name: 'sofie-mongo'
+    image: 'mongo:6.0.1'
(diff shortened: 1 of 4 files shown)


LindvedKrvang (dismissed):

k-primeholding-dim (approved):
Was able to start/stop mongo locally.
I see local along with meteor one in MDB Compass GUI.

PR #23, opened 2023-10-11
Actions have blueprints configurable metadata.

KvelaGorrrrnio:
- `ActionDto` now has a description attribute.
- `ActionDto` now has a metadata attribute.
- Tv2 specific Action types have been added to the TV2 blueprints folder and is used for the action factories.
- The actions endpoint is changed from `/api/actions` to `/api/actions/rundowns/:rundownId` in preparation for rundowns to be fetched on a rundown basis.


src/presentation/controllers/action-controller.ts
```diff
@@ -13,7 +13,7 @@ export class ActionController extends BaseController {
     super()
   }
 
-  @GetRequest()
+  @GetRequest('/rundowns/:rundownId')
```
andr9528, line 16: I know you made this change in preperation, but wouldn't it be better to make this change, when needed?

At the moment it will just result in us needing to send a rundownId, which will not be used, when performing the request.

KvelaGorrrrnio: I think it is fine to have it setup already now, since it allows the frontend to implement logic towards it. And it will be needed when implementing some of the actions, e.g. video clips.

src/blueprints/tv2/value-objects/tv2-action.ts
```diff
@@ -0,0 +1,46 @@
+import { PartAction, PieceAction } from '../../../model/entities/action'
+
+export enum Tv2ActionContentType {
```
LindvedKrvang, line 3: Are these going to be one-to-one with `PieceType`?

LindvedKrvang (approved):

src/blueprints/tv2/value-objects/tv2-action.ts
KvelaGorrrrnio: Most likely.

andr9528 (approved):

PR #24, opened 2023-10-12
Sof 1555/ingest updates

LindvedKrvang:
Sofie Server now listens for Ingest updates, updates its own state and emits the updates.

Implemented:
- Rundown Created
- Rundown Updated
- Rundown Deleted
- Segment Created - Handles if the Segment is Active or not.
- Segment is Updated - Handles if the Segment is Active or not.
- Segment Deleted - Handles if the Segment is Active or not.
- Part Created - Handles if the Part is Active or not
- Part is Updated - Handles if the Part is Active or not
- Part Deleted - Handles if the Part is Active or not
- Floating/unfloating a Segment in iNews - Handles if the Segment is Active or not.
- Infinite Pieces - Handles if the Rundown is still using the Infinit Piece
- Ingest events is executed in "bulks". At the end of each "bulk" the Timeline is build and persisted and a SetNextEvent is emitted.
- Rundown now uses Cursors to keep track of the active/next Parts and Segments
- Able to differentiate if the SetNext was made by the system of by an external "user"
- Update monogdb to version 6.1.0


src/model/value-objects/rundown-event.ts
```diff
@@ -36,71 +35,56 @@ export interface PartSetAsNextEvent extends PartEvent {
   type: RundownEventType.SET_NEXT
 }
 
-// TODO: Find a better way to type Parts and Piece for Inserted events.
+// TODO: Find a better way to type Segments, Parts and Piece for Inserted events.
```
LindvedKrvang, line 38: Anders is currently working on this one.

src/business-logic/services/database-change-ingest-service.ts
```diff
@@ -0,0 +1,283 @@
...
+    const rundownDeletedEvent: RundownDeletedEvent = this.eventBuilder.buildRundownDeletedEvent(rundownId)
+    this.eventEmitter.emitRundownEvent(rundownDeletedEvent)
+    await this.rundownRepository.deleteRundown(rundownId)
+  }
+
+  private async createSegment(segment: Segment): Promise<void> {
+    const rundown: Rundown = await this.rundownRepository.getRundown(segment.rundownId)
+    try {
+      rundown.addSegment(segment)
+    } catch (error) {
```
KvelaGorrrrnio, line 175: There is a inconsistency in the naming of caught errors. Here error is used. Further down it is exception.

src/business-logic/services/superfly-timeline-builder.ts
```diff
@@ -63,7 +68,7 @@ export class SuperflyTimelineBuilder implements TimelineBuilder {
     timeline = this.createTimelineWithInfiniteGroups(rundown, timeline)
     timeline = this.updateTimelineAutoNextEpochTimestampFromNextPart(rundown, activePartTimelineGroup, timeline)
 
-    return timeline
+    return Promise.resolve(timeline)
```
KvelaGorrrrnio, line 71: Either the `Promise.resolve` or `async` should go from this method.

src/data-access/repositories/cache/cached-rundown-repository.ts
```diff
@@ -26,6 +26,26 @@ export class CachedRundownRepository implements RundownRepository {
...
+    for (const rundown of this.cachedRundowns.values()) {
+      const rundownHasSegment: boolean = rundown.getSegments().some(segment => segment.id === segmentId)
+      if (rundownHasSegment) {
+        return Promise.resolve(rundown)
+      }
+    }
+    throw new NotFoundException(`No Rundown found with a Segment for Segment id: ${segmentId}`)
+  }
+
+  public getRundownByPartId(partId: string): Promise<Rundown> {
```
KvelaGorrrrnio, line 39: Is it correctly assumed that this is created to handle the database changes and can be removed when sofie server controls ingest?

src/data-access/repositories/mongo/base-mongo-repository.ts
```diff
@@ -1,17 +1,17 @@
...
-import { MongoEntityConverter } from './mongo-entity-converter'
+import { MongoEntityConverter, MongoId } from './mongo-entity-converter'
 import { Collection } from 'mongodb'
 import { DatabaseNotConnectedException } from '../../../model/exceptions/database-not-connected-exception'
 
-export abstract class BaseMongoRepository {
+export abstract class BaseMongoRepository<T extends MongoId = MongoId> {
   protected constructor(
-    protected mongoDatabase: MongoDatabase,
+    protected mongoDatabase: MongoDatabase<T>,
```
KvelaGorrrrnio, line 8: This seems like a wrong use of generics. When would you restrict the entire database to a single model type?
My guess is that you want to have the generic type applying to the result of `getCollection`. If that is the case, you can achieve by making the getCollection method generic, like so:
```typescript
  protected <Model extends MongoId>getCollection(): Collection<Model>
```
And it would be nice if we used more descriptive names for generics than single letters :smile:

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -115,29 +132,77 @@ interface MongoLayerMapping {
...
-  public convertRundown(mongoRundown: MongoRundown, baselineTimelineObjects?: TimelineObject[]): Rundown {
+  public convertRundown(mongoRundown: MongoRundown, segments: Segment[], baselineTimelineObjects?: TimelineObject[], infinitePieces?: Piece[]): Rundown {
+    let alreadyActiveProperties: RundownAlreadyActiveProperties | undefined
+    if (mongoRundown.isActive) {
+      alreadyActiveProperties = {
+        activeCursor: this.convertMongoRundownCursorToRundownCursor(mongoRundown.activeCursor, segments),
+        nextCursor: this.convertMongoRundownCursorToRundownCursor(mongoRundown.nextCursor, segments),
+        infinitePieces: this.mapToInfinitePieceMap(infinitePieces ?? [])
+      }
+    }
```
KvelaGorrrrnio, line 143: We should use immutable variables as much as possible in calculations to make the flow deterministic and hereby easier to understand. When I see a let binding, I expect (and fear) that the value is replaced during its lifetime.
In this case you can make it a ternary (like below) or if that becomes too messy, you can extract to a separate method.
```typescript
    const alreadyActiveProperties: RundownAlreadyActiveProperties | undefined = !mongoRundown.isActive ? undefined : {
      activeCursor: this.convertMongoRundownCursorToRundownCursor(mongoRundown.activeCursor, segments),
      nextCursor: this.convertMongoRundownCursorToRundownCursor(mongoRundown.nextCursor, segments),
      infinitePieces: this.mapToInfinitePieceMap(infinitePieces ?? [])
    }
```

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -209,15 +280,25 @@ export class MongoEntityConverter {
...
 
   public convertToMongoPart(part: Part): MongoPart {
+    let timings: PartTimings | undefined = undefined
+    try {
+      timings = part.getTimings()
+    } catch (error) {
+      if ((error as Exception).errorCode !== ErrorCode.UNSUPPORTED_OPERATION) {
+        throw error
+      }
+    }
```
KvelaGorrrrnio, line 290: By extracting this to its own method, both methods would be simpler and you can avoid working with mutable variables.

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -209,15 +280,25 @@ export class MongoEntityConverter {
...
       title: part.name,
       segmentId: part.getSegmentId(),
-      _rank: part.rank,
+      _rank: part.getRank(),
       isOnAir: part.isOnAir(),
       isNext: part.isNext(),
+      isUnsynced: part.isUnsynced(),
+      timings,
       endState: part.getEndState()
     } as MongoPart
```
KvelaGorrrrnio, line 303: It would be nice if we could get rid of these type assertion.

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -284,6 +367,74 @@ export class MongoEntityConverter {
...
+  public convertToMongoPiece(piece: Piece): MongoPiece {
+    let enable: { start: number, duration?: number} = {
+      start: piece.getStart()
+    }
+    if (piece.duration) {
+      enable = {
+        ...enable,
+        duration: piece.duration
+      }
+    }
```
KvelaGorrrrnio, line 379: We should avoid as much as possible using mutable variables. It looks like this can be rewritten as:
```typescript
     const enable: MongoPiece['enable'] = {
       start: piece.getStart()
       duration: piece.duration
     }
```

src/data-access/repositories/mongo/mongo-enums.ts
```diff
@@ -0,0 +1,6 @@
+export enum MongoChangeEvent {
+  INSERT = 'insert',
```
KvelaGorrrrnio, line 2: Are we okay with the value being lowercased?

src/data-access/repositories/mongo/mongo-part-changed-listener.ts
```diff
@@ -0,0 +1,84 @@
...
+    mongoDatabase: MongoDatabase,
+    mongoEntityConverter: MongoEntityConverter,
+    private readonly partRepository: PartRepository
+  ) {
+    super(mongoDatabase, mongoEntityConverter)
+    mongoDatabase.onConnect(PART_COLLECTION_NAME, () => this.listenForChanges())
+  }
+
+  private listenForChanges(): void {
+    const options = { fullDocument: 'updateLookup' }
```
KvelaGorrrrnio, line 34: Do we have a type for `options`?

src/data-access/repositories/mongo/mongo-rundown-changed-listener.ts
```diff
@@ -0,0 +1,80 @@
...
+      case MongoChangeEvent.INSERT: {
+        const insertChange: ChangeStreamInsertDocument<MongoRundown> = change as ChangeStreamInsertDocument<MongoRundown>
+        const rundownId: string = insertChange.fullDocument._id
+        const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
+        this.onCreatedCallback(rundown)
+        break
+      }
+      case MongoChangeEvent.REPLACE: {
+        const replaceChange: ChangeStreamReplaceDocument<MongoRundown> = change as ChangeStreamReplaceDocument<MongoRundown>
+        const rundownId: string = replaceChange.documentKey._id as unknown as string
```
KvelaGorrrrnio, line 55: It looks like you are casting from `string` to `unknown` to `string`.

src/model/entities/segment.ts
```diff
@@ -88,7 +109,51 @@ export class Segment {
...
+  private compareParts(partOne: Part, partTwo: Part): number {
+    return partOne.getRank() - partTwo.getRank()
+  }
+
+  public addPart(part: Part): void {
+    if (!this.isInsertingPartAllowed(part)) {
+      return
+    }
+
+    const doesPartAlreadyExistOnSegment: boolean = this.parts.some(p => p.id === part.id)
```
KvelaGorrrrnio, line 124: Suggestion: You could rename the parameter to `partToAdd`. That way, you can avoid using `p` as an abbreviation for part.

src/model/enums/owner.ts
```diff
@@ -0,0 +1,4 @@
+export enum Owner {
```
KvelaGorrrrnio, line 1: Do we need to be more specific in what kind of owner?

src/model/exceptions/already-exist-exception.ts
```diff
@@ -0,0 +1,8 @@
+import { Exception } from './exception'
+import { ErrorCode } from '../enums/error-code'
+
+export class AlreadyExistException extends Exception {
+  constructor(message?: string) {
```
KvelaGorrrrnio, line 5: I would prefer that we make the message required for all exceptions. It forces us to write (hopefully) useful information  about what goes wrong in specific scenarios.

src/presentation/index.ts
KvelaGorrrrnio, line 1: Is it common practice to have the entry point lie in the presentation layer?

KvelaGorrrrnio (commented):
Good job :rocket: I have left some comments on structure and semantics.

src/business-logic/services/database-change-ingest-service.ts
```diff
@@ -0,0 +1,283 @@
...
+    if (!eventCallback) {
+      return
+    }
+
+    this.isExecutingEvent = true
+    eventCallback()
+      .catch(error => console.error('Error when executing Ingest event:', error))
+      .finally(() => {
+        this.isExecutingEvent = false
+        void this.buildRundowns()
```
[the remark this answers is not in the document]
LindvedKrvang: What do you suggest?

src/data-access/repositories/cache/cached-rundown-repository.ts
LindvedKrvang: Yes, is so we can find the Rundown a Part belongs to when we get an Ingest update for a Part, since the database only gives us information about the Part.

Whether it can be removed when we control ingest entirely depends on if the Ingest update will give us any information about the Rundown (since we would control it, we would give it some information, so yes, this should be deletable by then).

src/data-access/repositories/mongo/mongo-enums.ts
LindvedKrvang: Not sure if we have a choice. This is the value that MongoDB emits to indicate what event has happend.
I can try and see if I can make it uppercase. That or maybe an enum is the wrong choice here?

src/data-access/repositories/mongo/mongo-rundown-changed-listener.ts
LindvedKrvang: `documentKey._id` is an `ObjectId`. I will try and see if I can do something different.

src/model/enums/owner.ts
LindvedKrvang: Isn't that what `Owner.SYSTEM` and `Owner.EXTERNAL` does?

src/presentation/index.ts
LindvedKrvang: I don't think so.
I'm up for moving it

src/blueprints/tv2/helpers/tv2-sisyfos-persistent-layer-finder.ts
```diff
@@ -53,7 +53,9 @@ export class Tv2SisyfosPersistentLayerFinder {
   }
 
   private isPiecePlaying(piece: Piece, partExecutedAt: number, time: number): boolean{
-    const hasPieceStoppedPlaying: boolean = piece.duration > 0 && piece.getStart() + piece.duration + partExecutedAt <= time
+    const hasPieceStoppedPlaying: boolean = piece.duration
+      ? (piece.duration > 0 && piece.getStart() + piece.duration + partExecutedAt <= time)
```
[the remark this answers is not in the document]
LindvedKrvang: I guess not.
It's used to check if the Piece has started playing at all. If `duration` is zero, then the Piece hasn't begun playing. This is quite implicit, so perhaps we should make that more explicit?

src/business-logic/services/database-change-ingest-service.ts
```diff
@@ -0,0 +1,283 @@
...
+    this.eventEmitter.emitRundownEvent(rundownDeletedEvent)
+    await this.rundownRepository.deleteRundown(rundownId)
+  }
+
+  private async createSegment(segment: Segment): Promise<void> {
+    const rundown: Rundown = await this.rundownRepository.getRundown(segment.rundownId)
+    try {
+      rundown.addSegment(segment)
+    } catch (error) {
+      this.throwIfNot(ErrorCode.ALREADY_EXIST, error as Exception)
```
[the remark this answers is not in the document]
LindvedKrvang: > Another option is to do the check like it is done in most other files (in the catch block).

I went for the `throwIfNot` to not duplicate too much code in these functions. If this is an undesired approach. I will change it.

src/business-logic/services/database-change-ingest-service.ts
```diff
@@ -0,0 +1,283 @@
...
+      // Segment already exist on Rundown, so we don't need to do anything else.
+      return
+    }
+
+    const segmentCreatedEvent: SegmentCreatedEvent = this.eventBuilder.buildSegmentCreatedEvent(rundown, segment)
+    await this.persistRundown(rundown, segmentCreatedEvent)
```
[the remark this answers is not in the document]
LindvedKrvang: I left them out on purpose because I wanted the try-catch to just focus on the error from `rundown.addSegment()`.
If this is undesired let me know.

src/data-access/repositories/mongo/base-mongo-repository.ts
LindvedKrvang: This generic is to tell MongoDB that we are not using `ObjectId` on our Collection. If I just set the generic on the method, MongoDB still thinks that we are using `ObjectId` which gives us compilation errors on MongoDB version 6.

If I place the generic only on the method I get the error `Collection<MongoId> is not assignable to Collection<T>`.

I'm open for solutions to this.

> And it would be nice if we used more descriptive names for generics than single letters

`T` is the standard the letter to use when declaring a generic in most languages. Notice that even the error message we got from MongoDB is using `T`.
But it does seem like that it is not the standard in Typescript :p

LindvedKrvang: Which version of Mongo are you using? I belive your suggestion does not give any errors on versions below version 6.

src/data-access/repositories/mongo/mongo-entity-converter.ts
LindvedKrvang: I agree, but that is not the scope of this PR.

src/business-logic/services/superfly-timeline-builder.ts
LindvedKrvang: Removing the async makes Jest unable to catch errors in test and assert them.
Removing the Promise.resolve makes lint complaint about us not awaiting anyting.
Removing Promise from the return type is not an option.

I am up for suggestions

src/business-logic/services/database-change-ingest-service.ts
LindvedKrvang: Updated it so all try-catch for the "events" calls them `exception`. I left it as `error` for `executNextEvent()` and `buildRundowns()` because they have nothing "exception" specific going on.
Let me know if I should rename those as well.

src/data-access/repositories/mongo/mongo-entity-converter.ts
LindvedKrvang: I don't share you fear, but it has been updated :)

src/data-access/repositories/mongo/mongo-entity-converter.ts
LindvedKrvang: Updated

src/data-access/repositories/mongo/mongo-part-changed-listener.ts
LindvedKrvang: We do: `ChangeStreamOptions`
It's been updated.

src/model/entities/segment.ts
LindvedKrvang: Updated

src/data-access/repositories/mongo/mongo-enums.ts
LindvedKrvang: The values must be lowercased to match the values in Mongo. So unless we want to do some case-gymnastics, this seems to be the simplest way.
That or we should just make each a constant?

src/data-access/repositories/mongo/mongo-rundown-changed-listener.ts
LindvedKrvang: Updated to specify the generic type instead of casting.

src/presentation/index.ts
KvelaGorrrrnio: Personally I think it would be neat if it is `src/index.ts`, then it is easy to find :grin:

src/model/exceptions/already-exist-exception.ts
LindvedKrvang: Updated

src/blueprints/tv2/helpers/tv2-sisyfos-persistent-layer-finder.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: It would be nice to make it explicit :)

src/business-logic/services/database-change-ingest-service.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: Either of the two snippets below. What I want is that if buildRundowns fails, we catch the error. I know that currently that shouldn't happen, but we will most likely forget to update this, if we change the internals of buildRundowns.

```typescript
eventCallback()
  .catch(error => ...)
  .then(() => this.buildRundowns())
  .catch(error => console.log('Failed building rundowns after event execution', error))
  .finally(() => {
    this.isExecutingEvent = false
    this.executeNextEvent()
  })
```
or
```typescript
eventCallback()
  .catch(error => ...)
  .finally(() => {
    this.isExecutingEvent = false
    this.buildRundowns().catch(error => console.error('Failed building rundowns after event execution', error))
    this.executeNextEvent()
  })
```

src/business-logic/services/database-change-ingest-service.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: I would say that it is a single responsibility :D It can be extracted to its own method.

src/data-access/repositories/mongo/base-mongo-repository.ts
KvelaGorrrrnio: > T is the standard the letter to use when declaring a generic in most languages. Notice that even the error message we got from MongoDB is using T.
Just because something is widely used, does not mean that it is a good idea :wink:

The mongo driver version that I use is the one specified in package.json on this branch.
I tried applying the suggested change on this branch locally, and I got no warnings nor errors when compiling or running the code.

src/data-access/repositories/mongo/mongo-enums.ts
KvelaGorrrrnio: Personally, I am fine with having the enum values lowercased.

src/model/enums/owner.ts
KvelaGorrrrnio: To a degree. What can this owner own? Is it the entire system, a rundown or an action taken in the system?
We don't have to change it, but just be aware that the term is a bit ambiguous.

src/business-logic/services/superfly-timeline-builder.ts
LindvedKrvang: Beats me

src/data-access/repositories/mongo/base-mongo-repository.ts
LindvedKrvang: > Just because something is widely used, does not mean that it is a good idea

Notice my last sentece ;)

src/model/enums/owner.ts
LindvedKrvang: The Owner owns whatever the Owner is placed on. 
So currently, Owner can be place on a RundownCursor, so the Owner owns the RundownCursor. 
This enum is to tell _who_ owns it, not _what_ it owns.

src/presentation/index.ts
LindvedKrvang: Be my guest ;)
It is out of the scope of this task

src/data-access/repositories/mongo/mongo-part-changed-listener.ts
```diff
@@ -0,0 +1,85 @@
...
+  private listenForChanges(): void {
+    const options: ChangeStreamOptions = { fullDocument: 'updateLookup' }
+    const changeStream: ChangeStream = this.getCollection().watch<MongoSegment, ChangeStreamDocument<MongoSegment>>([], options)
+    changeStream.on('change', (change: ChangeStreamDocument<MongoPart>) => void this.onChange(change))
+    console.debug('### Listening for Part collection changes...')
+  }
+
+  private async onChange(change: ChangeStreamDocument<MongoPart>): Promise<void> {
+    switch (change.operationType) {
+      case MongoChangeEvent.INSERT: {
```
andr9528, line 43: This method has multiple responsibilities, as it both figures out what event it is, and then handles it.
Each case should be extracted to their own method, to get rid of these extra responsibilities.

The new method would look something along the lines of:
```typescript
...
case MongoChangeEvent.INSERT: {
  return this.onInsertedEvent(change as ChangeStreamInsertDocument<MongoPart>)
}
...
```

With luck we might not need to await the new method, allowing for some parallel processing of events.
Not entirely sure if not awaiting, could become an issue during the callbacks.

src/business-logic/services/superfly-timeline-builder.ts
KvelaGorrrrnio: Removing the async keyword makes it such that thrown errors are not picked up and converted to a rejected promise.
In order to remove async, the method body should be wrapped in a try-catch that converts it to a rejected promise.
I'll let it be up to you whether to keep async or do the promise conversion.

src/data-access/repositories/mongo/mongo-part-changed-listener.ts
LindvedKrvang: The responsibility of this method is to facilitate what happens when an event is received. It gets the data from the event and call the apropiated callback (sometimes it also calls a repository).

You can't avoid "facilitating" methods. What you can do is keep them on the same abstraction level.

You are not really gaining much extracting those methods and since `onChange` is current less than thirty lines, I would say it's overkill to extract them. If `onChange` grew larger, then I would also be in favour of extracting.

src/business-logic/services/database-change-ingest-service.ts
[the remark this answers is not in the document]
LindvedKrvang: But we need to exit `createSegment` if it throws. If you extract that part to a different function, then we just exit back into `createSegment` and run the emit code which we shouldn't.

This gives a vastly different behaviour than what we are trying to achieve:
image

src/business-logic/services/database-change-ingest-service.ts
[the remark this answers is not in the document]
LindvedKrvang: So I assume you then also wants to remove the try-catch from `buildRundowns` because I am handling the error there? 
image
If we don't remove it there then we just handle the same error twice (meaning that the second catch is redundant since it will never be called)

src/business-logic/services/superfly-timeline-builder.ts
LindvedKrvang: Keeping the async then

src/data-access/repositories/mongo/base-mongo-repository.ts
LindvedKrvang: Changed `T` to `Model`

src/blueprints/tv2/helpers/tv2-sisyfos-persistent-layer-finder.ts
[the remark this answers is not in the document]
LindvedKrvang: I have removed the `piece.duration > 0` since `piece.duration` evaluates to false if the duration is zero.

src/business-logic/services/database-change-ingest-service.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: I missed that part :) Keep the try-catch in the same method.

src/data-access/repositories/mongo/base-mongo-repository.ts
LindvedKrvang: Merged your changes for this :)

KvelaGorrrrnio (approved):
:rocket:

andr9528 (approved):

PR #25, opened 2023-10-14
Events moved from Business Logic layer to Presentation layer

KvelaGorrrrnio:
By moving the event building and handling from the business logic layer to the presentation layer, we can use DTOs for our entities in events and the business logic is freed from the burden of making presentation layer specific computations.

The only thing that is exposed from the presentation layer into the business logic layer is the `EventEmitterFacade`. The actual `RundownEventService` is hidden behind the `RundownEventListener` interface.

`Part.duration` and `PartDto.duration` are marked optional.


src/business-logic/facades/service-facade.ts
```diff
@@ -1,26 +1,23 @@
...
 import { TimelineBuilder } from '../services/interfaces/timeline-builder'
 import { SuperflyTimelineBuilder } from '../services/superfly-timeline-builder'
-import { RundownEventBuilderImplementation } from '../services/rundown-event-builder-implementation'
-import { RundownEventBuilder } from '../services/interfaces/rundown-event-builder'
 import { TimeoutCallbackScheduler } from '../services/timeout-callback-scheduler'
 import { JsonObjectCloner } from '../services/json-object-cloner'
 import { BlueprintsFacade } from '../../blueprints/blueprints-facade'
 import { ActionService } from '../services/interfaces/action-service'
 import { ExecuteActionService } from '../services/execute-action-service'
+import { EventEmitterFacade } from '../../presentation/facades/event-emitter-facade'
```
LindvedKrvang, line 11: Just leaving a comment so people are aware that we are going to be very strict about how much of the presentation layer the service layer is allowed to know.

Only the facades in the service layer are allowed to know anything about the presentation layer and the only think they are allowed to know about is facades in the presentation layer.
Further more, we are going to limit how many places a Service layer facade can call a presentation layer facade.

We currently _only_ allow this one place. And there needs to be a _strong_ argument for letting another place know about it too.

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -1,16 +1,16 @@
...
-import { RundownEventBuilder } from '../interfaces/rundown-event-builder'
+import { RundownEventBuilder } from '../../../presentation/interfaces/rundown-event-builder'
 import { ActiveRundownException } from '../../../model/exceptions/active-rundown-exception'
-import { RundownEventType } from '../../../model/enums/rundown-event-type'
+import { RundownEventType } from '../../../presentation/enums/rundown-event-type'
 import { RundownTimelineService } from '../rundown-timeline-service'
 import { CallbackScheduler } from '../interfaces/callback-scheduler'
 import { EntityMockFactory } from '../../../model/entities/test/entity-mock-factory'
-import { RundownDeletedEvent } from '../../../model/value-objects/rundown-event'
+import { RundownDeletedEvent } from '../../../presentation/value-objects/rundown-event'
```
LindvedKrvang, line 13: This is not allowed to know about the Presentation layer

LindvedKrvang (approved):
We need to be very aware of that now there is a single place in the service layers that knows about the presentation layer.

Else it looks good.

src/business-logic/services/test/rundown-timeline-service.spec.ts
KvelaGorrrrnio: I have removed the imports :+1:

PR #33, opened 2023-10-24
SOF-1672: Throw exception when trying to set active part as next.

KvelaGorrrrnio:
When trying to set an active part as next, an ActivePartException is thrown.


andr9528 (approved):

RasmusAlbrektsen (approved):

src/model/exceptions/active-part-exception.ts
```diff
@@ -0,0 +1,8 @@
+import { ErrorCode } from '../enums/error-code'
+import { Exception } from './exception'
+
+export class ActivePartException extends Exception {
```
LindvedKrvang, line 4: Suggestion: We could call it something like `AlreadyOnAirException`. Then the exception is a bit more reuseable and we don't risk as much having exceptions only use for a single specific things

src/model/entities/test/rundown.spec.ts
```diff
@@ -2805,12 +2804,48 @@ describe(Rundown.name, () => {
...
+          isRundownActive: true,
+          alreadyActiveProperties: {
+            activeCursor: {
+              segment: activeSegment,
+              part: activePart,
+            },
+            nextCursor: {
+              segment: nextSegment,
+              part: nextPart,
+            },
```
LindvedKrvang, line 2836: You don't need a `nextCursor` for this test so you can skip all the setup for the nextCursor.

src/model/entities/test/rundown.spec.ts
```diff
@@ -2805,12 +2804,48 @@ describe(Rundown.name, () => {
...
       testee.setNext(activeSegment.id, otherPartInActiveSegment.id)
 
       verify(mockedNextPart.reset()).once()
     })
+
+    describe('when next part is on air', () => {
+      it('throws an active part exception', () => {
+        const activePart: Part = EntityMockFactory.createPart({ id: 'active-part-id', isOnAir: true })
+        const mockedActiveSegment: Segment = EntityMockFactory.createSegmentMock({ id: 'active-segment-id', parts: [activePart] })
+        when(mockedActiveSegment.findPart(activePart.id)).thenReturn(activePart)
```
LindvedKrvang, line 2822: Instead of mocking you could also use the `EntityTestFactory`.

LindvedKrvang (approved):

src/model/exceptions/active-part-exception.ts
KvelaGorrrrnio: Do you want `ActivePartException` and `ActiveRundownException` merged together to `AlreadyOnAirException`?

src/model/entities/test/rundown.spec.ts
KvelaGorrrrnio: There is a check in rundown.setNext that throws if no nextCursor is set.

src/model/entities/test/rundown.spec.ts
KvelaGorrrrnio: Updated to using the EntityTestFactory.

src/model/exceptions/active-part-exception.ts
LindvedKrvang: Initial thoughs: no. A Rundown being active doesn't necessarily means that it is on Air. You can have an active Rundown that has no Parts on Air (when we activate the Rundown before doing a Take)

KvelaGorrrrnio: I have changed it to `OnAirException`. (ActiveRundown is still kept)

PR #35, opened 2023-10-25
SOF-1623: Piece rendering

KvelaGorrrrnio:
Sofie server changes:
- `Piece.type` is moved to `Piece.metadata.type` and is hereby changed into a TV2 blueprint specific attribute.
- TimeoutCallbackScheduler uses a logarithmic approach for scheduling a callback.

TV2 blueprints changes:
- The duration is set to undefined for unplanned video clips without a known duration.
- Output layer is added to `Piece.metadata`.
- `DSK` actions in TV2 blueprints have no output layer.


src/blueprints/tv2/entities/tv2-piece.ts
```diff
@@ -0,0 +1,6 @@
+import { PieceInterface } from '../../../model/entities/piece'
+import { Tv2PieceMetadata } from '../value-objects/tv2-metadata'
+
+export interface Tv2Piece extends PieceInterface {
```
LindvedKrvang, line 4: You are extending `PieceInterface`, so this should be `Tv2PieceInterface`.
Right now there are variables declared as `const pieceInterface: Tv2Piece` which leaves the question "why did they name it `PieceInterface` when it has the type of `Piece`?"

LindvedKrvang (changes requested):

src/blueprints/tv2/entities/tv2-piece.ts
KvelaGorrrrnio: I'll change the Tv2Piece type to be called Tv2PieceInterface, such that it matches, but pieceInterface is a bad variable name. It is not an interface, it is real data.

LindvedKrvang: It's the data needed to create an actual Piece entity.
I'm up for naming suggestions.

KvelaGorrrrnio: Currently the "best" suggestion is `PieceData`.

andr9528: I support that suggested name. It is 'Data' needed to create a 'Piece'. It makes sense, atleast for me.

src/blueprints/tv2/enums/tv2-piece-type.ts
```diff
@@ -0,0 +1,16 @@
+export enum Tv2PieceType {
+  CAMERA = 'CAMERA',
+  REPLAY = 'REPLAY',
+  REMOTE = 'REMOTE',
+  GRAPHICS = 'GRAPHICS',
+  OVERLAY_GRAPHICS = 'OVERLAY_GRAPHICS',
```
andr9528, line 6: Should e.g. an `Ident` use this one, or create a new one that is even more specific? i.e a `IDENT_OVERLAY_GRAPHICS`, or something along those lines.

src/blueprints/tv2/tv2-action-service.ts
```diff
@@ -84,4 +87,21 @@ export class Tv2ActionService implements BlueprintGenerateActions {
         }
       })
   }
+
+  private getPieceTypeFromMongoActionManifest(mongoActionManifest: ActionManifest): Tv2PieceType {
```
andr9528, line 91: It isn't a mongo type at this point.

KvelaGorrrrnio: Nicely spotted :) I have removed the mongo references.

src/blueprints/tv2/enums/tv2-piece-type.ts
KvelaGorrrrnio: I would say that it should use `OVERLAY_GRAPHICS`, and then if it needs further specification, then one could add a new metadata attribute `overlayGraphicsType`.

src/blueprints/tv2/entities/tv2-piece.ts
LindvedKrvang: I don't have any strong arguments for keeping the `xInterface` naming. Just remember to change it for `Rundown`, `Segment`, `Part` and `Piece` if you change it.

src/blueprints/tv2/tv2-action-service.ts
```diff
@@ -100,7 +100,7 @@ export class Tv2ActionService implements BlueprintGenerateActions {
         return Tv2PieceType.SPLIT_SCREEN
       }
       default: {
-        throw new UnexpectedCaseException(`Unknown MongoActionManifestId: ${mongoActionManifest.actionId}`)
+        throw new UnexpectedCaseException(`Unknown MongoActionManifestId: ${actionManifest.actionId}`)
```
LindvedKrvang, line 103: `MongoActionManifestId` should be changed to `ActionManifestId`

LindvedKrvang (approved):

andr9528 (approved):

src/blueprints/tv2/tv2-action-service.ts
KvelaGorrrrnio: Updated.

PR #37, opened 2023-10-26
ST-1668 Throw exception from backend when try to activate rundown whe…

temk-tv2:
Throw AlreadyActivatedException when trying to activate a rundown and other active rundown exists.


src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -55,6 +56,22 @@ describe(RundownTimelineService.name, () => {
 
       await expect(() => testee.deleteRundown(rundown.id)).rejects.toThrow(ActiveRundownException)
     })
+
+    it('throws an exception, when trying to active a rundown when there is another already activated rundown', async () => {
```
LindvedKrvang, line 60: You have put this test under the `describe` for the `deleteRundown` method.
You should make a new `describe` for the `activateRundown` method to put this test in.
Something like this:
```
describe(RundownTimelineService.name, () => {
  describe(RundownTimelineService.prototype.activateRundown.name, () => {
    // Your test here
  })

  describe(RundownTimelineService.prototype.deleteRundown.name, () => {
    ...
    // This is where you currently have your test.
  })
})
```

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -55,6 +56,22 @@ describe(RundownTimelineService.name, () => {
...
+      const mockRundownRepository: RundownRepository = mock<RundownRepository>()
+
+      const activeBasicRundown: Rundown = EntityMockFactory.createRundown({ isRundownActive: true })
+
+      const basicRundownArray: Rundown[] = [activeBasicRundown];
+
+      const rundownToActivate: Rundown = EntityMockFactory.createRundown({isRundownActive: false})
```
LindvedKrvang, line 67: Since you are making two different Rundowns it would be a good idea to also provide an id.
The current implementation of `EntityMockFactory.createRundown()` defaults the `rundownId` to be "rundownId" if no other id is provided. So right now you have made two different Rundowns with the same id.

All you would to do is add the id like this:
`EntityMockFactory.createRundown({ id: 'inactiveRundown', isRundownActive: false })`

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -55,6 +56,22 @@ describe(RundownTimelineService.name, () => {
...
+
+      when(mockRundownRepository.getBasicRundowns()).thenResolve(basicRundownArray)
+
+      const testee: RundownTimelineService = createTestee({rundownRepository: instance(mockRundownRepository)})
+
+      await expect(() => testee.activateRundown(rundownToActivate.name)).rejects.toThrow(AlreadyActivatedException)
```
LindvedKrvang, line 73: You should use `rundownToActivate.id`, not `rundownToActivate.name` since `activateRundown()` expects a Rundown id. 
Both id and name are strings, so that's why you don't get an error, but you should still give it the id :)

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -31,6 +33,13 @@ export class RundownTimelineService implements RundownService {
   ) {}
 
   public async activateRundown(rundownId: string): Promise<void> {
+
+    const basicRundowns: BasicRundown[] = await this.rundownRepository.getBasicRundowns();
+    const activeRundown: BasicRundown | undefined = basicRundowns? basicRundowns.find(rundown => rundown.isActive()) : undefined;
+
+    if (activeRundown) {
+      throw new AlreadyActivatedException(`Unable to activate rundown, because: ${activeRundown.name} is active. `)
+    }
```
LindvedKrvang, line 42: A nice thing you could do is extract this into a separate method called `assertNoRundownIsActive()`. 
Then you stick to single responsibility :)

LindvedKrvang (commented):
Looks good :)
Only a few comments.

LindvedKrvang:
FYI: Your build is failing with a lint error that needs to be addressed :)
image

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -31,6 +33,13 @@ export class RundownTimelineService implements RundownService {
   ) {}
 
   public async activateRundown(rundownId: string): Promise<void> {
+
+    const basicRundowns: BasicRundown[] = await this.rundownRepository.getBasicRundowns();
+    const activeRundown: BasicRundown | undefined = basicRundowns? basicRundowns.find(rundown => rundown.isActive()) : undefined;
```
KvelaGorrrrnio, line 38: The ternary is unnecessary :) If the the getBasicRundowns call resolves, then you are guaranteed to have an array of basic rundowns.

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -55,6 +56,22 @@ describe(RundownTimelineService.name, () => {
 
       await expect(() => testee.deleteRundown(rundown.id)).rejects.toThrow(ActiveRundownException)
     })
+
+    it('throws an exception, when trying to active a rundown when there is another already activated rundown', async () => {
+      const mockRundownRepository: RundownRepository = mock<RundownRepository>()
+
+      const activeBasicRundown: Rundown = EntityMockFactory.createRundown({ isRundownActive: true })
+
+      const basicRundownArray: Rundown[] = [activeBasicRundown];
```
KvelaGorrrrnio, line 65: Instead of naming it with what technical construct is used to hold the entitites, you can refer to the entities in plural such that this becomes: `const basicRundowns: Rundown = [activeBasicRundown]`.

src/business-logic/services/rundown-timeline-service.ts
LindvedKrvang, line 38: `this.rundownRepository.getBasicRundowns()` returns an array, so there is no need to do the undefined check on line 38.
You can just write: 
```
const activeRundown: BasicRundown | undefined = basicRundowns.find(rundown => rundown.isActive())
```

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -55,6 +56,22 @@ describe(RundownTimelineService.name, () => {
...
+
+      const rundownToActivate: Rundown = EntityMockFactory.createRundown({isRundownActive: false})
+
+      when(mockRundownRepository.getBasicRundowns()).thenResolve(basicRundownArray)
+
+      const testee: RundownTimelineService = createTestee({rundownRepository: instance(mockRundownRepository)})
+
+      await expect(() => testee.activateRundown(rundownToActivate.name)).rejects.toThrow(AlreadyActivatedException)
+    })
```
KvelaGorrrrnio, line 74: You can use newlines to group the test into different logical sections, like in the code snippet below, with a setup section, a section where you apply the functionality to test and an assertion section.
I have taken the liberty to move the variable instantiation of the mocked RundownRepository down to right before the `when` statement to have the reading flow match the execution flow as close as possible.
I have also moved the expression to test out into the variable  `result` to make that section explicit.

```typescript
    it('throws an exception, when trying to active a rundown when there is another already activated rundown', async () => {
      const activeBasicRundown: Rundown = EntityMockFactory.createRundown({ isRundownActive: true })
      const basicRundownArray: Rundown[] = [activeBasicRundown];
      const rundownToActivate: Rundown = EntityMockFactory.createRundown({isRundownActive: false})
      const mockedRundownRepository: RundownRepository = mock<RundownRepository>()
      when(mockedRundownRepository.getBasicRundowns()).thenResolve(basicRundownArray)
      const testee: RundownTimelineService = createTestee({rundownRepository: instance(mockedRundownRepository)})
      
      const result: () => Promise<void> = () => testee.activateRundown(rundownToActivate.name)
      
      await expect(result).rejects.toThrow(AlreadyActivatedException)
    })
```

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -31,6 +33,13 @@ export class RundownTimelineService implements RundownService {
...
+
+    if (activeRundown) {
+      throw new AlreadyActivatedException(`Unable to activate rundown, because: ${activeRundown.name} is active. `)
```
KvelaGorrrrnio, line 41: The colon in the error message is unnecessary. A suggestion would be to also mention what `activeRundown.name` is. E.g. `Unable to activate rundown, because the rundown ${activeRundown.name} is active.`

src/business-logic/services/test/rundown-timeline-service.spec.ts
temk-tv2: correct.

src/business-logic/services/test/rundown-timeline-service.spec.ts
temk-tv2: I have made the requested changes.

src/business-logic/services/test/rundown-timeline-service.spec.ts
temk-tv2: changes made

src/business-logic/services/rundown-timeline-service.ts
temk-tv2: I think I made what you suggest

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -12,6 +12,7 @@ import { Blueprint } from '../../../model/value-objects/blueprint'
 import { PartRepository } from '../../../data-access/repositories/interfaces/part-repository'
 import { SegmentRepository } from '../../../data-access/repositories/interfaces/segment-repository'
 import { PieceRepository } from '../../../data-access/repositories/interfaces/piece-repository'
+import {AlreadyActivatedException} from '../../../model/exceptions/already-activated-exception'
```
RasmusAlbrektsen, line 15: Space before and after the brackets would be nice :)  Such that it looks like this:`{ AlreadyActivatedException }`.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -16,6 +16,8 @@ import { PartRepository } from '../../data-access/repositories/interfaces/part-r
 import { Segment } from '../../model/entities/segment'
 import { SegmentRepository } from '../../data-access/repositories/interfaces/segment-repository'
 import { PieceRepository } from '../../data-access/repositories/interfaces/piece-repository'
+import {BasicRundown} from '../../model/entities/basic-rundown'
+import {AlreadyActivatedException} from '../../model/exceptions/already-activated-exception'
```
RasmusAlbrektsen, line 20: Space before and after the brackets would be nice :) Such that it looks like this:
{ BasicRundown }
{ AlreadyActivatedException }

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -45,6 +52,12 @@ export class RundownTimelineService implements RundownService {
     await this.rundownRepository.saveRundown(rundown)
   }
 
+  private assertNoRundownIsActive(activeRundown: BasicRundown | undefined): void {
```
KvelaGorrrrnio, line 55: This does not assert whether there is a rundown active. It asserts if a basic rundown is given to it.

LindvedKrvang (approved):

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -45,6 +49,13 @@ export class RundownTimelineService implements RundownService {
     await this.rundownRepository.saveRundown(rundown)
   }
 
+  private async assertNoRundownIsActive(): Promise<void> {
+    const activeRundown: BasicRundown | undefined = (await this.rundownRepository.getBasicRundowns()).find(rundown => rundown.isActive())
+    if (activeRundown) {
+      throw new AlreadyActivatedException(`Unable to activate rundown, because the rundown ${activeRundown.name} is active. `)
```
KvelaGorrrrnio, line 55: The trailing space in the message can be removed.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -31,6 +33,8 @@ export class RundownTimelineService implements RundownService {
   ) {}
 
   public async activateRundown(rundownId: string): Promise<void> {
+
```
KvelaGorrrrnio, line 36: The leading newline can be removed.

KvelaGorrrrnio (approved):
Nice work :rocket:

PR #38, opened 2023-10-31
Sof 1664/expose endpoint to reload ingest data

RasmusAlbrektsen:
This PR introduces an `IngestService` and renames the former one uses for ingest changes. The service makes use of a specific `got`-implementation of the new `HttpService` interface and calls the iNews Gateway in this manner. This works by utilizing the changes made in: https://github.com/tv2/inews-ftp-gateway/pull/88.


src/business-logic/services/tv2-inews-ingest-service.ts
```diff
@@ -0,0 +1,12 @@
+import { IngestService } from './interfaces/ingest-service'
+import {HttpRequestParameters, HttpService} from './interfaces/http-service'
+
+export class Tv2InewsIngestService implements IngestService {
```
andr9528, line 4: Should the name not be `Tv2INewsIngestService`?

src/presentation/facades/controller-facade.ts
```diff
@@ -6,14 +6,16 @@ import { BaseController } from '../controllers/base-controller'
 import { TimelineController } from '../controllers/timeline-controller'
 import { ActionController } from '../controllers/action-controller'
 import { ConfigurationController } from '../controllers/configuration-controller'
+import {IngestController} from '../controllers/ingest-controller'
```
andr9528, line 9: Has Lint been applied?

andr9528 (approved):

package.json
```diff
@@ -35,8 +35,10 @@
     "typescript": "^5.2.2"
   },
   "dependencies": {
+    "@types/got": "^9.6.12",
```
LindvedKrvang, line 38: If I'm not mistaken, this should be a devDependency

src/business-logic/services/tv2-inews-ingest-service.ts
LindvedKrvang: We spell "iNews" with lowercase `i` and uppercase `N`, so yeah, it should technically be `Tv2INewsIngestService`.

src/business-logic/services/tv2-inews-ingest-service.ts
```diff
@@ -0,0 +1,12 @@
...
+  constructor(private readonly httpService: HttpService) {}
+  public async reloadIngestData(rundownName: string): Promise<void> {
```
LindvedKrvang, line 6: Be nice with a newline between each method. That includes the constructor.

src/business-logic/services/tv2-inews-ingest-service.ts
```diff
@@ -0,0 +1,12 @@
...
+    const domain = process.env.INEWS_DOMAIN ?? 'localhost:3007'
+    const url = `http://${domain}/reloadData/${rundownName}`
```
LindvedKrvang, line 8: This should technically be on the iNews-gateway PR, but we are not following REST standards here with this URL.
We can't follow the standard completely since we are not using an id for the Rundown, but the URL should either be `/reloadData/rundowns/{rundownName}` or `/rundowns/{rundownName}/reloadData` dependeing on how you view the `reloadData` operation.

src/business-logic/services/tv2-inews-ingest-service.ts
```diff
@@ -0,0 +1,12 @@
+import { IngestService } from './interfaces/ingest-service'
+import {HttpRequestParameters, HttpService} from './interfaces/http-service'
+
+export class Tv2InewsIngestService implements IngestService {
+  constructor(private readonly httpService: HttpService) {}
+  public async reloadIngestData(rundownName: string): Promise<void> {
+    const domain = process.env.INEWS_DOMAIN ?? 'localhost:3007'
```
LindvedKrvang, line 7: Magic string.
You could extract it into a variable called something liked `const DEFAULT_URL: string = 'localhost:3007'`

src/presentation/controllers/ingest-controller.ts
```diff
@@ -0,0 +1,24 @@
+import { BaseController, PostRequest, RestController } from './base-controller'
+import { HttpErrorHandler } from '../interfaces/http-error-handler'
+import { Exception } from '../../model/exceptions/exception'
+import { Request, Response } from 'express'
+import { IngestService } from '../../business-logic/services/interfaces/ingest-service'
+
+@RestController('/ingest')
```
LindvedKrvang, line 7: This is a quite interesting scenario.
To follow REST standards we should use plural forms, but plural of `ingest` doesn't make any sense.
Maybe this shows that the right controller might be the `RundownController` with a URL of `rundowns/{rundownId}/reloadData`?

src/presentation/controllers/ingest-controller.ts
```diff
@@ -0,0 +1,24 @@
...
+export class IngestController extends BaseController {
+
+  constructor(private readonly ingestService: IngestService, private readonly httpErrorHandler: HttpErrorHandler) {
+    super()
+  }
+
+  @PostRequest('/reloadData/:rundownName')
```
LindvedKrvang, line 14: This is the endpoint for clients to call SofieServer. In SofieServer we use Ids.
This should be using `rundownId` and not `rundownName`.

LindvedKrvang: And to follow REST we need to prepend the plural form of the entity of which we are parsing an id.
So it should be `reloadData/rundowns/{rundownId}`

src/presentation/controllers/ingest-controller.ts
```diff
@@ -0,0 +1,24 @@
...
+  public async reloadRundownData(req: Request, res: Response): Promise<void> {
+    try {
+      const rundownName: string = req.params.rundownName
+      await this.ingestService.reloadIngestData(rundownName)
+      res.send(`Attempted reload of rundown data for ${rundownName}`)
```
LindvedKrvang, line 19: Why 'attemped'?

src/presentation/facades/controller-facade.ts
RasmusAlbrektsen: I don't think the rules you are thinking about are enforced on sofie-server yet :-)

src/business-logic/services/tv2-inews-ingest-service.ts
RasmusAlbrektsen: I have changed the naming, good catch :-)

src/business-logic/services/tv2-inews-ingest-service.ts
RasmusAlbrektsen: This has been changed!

package.json
RasmusAlbrektsen: Absolutely right, a mishap on my behalf! I have moved it.

src/presentation/controllers/ingest-controller.ts
RasmusAlbrektsen: I was actually not sure what to say here, all we know is that we sent the signal to try and reload the data. We don't have a way of ensuring that happened, just that we "attempted" it.

src/business-logic/services/tv2-inews-ingest-service.ts
RasmusAlbrektsen: I have changed it to use `/reloadData/rundowns/{rundownName}`.

src/business-logic/services/tv2-inews-ingest-service.ts
RasmusAlbrektsen: Moved to constant!

src/presentation/controllers/ingest-controller.ts
RasmusAlbrektsen: I agree this is interesting. I wanted a new controller since the ingest is not directly a manipulation of a rundown. Triggering this action does not directly signify that we alter the rundown, just that trigger the ingest to reload the data. For that reason i kept it seperate. I see the point about `ingest` but i don't have a better suggestion for now...

src/presentation/controllers/ingest-controller.ts
RasmusAlbrektsen: It has been changed to follow this!

src/presentation/controllers/ingest-controller.ts
LindvedKrvang: I'm also in favour of having the `IngestController`.
I'm not sure if it is better, but we could call it `ingestOperations` or something similar?
Or we could just keep it as is. Just not the biggest fan of not following REST conventions in a REST endpoint.

src/business-logic/services/tv2-inews-ingest-service.ts
```diff
@@ -1,11 +1,18 @@
...
-  public async reloadIngestData(rundownName: string): Promise<void> {
-    const domain = process.env.INEWS_DOMAIN ?? 'localhost:3007'
-    const url = `http://${domain}/reloadData/${rundownName}`
+const DEFAULT_URL: string = 'localhost:3007'
+
+export class Tv2INewsIngestService implements IngestService {
+
+  constructor(private readonly httpService: HttpService, private readonly rundownRepository: RundownRepository) {}
+
+  public async reloadIngestData(rundownId: string): Promise<void> {
```
LindvedKrvang, line 12: This method is getting a bit "clamped".
You are also technically mixing abstractions levels.
A method called `getIngestUrl()` would help it tremendiously.
```
public async reloadIngestData(...): ... {
  const rundown: Rundown = await this.rundownRepository.getRundown(...)
  const url: string = this.getIngestUrl()
  const parameters: HttpRequestParameters = { url: url }
  await this.httpService.post(parameters)
}
```

LindvedKrvang (approved):

src/presentation/controllers/ingest-controller.ts
KvelaGorrrrnio: I have put a comment on the iNews gateway PR regarding, where i point out that the response should be sent after the reload/resync is done. If that is the case, then we know whether or not it has reloaded at this point.

src/presentation/controllers/ingest-controller.ts
KvelaGorrrrnio: Reloading the data for a rundown is an action on the rundown. The frontend(s) (probably) shouldn't be aware about ingest data. From the frontend(s) perspective, it/they are still doing an operation on a rundown. When we get to actually receiving data from our ingest gateway, then an ingest controller, which the ingest gateway interacts with, will be appropriate.

src/business-logic/services/tv2-inews-ingest-service.ts
RasmusAlbrektsen: I have moved it to a private method! Good suggestion.

src/presentation/controllers/ingest-controller.ts
RasmusAlbrektsen: I have changed this to reflect that we have actually triggered reingest since we now know this with the latest changes to the iNews Gateway.

src/presentation/controllers/ingest-controller.ts
RasmusAlbrektsen: I have changed the URL to be `ingest-operations` instead, let me know if this is not what you wanted. 
Also, i see your point @KvelaGorrrrnio, but is triggering the data reload from the gateway really an action on the rundown itself in the same way as activating it is? If we'd rather have it inside the `RundownController`, just let me know and i will change it!

src/business-logic/services/got-http-service.ts
```diff
@@ -0,0 +1,10 @@
+import {HttpRequestParameters, HttpService} from './interfaces/http-service'
+import got, { Response } from 'got'
+
+export class GotHttpService implements HttpService {
+
+  public async post(httpParameters: HttpRequestParameters): Promise<Response<string>> {
```
KvelaGorrrrnio, line 6: It seems a bit overkill to have an interface for this. We can just take two parameters. The URL and the body.

src/business-logic/services/tv2-inews-ingest-service.ts
```diff
@@ -0,0 +1,23 @@
+import { IngestService } from './interfaces/ingest-service'
+import { HttpRequestParameters, HttpService } from './interfaces/http-service'
+import { RundownRepository } from '../../data-access/repositories/interfaces/rundown-repository'
+import { Rundown } from '../../model/entities/rundown'
+
+const DEFAULT_URL: string = 'localhost:3007'
```
KvelaGorrrrnio, line 6: This is not a url. And since the env variables don't change after this point, you check it here, such that this becomes:
```typescript
const INEWS_HOST: string = process.env.INEWS_HOST ?? 'localhost:3007'
```

KvelaGorrrrnio (approved):
I am approving under the assumption, that the below comments are addressed.

src/presentation/controllers/ingest-controller.ts
```diff
@@ -0,0 +1,24 @@
+import { BaseController, PostRequest, RestController } from './base-controller'
+import { HttpErrorHandler } from '../interfaces/http-error-handler'
+import { Exception } from '../../model/exceptions/exception'
+import { Request, Response } from 'express'
+import { IngestService } from '../../business-logic/services/interfaces/ingest-service'
+
+@RestController('/ingest-operations')
```
KvelaGorrrrnio, line 7: Is the ingest-operations needed? Won't it suffice with the url `/rundowns/:rundownId/reingest`?
And if so, is it necessary with a ingest controller?

src/presentation/controllers/ingest-controller.ts
```diff
@@ -0,0 +1,24 @@
...
+export class IngestController extends BaseController {
+
+  constructor(private readonly ingestService: IngestService, private readonly httpErrorHandler: HttpErrorHandler) {
+    super()
+  }
+
+  @PostRequest('/rundowns/:rundownName/reingest-data')
```
KvelaGorrrrnio, line 14: `-data` is a fill word here. `reingest` should be enough to tell what this does.

KvelaGorrrrnio (commented):
I found some extra things :)

src/business-logic/services/got-http-service.ts
RasmusAlbrektsen: It has been removed and replaced with `url` and `body`.

src/business-logic/services/tv2-inews-ingest-service.ts
RasmusAlbrektsen: I have changed the constant to fit this!

src/presentation/controllers/ingest-controller.ts
RasmusAlbrektsen: Good suggestion, this now resides in the `RundownController` instead.

src/presentation/controllers/ingest-controller.ts
RasmusAlbrektsen: Filler-word has been removed!

LindvedKrvang (approved):

src/business-logic/services/got-http-service.ts
```diff
@@ -0,0 +1,9 @@
+import { HttpService } from './interfaces/http-service'
+import got from 'got'
+
+export class GotHttpService implements HttpService {
+
+  public post(url: string, body: unknown): unknown {
+    return got.post(url, body ?? undefined)
```
KvelaGorrrrnio, line 7: Is the nullish coalescing necessary? And in this implementation I can't send `null` as the body.
Instead the body parameter could default to undefined. That way you don't have to specify it when calling it.

KvelaGorrrrnio (approved):

src/business-logic/services/got-http-service.ts
RasmusAlbrektsen: I've made the change you suggested :-)

PR #39, opened 2023-11-01
SOF-1549: Tv2CameraActionFactory uses Tv2VideoMixerTimelineObjectFactory to create timeline objects

KvelaGorrrrnio:
- Moved subtypes of DVE related content type values to the enum `Tv2ActionSubtype`.
- Tv2CameraActionFactory uses Tv2VideoMixerTimelineObjectFactory to create timeline objects.


src/blueprints/tv2/timeline-object-factories/interfaces/tv2-video-mixer-timeline-object-factory.ts
```diff
@@ -6,9 +6,9 @@ import { Tv2BlueprintConfiguration } from '../../value-objects/tv2-blueprint-con
 import { Tv2BlueprintTimelineObject } from '../../value-objects/tv2-metadata'
 
 export interface Tv2VideoMixerTimelineObjectFactory {
-  createProgramTimelineObject(sourceInput: number, enable: TimelineEnable): TimelineObject
+  createProgramTimelineObject(sourceInput: number, enable: TimelineEnable, id?: string): TimelineObject
```
LindvedKrvang, line 9: Is the optional `id` parameter really necessary?
Also, having the `id` parameter as  the first parameter would be nice and would be consistent with the rest of the codebase. That however makes it difficult to make it optional.

RasmusAlbrektsen (approved):

src/blueprints/tv2/timeline-object-factories/interfaces/tv2-video-mixer-timeline-object-factory.ts
KvelaGorrrrnio: The `id` parameters are required and are rearranged to be the first parameter of the methods.

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
```diff
@@ -0,0 +1,120 @@
...
+        ...videoMixerTimelineObjects,
+        ...audioTimelineObjects
+      ]
+    }
+  }
+
+  private createVideoMixerTimelineObjects(source: Tv2SourceMappingWithSound): TimelineObject[] {
+    const enable: TimelineEnable = { start: 0 }
+    return [
+      this.videoMixerTimelineObjectFactory.createProgramTimelineObject(`insertedProgram_${source._id}`, source.SwitcherSource, enable),
```
LindvedKrvang, line 94: You are missing a Cleanfeed TimelineObject.

LindvedKrvang (approved):
The approve assumes that you add the Cleanfeed TimelineObject as well.

src/blueprints/tv2/action-factories/tv2-dve-action-factory.ts
```diff
@@ -69,32 +73,33 @@ export class Tv2DveActionFactory {
...
-      Tv2ActionContentType.RECALL_DVE,
-      Tv2ActionContentType.DVE_INSERT_LAST_VIDEO_CLIP_TO_INPUT
-    ].includes(tv2Action.metadata.contentType)
+    const actionSubtype: Tv2ActionSubtype | undefined = tv2Action.metadata.actionSubtype
+    return actionSubtype !== undefined && [
+      Tv2ActionSubtype.DVE_LAYOUT,
+      Tv2ActionSubtype.DVE_INSERT_SOURCE_TO_INPUT,
+      Tv2ActionSubtype.RECALL_DVE,
+      Tv2ActionSubtype.DVE_INSERT_LAST_VIDEO_CLIP_TO_INPUT
+    ].includes(actionSubtype)
```
andr9528, line 82: The array looks to contain all the values of the used enum. 
Why not just get the enum values as a `string[]`, straight from the enum instead of creating an array with them?

andr9528 (approved):

src/blueprints/tv2/action-factories/tv2-dve-action-factory.ts
LindvedKrvang: They won't continue to use all values of the enum. Right now it's just DVE that has added entries to it.
Your graphics Actions will also add entries to this enum.

PR #40, opened 2023-11-02
SOF-1628: Insert remote source as next actions

KvelaGorrrrnio:
Actions for inserting remote/live sources as next are being generated.

### DIFF
--- a/src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
+++ b/src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
@@ -1,0 +1,120 @@
+import { PieceInterface } from '../../../model/entities/piece'
+import { PartInterface } from '../../../model/entities/part'
+import { PartActionType } from '../../../model/enums/action-type'
+import { Tv2BlueprintConfiguration } from '../value-objects/tv2-blueprint-configuration'
+import { Tv2SourceMappingWithSound } from '../value-objects/tv2-studio-blueprint-configuration'
+import { TimelineObject } from '../../../model/entities/timeline-object'
+import { Tv2PieceMetadata } from '../value-objects/tv2-metadata'
+import { PieceType } from '../../../model/enums/piece-type'
+import { Tv2SourceLayer } from '../value-objects/tv2-layers'
+import { PieceLifespan } from '../../../model/enums/piece-lifespan'
+import { TransitionType } from '../../../model/enums/transition-type'
+import { Tv2ActionContentType, Tv2RemoteAction } from '../value-objects/tv2-action'
+import {
+  Tv2AudioTimelineObjectFactory
+} from '../timeline-object-factories/interfaces/tv2-audio-timeline-object-factory'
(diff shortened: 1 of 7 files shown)


src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
```diff
@@ -0,0 +1,120 @@
...
+      metadata: {
+        contentType: Tv2ActionContentType.REMOTE,
+        remoteNumber: remoteSource.SourceName,
+      },
+    }
+  }
+
+  private createRemotePiece(configuration: Tv2BlueprintConfiguration, source: Tv2SourceMappingWithSound, parentPartId: string): PieceInterface {
+    const remoteTimelineObjects: TimelineObject[] = this.createRemoteTimelineObjects(source)
+    const audioTimelineObjects: TimelineObject[] = this.audioTimelineObjectFactory.createTimelineObjectsForSource(configuration, source)
```
LindvedKrvang, line 56: It's a bit weird that you have a method called `createRemoteTimelineObjects()` but you have to create audioTimelineObjects right after calling it because `createRemoteTimelineObjects()` does not create all the timelineObjects for the Remote.

Either, you should move the creation of the audioTimelineObjects into `createRemoteTimelineObjects()` or you should rename `createRemoteTimelineObjects()` to `createVideoMixerTimelineObjects()` since that is what is does.

src/blueprints/tv2/value-objects/tv2-action.ts
```diff
@@ -48,6 +49,13 @@ export interface Tv2CameraAction extends Tv2PartAction {
   }
 }
 
+export interface Tv2RemoteAction extends Tv2PartAction {
+  metadata: {
+    contentType: Tv2ActionContentType.REMOTE,
+    remoteNumber: string
```
LindvedKrvang, line 55: I assume this is for something in the frontend?

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/blueprints/tv2/value-objects/tv2-action.ts
KvelaGorrrrnio: Currently it is what is used to determine what key binding to assign the remote action to. When we get a customizable key binding framework, this can be removed.

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
KvelaGorrrrnio: I have opted for `createVideoMixerTimlineObjects`. It is also updated in the camera action factory.

LindvedKrvang (approved):

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
LindvedKrvang: Nicely spotted for Cameras as well :)

PR #42, opened 2023-11-07
SOF-1694 Able to replace Planned Pieces

LindvedKrvang:
SofieServer now allows to replace Planned Pieces on a Part.


src/model/entities/part.ts
```diff
@@ -292,7 +294,11 @@ export class Part {
   public reset(): void {
     this.executedAt = 0
     this.playedDuration = 0
-    this.pieces = this.pieces.filter(piece => piece.isPlanned)
+    this.pieces = [
+      ...this.pieces.filter(piece => piece.isPlanned),
+      ...this.replacedPlannedPieces
+    ]
```
KvelaGorrrrnio, line 300: Does the order of the pieces matter?

KvelaGorrrrnio (approved):
Besides the question below, I just want to point out that replaced pieces won't survive a crash or restart.

src/model/entities/part.ts
LindvedKrvang: For Pieces, no, the order doesn't matter. They don't have any attributes to indicate order.
Pieces have a `start` and `duration` and are "just placed" on the Timeline. Then the TimelineObjects battle it out to see who wins.

RasmusAlbrektsen (approved):

andr9528 (approved):

PR #43, opened 2023-11-07
SOF-1629 Move Replay Actions

LindvedKrvang:
- Creates Actions to insert Replay sources into DVE inputs
- Create Replay Actions both with and without Voice Over
- Create Replay as Studio Aux Action
- Create Replay as Viz Aux Action

Update: Created 'it compiles' test files for all ActionFactories so our test scripts will catch compilation errors in them


src/blueprints/tv2/action-factories/tv2-dve-action-factory.ts
```diff
@@ -215,20 +216,24 @@ export class Tv2DveActionFactory {
...
       ...this.createInsertToInputActionsForSources(blueprintConfiguration, cameraSources, CAMERA_SOURCE_NAME),
-      ...this.createInsertToInputActionsForSources(blueprintConfiguration, liveSources, LIVE_SOURCE_NAME)
+      ...this.createInsertToInputActionsForSources(blueprintConfiguration, liveSources, LIVE_SOURCE_NAME),
+      ...this.createInsertToInputActionsForSources(blueprintConfiguration, replaySources, `${REPLAY_SOURCE_NAME} VO`, true),
+      ...this.createInsertToInputActionsForSources(blueprintConfiguration, replaySourcesWithoutVoiceOver, REPLAY_SOURCE_NAME, false)
     ]
   }
 
-  private createInsertToInputActionsForSources(blueprintConfiguration: Tv2BlueprintConfiguration, sources: Tv2SourceMappingWithSound[], name: string): Tv2DveInsertSourceInputAction[] {
+  private createInsertToInputActionsForSources(blueprintConfiguration: Tv2BlueprintConfiguration, sources: Tv2SourceMappingWithSound[], name: string, isVoiceOver: boolean = false): Tv2DveInsertSourceInputAction[] {
```
KvelaGorrrrnio, line 226: This looks like a boolean flag :) We can consider using an enumeration for the audio mode (or what the term that describes the data best is).

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -0,0 +1,236 @@
...
+        ]
+      },
+      metadata: {
+        contentType: Tv2ActionContentType.REPLAY
+      }
+    }
+  }
+
+  private createReplayActionWithoutVoiceOver(configuration: Tv2BlueprintConfiguration, source: Tv2SourceMappingWithSound): Tv2ReplayAction {
+    const trimmedName: string = source.SourceName.replace(' ', '')
```
KvelaGorrrrnio, line 72: If it is leading or trailing whitespace that should be removed, then `String.trim` can be used. Otherwise, this should be extracted to a separate method, since it is being used a lot in this class.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -0,0 +1,236 @@
...
+      preRollDuration: 0,
+      postRollDuration: 0,
+      metadata,
+      tags: [],
+      isUnsynced: false,
+      timelineObjects
+    }
+  }
+
+  // TODO: Verify behaviour of Action infinite Pieces.
```
KvelaGorrrrnio, line 156: Has this been verified?

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -0,0 +1,236 @@
...
+        delayPiecesDuration: 0
+      },
+      outTransition: {
+        keepAliveDuration: 0
+      },
+      disableNextInTransition: false
+    }
+  }
+
+  private createReplayForSourcePieceInterface(configuration: Tv2BlueprintConfiguration, parentPartId: string, source: Tv2SourceMappingWithSound, isVoiceOver: boolean): PieceInterface {
```
KvelaGorrrrnio, line 116: The PieceInterface types should be `Tv2PieceInterface`.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -0,0 +1,236 @@
+import { Action } from '../../../model/entities/action'
+import { Tv2BlueprintConfiguration } from '../value-objects/tv2-blueprint-configuration'
+import { Tv2SourceMappingWithSound } from '../value-objects/tv2-studio-blueprint-configuration'
+import { PartActionType, PieceActionType } from '../../../model/enums/action-type'
+import { PartInterface } from '../../../model/entities/part'
+import { PieceInterface } from '../../../model/entities/piece'
+import { Tv2ActionContentType, Tv2ReplayAction, Tv2ReplayAuxAction } from '../value-objects/tv2-action'
+import { PieceType } from '../../../model/enums/piece-type'
```
KvelaGorrrrnio: This file does not exist any more.

src/blueprints/tv2/value-objects/tv2-layers.ts
```diff
@@ -25,7 +28,12 @@ export enum Tv2AtemLayer {
   LOOKAHEAD = 'atem_aux_lookahead',
   DOWNSTREAM_KEYER = 'atem_dsk',
   DVE = 'atem_dve',
-  DVE_BOXES = 'atem_dve_boxes'
+  DVE_BOXES = 'atem_dve_boxes',
+}
+
+export enum Tv2VideoMixerLayer {
+  AR = 'aux_ar',
```
KvelaGorrrrnio, line 35: What is AR?

src/blueprints/tv2/value-objects/tv2-layers.ts
```diff
@@ -25,7 +28,12 @@ export enum Tv2AtemLayer {
...
+  VIZ_OVL_AUX = 'aux_viz_ovl_in_1'
```
KvelaGorrrrnio, line 36: OVL and AUX could be fully written out such that this becomes `VIZ_OVERLAY_AUXILIARY`. Same goes for other enum keys.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
LindvedKrvang: It's for changing `EVS 1` to `EVS1`

src/blueprints/tv2/value-objects/tv2-layers.ts
LindvedKrvang: Augmented Reality? Else no clue. This is what it is configured as...
image

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
LindvedKrvang: Hence why it is deleted :) 

The compile error wasn't caugt since we have no test file testing Actions, so they are no being executed for tests. I have added test files for all ActionFactories to avoid that in the future.

src/blueprints/tv2/action-factories/tv2-dve-action-factory.ts
LindvedKrvang: I agree, it looks like a boolean flag and I'm not really a fan of it.
Technically it is not because it just sets a boolean value, but it's too easy to use it as boolean flag, so I do like your idea.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
LindvedKrvang: Yes, I just forgot to remove the TODO

LindvedKrvang: It's been removed

src/blueprints/tv2/action-factories/tv2-dve-action-factory.ts
LindvedKrvang: Updated with `Tv2AudioLevel`

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
LindvedKrvang: Extracted into a method more aptly named

src/blueprints/tv2/value-objects/tv2-layers.ts
LindvedKrvang: Updated the Viz OVL entry.

src/blueprints/tv2/value-objects/tv2-layers.ts
KvelaGorrrrnio: I think you are right. But do we use it? I don't think we integrate to zero density or unreal.
Perhaps we should call it `AUGMENTED_REALITY_AUXILIARY`, and then change it if we become wiser.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -0,0 +1,240 @@
...
+      },
+      metadata: {
+        contentType: Tv2ActionContentType.REPLAY
+      }
+    }
+  }
+
+  private removeAllWhitespace(value: string): string {
+    return value.replace(' ', '')
+  }
```
KvelaGorrrrnio, line 74: This does not remove all whitespace. It removes all space characters :) tabs, newlines, etc. are kept. If you want to remove all whitespace you can use:
```suggestion
  private removeAllWhitespace(value: string): string {
    return value.replace(/\s+/g, '')
  }
```

src/blueprints/tv2/timeline-object-factories/interfaces/tv2-audio-timeline-object-factory.ts
```diff
@@ -3,9 +3,10 @@ import { Tv2BlueprintConfiguration } from '../../value-objects/tv2-blueprint-con
 import { Tv2VideoClipManifestData } from '../../value-objects/tv2-action-manifest-data'
 import { Tv2SourceMappingWithSound } from '../../value-objects/tv2-studio-blueprint-configuration'
 import { DeviceType } from '../../../../model/enums/device-type'
+import { Tv2AudioMode } from '../../enums/tv2-audio-mode'
 
 export interface Tv2AudioTimelineObjectFactory {
-  createTimelineObjectsForSource(configuration: Tv2BlueprintConfiguration, source: Tv2SourceMappingWithSound): TimelineObject[]
+  createTimelineObjectsForSource(configuration: Tv2BlueprintConfiguration, source: Tv2SourceMappingWithSound, audioMode?: Tv2AudioMode): TimelineObject[]
```
KvelaGorrrrnio, line 9: Why is `audioMode` optional?

src/blueprints/tv2/value-objects/tv2-layers.ts
```diff
@@ -1,13 +1,16 @@
...
   JINGLE = 'studio0_jingle',
   AUDIO_BED = 'studio0_audio_bed',
   GRAPHICS_ACTION_COMMAND = 'studio0_adlib_graphic_cmd',
   AUDIO_ACTION_COMMAND = 'studio0_sisyfos_adlibs',
   DOWNSTREAM_KEYER_ACTION_COMMAND = 'studio0_dsk', // Original one from Blueprint includes the number of the DSK, followed by '_cmd'.
   DVE = 'dve',
-  REMOTE = 'studio0_live',
+  REPLAY = 'studio0_local',
+  REPLAY_STUDIO_AUX = 'studio0_aux_studio_screen',
+  REPLAY_VIZ_AUX = 'studio0_aux_viz_full1',
```
KvelaGorrrrnio, line 13: Would be nice with the AUXs being expanded to `AUXILIARY`.

src/blueprints/tv2/value-objects/tv2-layers.ts
KvelaGorrrrnio: I think you are right. But do we use it? I don't think we integrate to zero density or unreal.
Perhaps we should call it `AUGMENTED_REALITY_AUXILIARY`, and then change it if we become wiser.

KvelaGorrrrnio (approved):

src/blueprints/tv2/value-objects/tv2-layers.ts
LindvedKrvang: I think they might be used.
If Jacob is correct, then it's used to display the REPLAY feed to the commentators

src/blueprints/tv2/timeline-object-factories/interfaces/tv2-audio-timeline-object-factory.ts
LindvedKrvang: It's far more common to not use VoiceOver. Since `AudioMode` is used to indicate whether or not it should be VoiceOver, I thought nice not having to include everytime we make TimelineObjects for a Source when we don't care for VoiceOver.

I don't have a strong argument to keep it optional, so if it's preferred that it isn't optional, I'm up for it :)

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -0,0 +1,240 @@
...
+
+  private createReplayActionWithVoiceOver(configuration: Tv2BlueprintConfiguration, source: Tv2SourceMappingWithSound): Tv2ReplayAction {
+    const noWhitespaceName: string = this.removeAllWhitespace(source.SourceName)
+    const partId: string = `${noWhitespaceName}_VO_part_action`
+    const partInterface: PartInterface = this.createPartInterface(partId, `Replay Part ${source.SourceName} VO`)
+    const pieceInterface: Tv2PieceInterface = this.createReplayForSourcePieceInterface(configuration, partId, source, Tv2AudioMode.VOICE_OVER)
+
+    return {
+      id: `insert_${noWhitespaceName}_VO_as_next_part_action`,
```
andr9528, line 56: White spaces aren't the only potentially bad characters in the id. 
Don't know how the content of `source.SourceName` looks, but e.g '/' is also a bad character.

For graphics actions i created a helper class which hashes a string, for use in id's. 
Perhabs that should eventually be used here?

src/blueprints/tv2/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
```diff
@@ -257,4 +256,22 @@ export class Tv2AtemVideoMixerTimelineObjectFactory implements Tv2VideoMixerTime
     }
     return this.createTransitionEffectTimelineObject(meContent)
   }
+
+  public createAuxTimelineObject(sourceInput: number, layer: Tv2VideoMixerLayer): AtemAuxTimelineObject {
+    return {
+      id: `${ATEM_PREFIX}${layer}_input_${sourceInput}_timelineObject`,
```
andr9528, line 262: Any reason why this one is suffixed `_timelineObject`, when other timeline objects created in this file does not suffix that to their id's?

src/blueprints/tv2/value-objects/tv2-action-manifest-data.ts
```diff
@@ -59,7 +60,7 @@ export interface Tv2VideoClipManifestData {
   fileName: string // userData.partDefinition.videoId
   durationFromIngest: number // userData.duration
   adLibPix: boolean // userData.adLibPix // What does "adLibPix" mean?
-  isVoiceOver: boolean // userData.voLevels
+  audioMode: Tv2AudioMode // userData.voLevels
```
andr9528, line 63: Should the comment still be there?

LindvedKrvang: You could argue yes and no. Yes, it still correspond to the `voLevels` of the `userData` (albeit it's a different type now). No, but then we should remove all the commets (which might be a reasonable thing to do).

src/blueprints/tv2/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
LindvedKrvang: No specific reason. We haven't been consistent in formatting our ids for Actions and TimelineObjects...

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
LindvedKrvang: These are Source names. If they start putting random characters in here, they are doing something wrong.

src/blueprints/tv2/value-objects/tv2-layers.ts
LindvedKrvang: Updated

src/blueprints/tv2/value-objects/tv2-layers.ts
LindvedKrvang: We could change it to `AC` and then we could call it `AUXILIARY_CLINTON` ;)

src/blueprints/tv2/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
andr9528: Might be worth looking into at some point. Not the end of the world for now, would just be good eventually.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
andr9528: Can you be 100% sure that the name doesn't contain illigal character?
Don't want to talk ill about our users, but isn't the general concensus in programming that you should expect your users to acts like monkey?

andr9528 (approved):

PR #44, opened 2023-11-07
Recall last live action.

KvelaGorrrrnio:
Adds action for recalling the last taken live part.


src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
```diff
@@ -17,15 +23,23 @@ import {
...
   ) {}
 
   public createRemoteActions(blueprintConfiguration: Tv2BlueprintConfiguration): Tv2RemoteAction[] {
+    return [
+      ...this.createInsertInsertRemoteAsNextActions(blueprintConfiguration),
+      this.createRecallLastRemoteAsNextAction(),
+    ]
+  }
+
+  private createInsertInsertRemoteAsNextActions(blueprintConfiguration: Tv2BlueprintConfiguration): Tv2RemoteAction[] {
```
LindvedKrvang, line 42: Insertception :D

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
```diff
@@ -116,4 +131,108 @@ export class Tv2RemoteActionFactory {
...
+      },
+      data: {
+        partInterface: {} as PartInterface, // Is determined when called.
+        pieceInterfaces: [] as PieceInterface[], // Is determined when called.
+      }
+    }
+  }
+
+  public isRemoteAction(action: Action): action is Tv2RemoteAction {
+    const tv2Action: Tv2Action = action as Tv2Action
```
LindvedKrvang, line 154: You can just have the parameter be `Tv2Action`. Then you don't need to cast :)

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
```diff
@@ -116,4 +131,108 @@ export class Tv2RemoteActionFactory {
...
+    return [
+      {
+        type: MutateActionType.HISTORIC_PART,
+        updateActionWithPartData: this.updateInsertLastVideoClipToInputAction.bind(this),
+        partPredicate: (part: Part) => this.doesPartContainARemotePiece(part),
+      }
+    ]
+  }
+
+  private updateInsertLastVideoClipToInputAction(action: Action, historicPart: Part): Tv2RecallLastRemoteAsNextAction {
```
LindvedKrvang, line 178: I think you forgot to update the name :)

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
```diff
@@ -116,4 +131,108 @@ export class Tv2RemoteActionFactory {
...
+  }
+
+  private doesPartContainARemotePiece(part: Part): boolean {
+    return part.getPieces().some(this.isRemotePiece.bind(this))
+  }
+
+  private isRemotePiece(piece: Piece): boolean {
+    const metadata: Tv2PieceMetadata | undefined = piece.metadata as Tv2PieceMetadata | undefined
+    return metadata?.type === Tv2PieceType.REMOTE
+  }
```
LindvedKrvang, line 237: Do you really need to bind `this`? You are not using `this` in `isRemotePiece`.

LindvedKrvang, line 237: We are only allowed to recall the last _planned_ Live.
I don't see you checking for that anywhere?

src/blueprints/tv2/value-objects/tv2-action.ts
```diff
@@ -60,6 +61,13 @@ export interface Tv2RemoteAction extends Tv2PartAction {
   }
 }
 
+export interface Tv2RecallLastRemoteAsNextAction extends Tv2RemoteAction {
+  type: PartActionType.INSERT_PART_AS_NEXT,
+  metadata: Tv2RemoteAction['metadata'] & {
+    actionSubtype: Tv2ActionSubtype.RECALL_LAST_REMOTE,
+  }
```
LindvedKrvang, line 68: Couldn't this just be:
```
metadata: {
  actionSubtype: Tv2ActionSubtype:RECALL_LAST_REMOTE
}
```

You set the value of `remoteNumber` to `-1` with the comment `// Is not used for this action`. 
If the attributes of `Tv2RemoteAction['metada']` is not used for this Action, then why do the type gymnastics to include them?

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
KvelaGorrrrnio: Corrected :D

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
KvelaGorrrrnio: Thanks – updated.

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
KvelaGorrrrnio: I don't know why I wrote this :D But it has been updated.

src/blueprints/tv2/action-factories/tv2-remote-action-factory.ts
KvelaGorrrrnio: Lambda is used instead.

KvelaGorrrrnio: Check has been added.

src/blueprints/tv2/value-objects/tv2-action.ts
KvelaGorrrrnio: `contentType` is in that case missing. But you are right. I have made it extend from Tv2Part instead.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #45, opened 2023-11-09
refactor(tv2-blueprints): DVE renamed to split screen.

KvelaGorrrrnio:
_**Note:** Is depending on https://github.com/tv2/sofie-blueprints-inews/pull/223._
Renamed DVE to split screen, where possible.


LindvedKrvang (approved):
For the record, I still think we should keep calling it `DVE` and not `SPLIT_SCREEN`.
Now we have a lot of places where we need to remember that value should be `dve`, but there is no clear way to indicate that.

Since all "DVE/SPLIT_SCREEN" is contained within Blueprints i.e. it's company specific and that we call it DVE at TV2, I think we should continue calling it `DVE`. It's going to cause a lot of confusion that we have two different terms for the same thing...

src/blueprints/tv2/helpers/tv2-blueprint-configuration-mapper.ts
```diff
@@ -68,7 +68,7 @@ export class Tv2BlueprintConfigurationMapper {
       graphicsDefault: this.mapGraphicsDefault(coreConfiguration.GfxDefaults),
       graphicsSetups: this.mapGraphicsSetups(coreConfiguration.GfxSetups),
       selectedGraphicsSetup: this.findSelectedGraphicsSetup(coreConfiguration.GfxDefaults, coreConfiguration.GfxSetups),
-      dveConfigurations: this.mapDveConfigurations(coreConfiguration.DVEStyles),
+      splitScreenConfigurations: this.mapSplitScreenConfigurations(coreConfiguration.DVEStyles),
```
RasmusAlbrektsen, line 71: What is `DVEStyles`, and i'm guessing this is not to be renamed?

src/blueprints/tv2/helpers/tv2-blueprint-configuration-mapper.ts
```diff
@@ -111,7 +111,7 @@ export class Tv2BlueprintConfigurationMapper {
     }
   }
 
-  private mapDveConfigurations(coreDveStyles: CoreDveStyle[]): DveConfiguration[] {
+  private mapSplitScreenConfigurations(coreDveStyles: CoreDveStyle[]): SplitScreenConfiguration[] {
```
RasmusAlbrektsen, line 114: In reference to the comment above, do we then keep `coreDveStyles` because of this, or should both be renamed?

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -357,56 +357,56 @@ export class Tv2DveActionFactory {
...
       }
     }
   }
 
-  private createRecallLastDveAction(): Tv2RecallDveAction {
+  private createRecallLastSplitScreenAction(): Tv2RecallSplitScreenAction {
     return {
-      id: 'recall_last_dve_action',
+      id: 'recall_last_split_screen_action',
       name: 'Recall DVE',
```
RasmusAlbrektsen, line 410: So i'm assuming this is intention, but why would we not scrub the entirety out even though it is in the blueprints?

RasmusAlbrektsen (approved):
I must say that i'm a bit confused in terms of what i think about this. 
On one hand i agree with @LindvedKrvang that in the blueprints this can become confusing. For me, it is mostly because we still use "DVE" in the strings we supply, like in the recall action, the description still says DVE. In my opinion we should make a clean break from DVE or be sure where we draw the line. I at least like that the code itself refers to it as split screens, because that is what we should aim to comply with in Sofie Server.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
KvelaGorrrrnio: I am assuming you are referring to the `name` property?
If that is the case then, we don't want to rename it, since it is a user facing value, where it should be DVE.

src/blueprints/tv2/helpers/tv2-blueprint-configuration-mapper.ts
KvelaGorrrrnio: I have renamed the `coreDveX` variables and interfaces to `coreSplitScreenX`.

src/blueprints/tv2/helpers/tv2-blueprint-configuration-mapper.ts
KvelaGorrrrnio: `DVEStyles` are the configuration table `DVE Layouts` in showstyle blueprints configuration. As long as we don't have control over settings, I don't think it should be renamed.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
RasmusAlbrektsen: I can see the argument if they are user facing strings :-)

PR #48, opened 2023-11-13
(SOF-1692) Only generate/return general and specified rundown actions

andr9528:
NOTE: The target branch for this PR is the branch for #28. Reviewing that one before this is adviced. 

- Adds filtering of AdLibPieces/AdLibActions extracted from database, to only be the ones for the requested rundownId.
- Adds rundownId to actions generated from ActionManifests.

Has been tested locally, ensureing that only actions for the specified rundownId is returned.
Let me know if there are other actions being generated, that are rundown specific, other than those generated from ActionManifests.

PR #49, opened 2023-11-13
(SOF-1691) Delete rundown actions before saving new ones.

andr9528:
NOTE: The base of this is the branch for #48. Reviewing that one before this is adviced.

- Adds deletion of rundown specific actions, prior to saving newly generated ones.

Has been verified locally, that actions with the specified rundown id is the only actions being removed by the new call.

### DIFF
--- a/src/data-access/repositories/interfaces/action-repository.ts
+++ b/src/data-access/repositories/interfaces/action-repository.ts
@@ -3,4 +3,5 @@ import { Action } from '../../../model/entities/action'
 export interface ActionRepository {
   getAction(actionId: string): Promise<Action>
   saveActions(actions: Action[]): Promise<void>
+  deleteRundownActions(rundownId: string): Promise<void>
 }
--- a/src/data-access/repositories/interfaces/action-repository.ts
+++ b/src/data-access/repositories/interfaces/action-repository.ts
@@ -3,5 +3,5 @@ import { Action } from '../../../model/entities/action'
 export interface ActionRepository {
   getAction(actionId: string): Promise<Action>
   saveActions(actions: Action[]): Promise<void>
-  deleteRundownActions(rundownId: string): Promise<void>
+  deleteActionsForRundown(rundownId: string): Promise<void>
 }
--- a/src/data-access/repositories/mongo/mongo-action-repository.ts
+++ b/src/data-access/repositories/mongo/mongo-action-repository.ts
@@ -32,8 +32,8 @@ export class MongoActionRepository extends BaseMongoRepository implements Action
     )
   }
 
-  public async deleteRundownActions(rundownId: string): Promise<void> {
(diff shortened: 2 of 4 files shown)


src/data-access/repositories/interfaces/action-repository.ts
```diff
@@ -3,4 +3,5 @@ import { Action } from '../../../model/entities/action'
 export interface ActionRepository {
   getAction(actionId: string): Promise<Action>
   saveActions(actions: Action[]): Promise<void>
+  deleteRundownActions(rundownId: string): Promise<void>
```
LindvedKrvang, line 6: This makes it sound like we have a concept called `RundownAction` which we technically don't.
`deleteActionsForRundown` has a nicer ring to it.

LindvedKrvang (approved):

src/data-access/repositories/interfaces/action-repository.ts
andr9528: Renamed as suggested.

src/presentation/controllers/action-controller.ts
```diff
@@ -25,10 +25,10 @@ export class ActionController extends BaseController {
   }
 
   @PutRequest('/:actionId/rundowns/:rundownId')
-  public async executeAction(reg: Request, res: Response): Promise<void> {
+  public async executeAction(req: Request, res: Response): Promise<void> {
```
KvelaGorrrrnio, line 28: A general thing for all of us. It would be nice to have controller parameters be `request` and `response`.

src/data-access/repositories/mongo/mongo-action-repository.ts
```diff
@@ -17,14 +19,25 @@ export class MongoActionRepository extends BaseMongoRepository implements Action
...
       )
     )
   }
+
+  public async deleteActionsForRundown(rundownId: string): Promise<void> {
+    this.assertDatabaseConnection(this.deleteActionsForRundown.name)
+    const actionsDeleteResult: DeleteResult = await this.getCollection().deleteMany({ rundownId: rundownId })
+
+    if (!actionsDeleteResult.acknowledged) {
+      throw new DeletionFailedException(`Failed to delete Actions for Rundown: ${rundownId}`)
```
KvelaGorrrrnio, line 40: It would be nice if we had consensus on how to structure error and log messages. It is not something that should be done for this PR though :D

KvelaGorrrrnio (approved):

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/presentation/controllers/action-controller.ts
andr9528: Will make a small PR changing this on all our controllers, after this is merged.

PR #51, opened 2023-11-13
fix: Correctly sets the duration so we avoid issues in Core

RasmusAlbrektsen:
This PR fixes the fact that upon ingest with Sofie Server open, Core would no longer render the pieces in it's UI.

### DIFF
--- a/src/data-access/repositories/mongo/mongo-entity-converter.ts
+++ b/src/data-access/repositories/mongo/mongo-entity-converter.ts
@@ -387,7 +387,7 @@ export class MongoEntityConverter {
   public convertToMongoPiece(piece: Piece): MongoPiece {
     const enable: MongoPiece['enable'] = {
       start: piece.getStart(),
-      duration: piece.duration
+      ...(piece.duration ? { duration: piece.duration } : null)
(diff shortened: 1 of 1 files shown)


LindvedKrvang (approved):

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -387,7 +387,7 @@ export class MongoEntityConverter {
   public convertToMongoPiece(piece: Piece): MongoPiece {
     const enable: MongoPiece['enable'] = {
       start: piece.getStart(),
-      duration: piece.duration
+      ...(piece.duration ? { duration: piece.duration } : null)
```
andr9528, line 390: Not 100% sure, but i think setting `ignoreUndefined` to true on the options object given to `.updateOne` will have the same outcome. With that it should mean the field in the database will simply not exist, thus when core pulls it, it should be null.

That option is one i have added for when we are saving actions during #28. Perhabs it is needed for multiple of our entities.

andr9528 (approved):

PR #52, opened 2023-11-14
SOF-1696 Use interface segregation for GraphicsTimelineObjectFactories

LindvedKrvang:
Update: All dependent PRs has been merged.

- Introduce StudioConfigurationMapper so we can use proper naming in our code.
- Tv2GraphicsActionFactory is no longer dependent on concrete implementations of Viz- and CasparCgTimelineObjectFactories.
- Different smaller fixes lost in merge-conflicts.
- TimelineObjectFactories now depend on `Tv2BlueprintTimelineObject` instead of `TimelineObject`
- ActionFactories now all uses `Tv2BlueprintTimelineObject` and `Tv2PieceInterface` instead of `TimelineObject` and `PieceInterface`.
- Ensure all Pieces on Actions have an `OutputLayer`.
- ActionManifests now have a separate file to be mapped in instead of being mapped in `Tv2ActionService`.
- Introduce `createProgramTimelineObjectWithWipeTranstion` (and same for Cleanfeed) to remove duplicate VideoMixer types.


src/blueprints/timeline-state-resolver-types/viz-types.ts
```diff
@@ -43,20 +43,23 @@ export interface VizMseContinueTimelineObject extends TimelineObject {
   content: {
     deviceType: DeviceType.VIZMSE
     type: VizType.CONTINUE
-    direction: -1 | 1 | undefined
+    direction: VizMseContinueDirection
```
andr9528, line 46: It appears to no longer be optional. 
I take that we currently, don't have anywhere that matters?

src/blueprints/tv2/action-factories/tv2-camera-action-factory.ts
```diff
@@ -28,7 +28,7 @@ export class Tv2CameraActionFactory {
   ) {}
 
   public createCameraActions(blueprintConfiguration: Tv2BlueprintConfiguration): Action[] {
-    return blueprintConfiguration.studio.SourcesCam
+    return blueprintConfiguration.studio.cameraeSources
```
andr9528, line 31: typo in the rename.
`cameraeSources` -> `cameraSources`

src/blueprints/timeline-state-resolver-types/viz-types.ts
LindvedKrvang: When would you create a `continue` object and not tell which direction it should go?

We currently only create one `continue` TimelineObject and we always set it to "forward"

src/blueprints/tv2/action-factories/tv2-camera-action-factory.ts
LindvedKrvang: Fixed

src/data-access/repositories/mongo/mongo-database.ts
```diff
@@ -4,8 +4,10 @@ import { DatabaseNotConnectedException } from '../../../model/exceptions/databas
 import { MongoId } from './mongo-entity-converter'
 
 // TODO: Move to ENV variables
-const MONGO_CONNECTION_STRING: string = 'mongodb://localhost:3001'
-const MONGO_DB_NAME: string = 'meteor'
+// const MONGO_CONNECTION_STRING: string = 'mongodb://localhost:3001'
+// const MONGO_DB_NAME: string = 'meteor'
+const MONGO_CONNECTION_STRING: string = 'mongodb://localhost:27017?directConnection=true'
+const MONGO_DB_NAME: string = 'sofie'
```
andr9528, line 10: I gather you mistakenly commited changes for mongo-database.

LindvedKrvang: Yup, that was a mistake. And it's not even the first time...

LindvedKrvang: It's been removed

src/blueprints/tv2/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
```diff
@@ -72,7 +72,7 @@ export class Tv2AtemVideoMixerTimelineObjectFactory implements Tv2VideoMixerTime
   }
 
   public createUpstreamKeyerTimelineObject(downstreamKeyer: Tv2DownstreamKeyer, enable: TimelineEnable): AtemMeTimelineObject {
-    const downstreamKeyerNumber: number = downstreamKeyer.Number + 1
+    const downstreamKeyerNumber: number = downstreamKeyer.name + 1
```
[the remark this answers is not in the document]
LindvedKrvang: I considered it.
I ended up keeping it since the type was still `number`, but then you could argue that `downstreamKeyer.name` shouldn't be called `name` either.
Don't have any strong preferences here, so whatever the majority is up for.

src/blueprints/tv2/value-objects/tv2-studio-blueprint-configuration.ts
```diff
@@ -1,87 +1,90 @@
...
-  _id: string
-  Number: number
-  Key: number
-  Fill: number
-  DefaultOn: boolean
-  Roles: Tv2DownstreamKeyerRole[]
-  Clip: number,
-  Gain: number
+  id: string // This id isn't really used
+  name: number // For some reason this is a number
```
andr9528, line 75: It being named `name` implies that it is of type string, but in this case it is a number. The data being mapped into this field comes from a column of cells where the content is literally `[1, 2, 3, ...]`. Keeping the name as number makes sense in my mind, of cource with proper capalization. This will also allow you to remove the comment.

andr9528: Core saves them as `[0, 1, 2, ...]` though, which is why one is added to location where it is used.

KvelaGorrrrnio (commented):
A lot of good changes :rocket:

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -359,15 +377,15 @@ export class Tv2SplitScreenActionFactory {
...
         this.videoMixerTimelineObjectFactory.createLookaheadTimelineObject(splitScreenSource, videoSwitcherTimelineEnable),
-        this.casparCgTimelineObjectFactory.createSplitScreenKeyTimelineObject(this.assetPathHelper.joinAssetToFolder(splitScreenConfiguration.key, blueprintConfiguration.studio.DVEFolder)),
-        this.casparCgTimelineObjectFactory.createSplitScreenFrameTimelineObject(this.assetPathHelper.joinAssetToFolder(splitScreenConfiguration.frame, blueprintConfiguration.studio.DVEFolder)),
-        this.casparCgTimelineObjectFactory.createSplitScreenLocatorTimelineObject(),
+        this.graphicsSplitScreenTimelineObjectFactory.createSplitScreenKeyTimelineObject(this.assetPathHelper.joinAssetToFolder(splitScreenConfiguration.key, blueprintConfiguration.studio.splitScreenFolder?.name)),
+        this.graphicsSplitScreenTimelineObjectFactory.createSplitScreenFrameTimelineObject(this.assetPathHelper.joinAssetToFolder(splitScreenConfiguration.frame, blueprintConfiguration.studio.splitScreenFolder?.name)),
+        // this.graphicsSplitScreenTimelineObjectFactory.createSplitScreenLocatorTimelineObject(),
```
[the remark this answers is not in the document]
LindvedKrvang: It's fixed/uncommented in https://github.com/tv2/sofie-server/pull/55

src/blueprints/tv2/action-factories/tv2-video-mixer-configuration-action-factory.ts
```diff
@@ -11,24 +11,25 @@ import { TransitionType } from '../../../model/enums/transition-type'
...
   }
 
   private createDownstreamKeyerOffActions(blueprintConfiguration: Tv2BlueprintConfiguration): Tv2PieceAction[] {
-    return blueprintConfiguration.studio.SwitcherSource.DSK.map(config => this.createDownStreamKeyerAction(config, 'Off', false))
+    return blueprintConfiguration.studio.videoMixerBasicConfiguration.downstreamKeyers.map(config => this.createDownStreamKeyerAction(config, 'Off', false))
   }
 
   private createDownStreamKeyerAction(downstreamKeyer: Tv2DownstreamKeyer, actionName: string, isOn: boolean): Tv2PieceAction {
-    const downstreamKeyerNumber: string = String(downstreamKeyer.Number + 1)
+    const downstreamKeyerNumber: string = String(downstreamKeyer.name + 1)
```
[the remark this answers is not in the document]
LindvedKrvang: I agree.
The `name` is set in the "Number" column here
image

Don't have any good suggestion myself. I prefer `index` of the two you mentioned

src/blueprints/timeline-state-resolver-types/abstract-types.ts
```diff
@@ -1,9 +1,9 @@
-import { TimelineObject } from '../../model/entities/timeline-object'
 import { DeviceType } from '../../model/enums/device-type'
+import { Tv2BlueprintTimelineObject } from '../tv2/value-objects/tv2-metadata'
 
-export interface EmptyTimelineObject extends TimelineObject {
+export interface EmptyTimelineObject extends Tv2BlueprintTimelineObject {
```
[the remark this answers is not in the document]
LindvedKrvang: I agree. That was a blunder from my side. Good catch :)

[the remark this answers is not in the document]
LindvedKrvang: It's been updated

src/blueprints/tv2/timeline-object-factories/tv2-empty-graphics-command-timeline-object-factory.ts
```diff
@@ -0,0 +1,42 @@
+import { Tv2GraphicsCommandTimelineObjectFactory } from './interfaces/tv2-graphics-command-timeline-object-factory'
+import { Tv2BlueprintConfiguration } from '../value-objects/tv2-blueprint-configuration'
+import { EmptyTimelineObject } from '../../timeline-state-resolver-types/abstract-types'
+import { DeviceType } from '../../../model/enums/device-type'
+
+export class Tv2EmptyGraphicsCommandTimelineObjectFactory implements Tv2GraphicsCommandTimelineObjectFactory {
```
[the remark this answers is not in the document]
LindvedKrvang: The interface `Tv2GraphicsCommandTimelineObjectFactory` only has, as to my knowlegde, a Viz implementation.
If there is no Viz in the Gallery, then we shouldn't be creating Viz TimelineObjects, so we are creating instance of this class instead.

I guess one could argue that it's fine to just create the Viz TimelineObjects anyway in a Gallery without a Viz since there is no hardware to send it to, so the TimelineObjects wouldn't have an effect?
Any preference for one or the other?

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
```diff
@@ -149,14 +140,14 @@ export class Tv2VizTimelineObjectFactory implements Tv2GraphicsTimelineObjectFac
...
-    if (!blueprintConfiguration.studio.PreventOverlayWithFull) {
+    if (!blueprintConfiguration.studio.preventOverlayWhileFullscreenGraphicsIsOnAir) {
       return {}
     }
     return {
       delayTakeAfterOutTransition: true,
       outTransition: {
         type: VizMseTransitionType.DELAY,
-        delay: blueprintConfiguration.studio.VizPilotGraphics.OutTransitionDuration
+        delay: blueprintConfiguration.studio.vizPilotGraphics.msKeepPilotGraphicsAliveBeforeTakingNext
```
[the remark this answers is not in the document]
LindvedKrvang: I was going for something that would make sense in this context:
image

If `outTransitionDuration` is prefered I can change it back. 
Any naming suggestions to the other attributes in the picture?

src/blueprints/tv2/helpers/tv2-studio-blueprint-configuration-mapper.ts
```diff
@@ -0,0 +1,254 @@
...
+      msKeepOldPartAliveBeforeTakingGraphics: coreGraphics.KeepAliveDuration,
+      graphicsUrl: coreGraphics.GraphicURL,
+      transitionSettings: {
+        wipeRate: coreGraphics.TransitionSettings ? coreGraphics.TransitionSettings.wipeRate : 0,
+        borderSoftness: coreGraphics.TransitionSettings ? coreGraphics.TransitionSettings.borderSoftness: 0
+      }
+    }
+  }
+
+  private mapAudiBedSettings(coreAudioBed: CoreAudioBedSettings): AudioBedSettings {
```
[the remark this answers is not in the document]
LindvedKrvang: That's why they (don't) pay us the big bucks ;)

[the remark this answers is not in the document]
LindvedKrvang: It's been updated

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
```diff
@@ -149,14 +140,14 @@ export class Tv2VizTimelineObjectFactory implements Tv2GraphicsTimelineObjectFac
       delayTakeAfterOutTransition?: boolean
       outTransition?: VizMseElementPilotTimelineObject['content']['outTransition']
     } {
-    if (!blueprintConfiguration.studio.PreventOverlayWithFull) {
+    if (!blueprintConfiguration.studio.preventOverlayWhileFullscreenGraphicsIsOnAir) {
```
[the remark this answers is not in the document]
LindvedKrvang: It's been updated

src/blueprints/tv2/action-factories/tv2-video-mixer-configuration-action-factory.ts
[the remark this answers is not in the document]
LindvedKrvang: Updated it to `index`

src/blueprints/tv2/value-objects/tv2-studio-blueprint-configuration.ts
LindvedKrvang: It's no longer called `name`

src/blueprints/tv2/timeline-object-factories/tv2-empty-graphics-command-timeline-object-factory.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: It is better that we create empty timeline objects, rather than Viz timeline objects when they aren't needed. So this is fine, but ideally we shouldn't create timeline objects that the setup is not using.

src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
```diff
@@ -225,7 +226,7 @@ export class Tv2AudioActionFactory {
     const audioAction: Tv2FadeAudioBedAction = action as Tv2FadeAudioBedAction
 
     const fadeDurationInMilliseconds: number =  this.getTimeFromFrames(this.isInteger(fadeDurationInFrames) ? fadeDurationInFrames : audioAction.metadata.defaultFadeDurationInFrames)
-    const fadeAudioBedTimelineObject: TimelineObject = this.casparCgTimelineObjectFactory.createFadeAudioBedTimelineObject(fadeDurationInMilliseconds)
+    const fadeAudioBedTimelineObject: Tv2BlueprintTimelineObject = this.videoClipTimelineObjectFactory.createFadeAudioBedTimelineObject(fadeDurationInMilliseconds)
```
andr9528, line 229: The `.createFadeAudioBedTimelineObject` sounds like a method that should be on a `audioTimelineObjectFactory`, or am i misunderstanding what a 'AudioBed' is?

Perhabs it is just the name of the method that is incorrect.

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: I have tried to give suggestions to what I think makes sense :)

image

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
```diff
@@ -11,31 +12,31 @@ import {
...
-  public createThemeOutTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration, duration: number): VizMseElementInternalTimelineObject {
+export class Tv2VizTimelineObjectFactory implements Tv2GraphicsCommandTimelineObjectFactory, Tv2GraphicsElementTimelineObjectFactory {
+
+  public createThemeOutTimelineObject(blueprintConfiguration: Tv2BlueprintConfiguration): VizMseElementInternalTimelineObject {
     return {
       id: '',
       enable: {
         start: 0,
-        duration
+        duration: 3000
```
KvelaGorrrrnio, line 34: Magic literal. Same goes for the other durations.

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
```diff
@@ -11,31 +12,31 @@ import {
...
-import { TimelineObject } from '../../../model/entities/timeline-object'
 import {
   Tv2FullscreenGraphicsManifestData,
   Tv2OverlayGraphicsManifestData
 } from '../value-objects/tv2-action-manifest-data'
+import { Tv2GraphicsCommandTimelineObjectFactory } from './interfaces/tv2-graphics-command-timeline-object-factory'
 
-
-enum ChannelName {
+enum Engine {
```
KvelaGorrrrnio, line 21: Engine seems to be referring to a broader data structure than a name/identifier. I suggest to rename this to `EngineName`.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -74,14 +74,14 @@ export class Tv2ReplayActionFactory {
   }
```
andr9528, line 74: This should use `.replaceAll`. 

When using a string for the pattern in the `.replace`, only the first occurrence is changed. 
If a regex is used instead in `.replace` then it updates all occurrences.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -229,13 +246,12 @@ export class Tv2SplitScreenActionFactory {
...
         .map(source => {
-
-          const audioTimelineObjects: TimelineObject[] = this.audioTimelineObjectFactory.createTimelineObjectsForSource(blueprintConfiguration, source, audioMode)
+          const audioTimelineObjects: Tv2BlueprintTimelineObject[] = this.audioTimelineObjectFactory.createTimelineObjectsForSource(blueprintConfiguration, source, audioMode)
 
           return {
-            id: `insert_${name}_${source.SourceName}_to_split_screen_input_${inputIndex}_action`,
-            name: `Insert ${name} ${source.SourceName} in DVE input ${inputIndex}`,
-            description: `Insert ${name} ${source.SourceName} in DVE input ${inputIndex}`,
+            id: `insert_${name}_${source.name}_to_split_screen_input_${inputIndex}_action`,
```
andr9528, line 252: It looks like `source.name` need to be sanitized for spaces.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -499,14 +517,14 @@ export class Tv2SplitScreenActionFactory {
...
+    const audioTimelineObjects: Tv2BlueprintTimelineObject[] = this.audioTimelineObjectFactory.createVideoClipAudioTimelineObjects(blueprintConfiguration, {
       fileName: inputIndex,
       audioMode
     } as unknown as Tv2VideoClipManifestData)
 
     return {
-      id: `insert_last_video_clip_to_split_screen_input_${inputIndex}${audioMode ? '_vo' : ''}_action`,
-      name: `Insert last Video ${audioMode ? ' Voice Over ' : ''} Clip in DVE input ${inputIndex}`,
+      id: `insert_last_video_clip_to_split_screen_input_${inputIndex}${audioMode === Tv2AudioMode.VOICE_OVER ? '_vo' : ''}_action`,
+      name: `Insert last Video ${audioMode === Tv2AudioMode.VOICE_OVER ? ' Voice Over ' : ''} Clip in DVE input ${inputIndex}`,
```
andr9528, line 527: One too many spaces in the resulting string, if the audioMode is `VOICE_OVER`.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -499,14 +517,14 @@ export class Tv2SplitScreenActionFactory {
...
       description: 'Insert last Video Clip in DVE input ${inputIndex}',
```
andr9528, line 528: Not using backticks, when it should.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
andr9528: Removing the space before 'Clip' should fix it.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
andr9528: Hmm, it didn't seem to want to put my comment on the correct line.
It was suppose to be on the line for `return value.replace(' ', '')` just above...

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
LindvedKrvang: Can you elaborate? What should be using backticks that isn't using it?

src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
LindvedKrvang: I understand your confusion. I had a hard time finding a place to put that method.
The trick is, it's CasparCG that plays the audiobed and it's also CasparCG that fades the audiobed.
So we need some way to get a TimelineObjectFactory for CasparCG when fading the audiobed. 
Putting `createFadeAudioBedTimelineObject` on `Tv2AudioTimelineObjectFactory` and then create a `Tv2CasparCgAudioTimelineObjectFactory` would mean all but one of the methods on the interface would have to throw an unsupported exception. Also, how would we differentiate between which TimelineObjectFactory to use in the AudioActionFactory because we still need the `Tv2SisyfosAudioTimelineObjectFactory`?

I agree, the current implementation isn't optimal and I'm open for suggestions.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
andr9528: The string should use backticks, so the `inputIndex` is put into the string as defined.
Currently the output string literally says `Insert last Video Clip in DVE input ${inputIndex}`

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
LindvedKrvang: I'm not sure what is the best way to handle it in the timelineObject generation, because they basically are just creating objects with a lot of specific values. We could spam the file with variables if we extracted every single one.

Any suggestions?

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
LindvedKrvang: Updated.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
LindvedKrvang: Updated

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
LindvedKrvang: Updated to replace all whitespaces with underscores in the id.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
LindvedKrvang: Updated.

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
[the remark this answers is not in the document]
LindvedKrvang: Don't have any strong feelings for or against this.
I have updated it to your suggestion :)

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
KvelaGorrrrnio: There are a couple of `3000` literals and a couple of `1000` literals in this file. Do they have the duration for the same reason? And if one were to be changed, should the others be changed? If so, then I think we can make som categoric constants. If they all have unique reasons for changing the matching values, then I have no issue with keeping the magic literals.

KvelaGorrrrnio (approved):

src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
andr9528: That explanation helped. I see now that the current implementation is the better of 2 rotten tomatoes.
Don't have any better suggestion at the moment.

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
LindvedKrvang: Ahh, I was looking a line 527 since that was the one with changes. I see it now. Good catch.
It has been updated

andr9528 (approved):

PR #53, opened 2023-11-14
Sof 1683/Rundown Timings

RasmusAlbrektsen:
This PR brings in the `untimed` properties on parts and sets it on the segment if all parts in the segment are untimed.


src/model/entities/segment.ts
```diff
@@ -241,4 +241,8 @@ export class Segment {
   public getUnsyncedCopy(): Segment {
     return Object.assign(Object.create(Object.getPrototypeOf(this)), this, { id: `${this.id}${UNSYNCED_ID_POSTFIX}`})
   }
+
+  public getIsSegmentUntimed(): boolean {
+    return !this.getParts().some(part => !part.isUntimed())
```
andr9528, line 246: So if a segment contains one or more parts that are timed, then the segment is untimed?

Are you sure the negations are suppose to be here?

LindvedKrvang: If `this.getParts().some(part => !part.isUntimed())` returns true, that means that atleast one Part is `timed`.
If atleast one Part is `timed`, then the Segment is also `timed` so in order for `getIsSegmentUntimed()` to return true when it's `untimed`, it needs to negate `this.getParts().some(part => !part.isUntimed())`.

So yes, the negations are needed.

RasmusAlbrektsen: This says that if a segment has at least one part that is timed. The segment is considered timed, and the method will therefore return `false`.
If we have no timed parts in the segment, it will return `true`, meaning the segment is untimed.

src/data-access/repositories/mongo/mongo-rundown-changed-listener.ts
```diff
@@ -53,8 +62,14 @@ export class MongoRundownChangedListener extends BaseMongoRepository implements
       }
       case MongoChangeEvent.REPLACE: {
         const replaceChange: ChangeStreamReplaceDocument<MongoRundown> = change as ChangeStreamReplaceDocument<MongoRundown>
-        const rundownId: string = replaceChange.documentKey._id
-        const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
+        const mongoRundown: MongoRundown = replaceChange.fullDocument
+        const rundownId: string = mongoRundown._id
+        const baselineTimelineObjects: TimelineObject[] = await this.rundownBaselineRepository.getRundownBaseline(
+          rundownId
+        )
```
LindvedKrvang, line 69: This could easily fit on one line

src/data-access/repositories/mongo/mongo-rundown-changed-listener.ts
```diff
@@ -25,7 +31,10 @@ export class MongoRundownChangedListener extends BaseMongoRepository implements
   constructor(
     mongoDatabase: MongoDatabase,
     mongoEntityConverter: MongoEntityConverter,
-    private readonly rundownRepository: RundownRepository
+    private readonly rundownRepository: RundownRepository,
+    private readonly rundownBaselineRepository: RundownBaselineRepository,
+    private readonly segmentRepository: SegmentRepository,
+    private readonly pieceRepository: PieceRepository
```
LindvedKrvang, line 37: I'm slightly concerned about injecting more repositories than the one be listening on, but I do see/understand why you are doing it.

It shouldn't become a habit though :)

src/model/entities/segment.ts
```diff
@@ -241,4 +241,8 @@ export class Segment {
   public getUnsyncedCopy(): Segment {
     return Object.assign(Object.create(Object.getPrototypeOf(this)), this, { id: `${this.id}${UNSYNCED_ID_POSTFIX}`})
   }
+
+  public getIsSegmentUntimed(): boolean {
```
LindvedKrvang, line 245: Methods returning `boolean` should follow the `isX` or `hasX` naming convention. The `get` here is redundant.

LindvedKrvang (approved):
The approve assumes that the `getIsSegmentUntimed()` name will be updated.

src/data-access/repositories/mongo/mongo-rundown-changed-listener.ts
RasmusAlbrektsen: Noted

LindvedKrvang: I'm been made aware that some `NOT_FOUND` errors has been thrown from the database changed listeners on your branch.
I suspect very much that the changes in this file is the reason for it.

LindvedKrvang (approved):

PR #54, opened 2023-11-16
(SOF-1689) Fix randomly failing tests

andr9528:
- Changes repository tests to be run in sequence.
- Changes repository tests such that only one in memory instance is created.
- Changes repository tests to drop the database between tests.

Would appreciate before merging is done, if @KvelaGorrrrnio and/or @RasmusAlbrektsen would pull the branch down locally, and run tests. They should find that the tests does not throw the Port error anymore.
Have run our GitHub check action, which did throw the error, 10 times in a row without the error popping up.

On a side note:
How the `mongo-repositories.spec.ts` file is setup, is along the lines of what i had in mind, if we wanted to split some of the large test files into smaller files. The `rundown.spec.ts` file is 3000+ lines long, which might benefit from a similar split, if we wanted to improve the readability, and could find a good describe level to split at. Some of the other entity test files might similarly benefit from this split concept. Just some food for thought.

### DIFF
--- a/src/data-access/repositories/test/mongo-part-repository.ts
+++ b/src/data-access/repositories/test/mongo-part-repository.ts
@@ -13,15 +13,14 @@ import { EntityTestFactory } from '../../../model/entities/test/entity-test-fact
 const COLLECTION_NAME = 'parts'
 
 export function MongoPartRepositoryTests(testDatabase: MongoTestDatabase): void {
-  afterEach(() => testDatabase.teardownDatabase(COLLECTION_NAME))
   describe(MongoPartRepository.prototype.deletePartsForSegment.name, () => {
     it('deletes one part successfully', async () => {
       const mongoConverter: MongoEntityConverter = mock(MongoEntityConverter)
       const segmentId: string = 'someSegmentId'
       const part: MongoPart = createMongoPart({ segmentId: segmentId })
       const entityPart: Part = EntityMockFactory.createPart({ segmentId: segmentId })
-      await testDatabase.populateDatabaseWithParts(COLLECTION_NAME, [part])
-      const db: Db = testDatabase.getDatabase(COLLECTION_NAME)
+      await testDatabase.populateDatabaseWithParts([part])
+      const db: Db = testDatabase.getDatabase()
 
       when(mongoConverter.convertParts(anything())).thenReturn([entityPart])
       const testee: PartRepository = createTestee({
@@ -34,7 +33,7 @@ export function MongoPartRepositoryTests(testDatabase: MongoTestDatabase): void 
     })
 
     it('deletes multiple parts successfully', async () => {
-      const db: Db = testDatabase.getDatabase(COLLECTION_NAME)
+      const db: Db = testDatabase.getDatabase()
       const mongoConverter: MongoEntityConverter = mock(MongoEntityConverter)
       const segmentId: string = 'someSegmentId'
       const mongoParts: MongoPart[] = [
@@ -45,7 +44,7 @@ export function MongoPartRepositoryTests(testDatabase: MongoTestDatabase): void 
         EntityMockFactory.createPart({ segmentId: segmentId }),
         EntityMockFactory.createPart({ segmentId: segmentId }),
       ]
-      await testDatabase.populateDatabaseWithParts(COLLECTION_NAME, mongoParts)
+      await testDatabase.populateDatabaseWithParts(mongoParts)
 
       when(mongoConverter.convertParts(anything())).thenReturn(parts)
       const testee: PartRepository = createTestee({
@@ -69,7 +68,7 @@ export function MongoPartRepositoryTests(testDatabase: MongoTestDatabase): void 
         EntityMockFactory.createPart({ segmentId: segmentId }),
         EntityMockFactory.createPart({ segmentId: segmentId }),
       ]
-      await testDatabase.populateDatabaseWithParts(COLLECTION_NAME, mongoParts)
+      await testDatabase.populateDatabaseWithParts(mongoParts)
 
       when(mongoConverter.convertParts(anything())).thenReturn(parts)
       when(pieceRepository.getPieces(anything())).thenResolve([])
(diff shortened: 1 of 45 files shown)


src/data-access/repositories/test/mongo-part-repository.ts
KvelaGorrrrnio, line 1: It can easily create confusion that the test file is called the same as the implementing file.
The filename could perhaps be suffixed with `-tests`, such that this become `mongo-part-repository-tests.ts`.

src/data-access/repositories/test/mongo-part-repository.ts
```diff
@@ -12,11 +12,7 @@ import { EntityTestFactory } from '../../../model/entities/test/entity-test-fact
 
 const COLLECTION_NAME = 'parts'
 
-describe(MongoPartRepository.name, () => {
-  const testDatabase: MongoTestDatabase = new MongoTestDatabase()
-  beforeEach(async () => testDatabase.setupDatabase())
-  afterEach(async () => testDatabase.teardownDatabase())
-
+export function MongoPartRepositoryTests(testDatabase: MongoTestDatabase): void {
```
KvelaGorrrrnio, line 15: The method name sounds like a collection of tests. To indicate that this is a function that runs the tests, it could be renamed to `testMongoPartRepository` or `runTestsForMongoPartRepository`.

src/data-access/repositories/test/mongo-repositories.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+import { MongoRundownRepository } from '../mongo/mongo-rundown-repository'
+import { MongoRundownRepositoryTests } from './mongo-rundown-repository'
+import { MongoSegmentRepository } from '../mongo/mongo-segment-repository'
+import { MongoSegmentRepositoryTests } from './mongo-segment-repository'
+import { MongoPartRepository } from '../mongo/mongo-part-repository'
+import { MongoPartRepositoryTests } from './mongo-part-repository'
+import { MongoPieceRepository } from '../mongo/mongo-piece-repository'
+import { MongoPieceRepositoryTests } from './mongo-piece-repository'
+
+describe('Mongo Repositories', () => {
```
KvelaGorrrrnio, line 11: Repositories should be with lowercased r :)

src/data-access/repositories/test/mongo-repositories.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+  const testDatabase: MongoTestDatabase = new MongoTestDatabase()
+  beforeAll(async () => testDatabase.setupDatabaseConnection())
+  afterAll(async () => testDatabase.teardownDatabaseConnection())
```
KvelaGorrrrnio, line 14: `async ` should be removed.

src/data-access/repositories/test/mongo-test-database.ts
```diff
@@ -11,12 +11,12 @@ export class MongoTestDatabase {
     jest.setTimeout(15000)
   }
 
-  public async setupDatabase(): Promise<void> {
+  public async setupDatabaseConnection(): Promise<void> {
```
KvelaGorrrrnio, line 14: If I were to call this method, I would expect that only the database connection is established – not that the entire database server is created.

src/data-access/repositories/test/mongo-test-database.ts
```diff
@@ -11,12 +11,12 @@ export class MongoTestDatabase {
...
     this.mongoServer = await MongoMemoryServer.create()
     this.client = await MongoClient.connect(this.mongoServer.getUri())
   }
 
-  public async teardownDatabase(): Promise<void> {
+  public async teardownDatabaseConnection(): Promise<void> {
```
KvelaGorrrrnio, line 19: If I were to call this method, I would expect that only the database connection is closed – not the entire database server.

src/data-access/repositories/test/mongo-test-database.ts
```diff
@@ -25,6 +25,10 @@ export class MongoTestDatabase {
     }
   }
 
+  public async teardownDatabase(): Promise<void> {
```
KvelaGorrrrnio, line 28: Since this is a mongo test database class, then it would fine to call this method for `dropDatabase` to avoid confusion whether the database was dropped or cleared.

KvelaGorrrrnio (commented):
I tested it locally and it seems to work! Nice work! :rocket:
I have added some comments on the naming.

src/data-access/repositories/test/mongo-part-repository.ts
andr9528: Suffixed `-tests` to the file names.

src/data-access/repositories/test/mongo-part-repository.ts
andr9528: Prefixed `Run-` to the function names.

src/data-access/repositories/test/mongo-repositories.spec.ts
andr9528: Fixed

src/data-access/repositories/test/mongo-repositories.spec.ts
andr9528: Gone.

src/data-access/repositories/test/mongo-test-database.ts
andr9528: Renamed to `setupDatabaseServer`, and similarly for the teardown method.

src/data-access/repositories/test/mongo-test-database.ts
andr9528: Renamed similarly as the setup method.

src/data-access/repositories/test/mongo-part-repository.ts
KvelaGorrrrnio: We use camelCase for function names, so r in Run should be lowercased.

KvelaGorrrrnio (approved):

LindvedKrvang (approved):

PR #55, opened 2023-11-17
SOF-1695 Implement locators for splitscreen

LindvedKrvang:
Dependent on https://github.com/tv2/sofie-server/pull/52

SplitScreen Actions now have Locators when relevant.

Includes some minor fixes some Graphics Actions.

### DIFF
--- a/src/blueprints/tv2/timeline-object-factories/interfaces/tv2-graphics-split-screen-timeline-object-factory.ts
+++ b/src/blueprints/tv2/timeline-object-factories/interfaces/tv2-graphics-split-screen-timeline-object-factory.ts
@@ -1,7 +1,8 @@
 import { Tv2BlueprintTimelineObject } from '../../value-objects/tv2-metadata'
+import { GraphicsSetup, SplitScreenConfiguration } from '../../value-objects/tv2-show-style-blueprint-configuration'
 
 export interface Tv2GraphicsSplitScreenTimelineObjectFactory {
   createSplitScreenKeyTimelineObject(keyFilePath: string): Tv2BlueprintTimelineObject
   createSplitScreenFrameTimelineObject(frameFilePath: string): Tv2BlueprintTimelineObject
-  createSplitScreenLocatorTimelineObject(): Tv2BlueprintTimelineObject
+  createSplitScreenLocatorTimelineObject(graphicsSetup: GraphicsSetup, splitScreenConfiguration: SplitScreenConfiguration, locatorLabels?: string[]): Tv2BlueprintTimelineObject
 }
--- a/src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
+++ b/src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
@@ -22,6 +22,7 @@ import {
   Tv2GraphicsSplitScreenTimelineObjectFactory
 } from './interfaces/tv2-graphics-split-screen-timeline-object-factory'
 import { Tv2VideoClipTimelineObjectFactory } from './interfaces/tv2-video-clip-timeline-object-factory'
+import { GraphicsSetup, SplitScreenConfiguration } from '../value-objects/tv2-show-style-blueprint-configuration'
 
 enum CasparCgSlot {
   FULL_GRAPHICS = '250_full',
@@ -29,6 +30,8 @@ enum CasparCgSlot {
   LOWER_THIRD = '450_lowerThird',
   UNKNOWN = 'UNKNOWN'
 }
+
+const HTML_GRAPHICS_INDEX_FILE: string = 'index'
 
 export class Tv2CasparCgTimelineObjectFactory implements Tv2GraphicsElementTimelineObjectFactory, Tv2GraphicsSplitScreenTimelineObjectFactory, Tv2VideoClipTimelineObjectFactory {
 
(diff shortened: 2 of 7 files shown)


src/blueprints/tv2/timeline-object-factories/interfaces/tv2-graphics-split-screen-timeline-object-factory.ts
```diff
@@ -1,7 +1,8 @@
 import { Tv2BlueprintTimelineObject } from '../../value-objects/tv2-metadata'
+import { GraphicsSetup, SplitScreenConfiguration } from '../../value-objects/tv2-show-style-blueprint-configuration'
 
 export interface Tv2GraphicsSplitScreenTimelineObjectFactory {
   createSplitScreenKeyTimelineObject(keyFilePath: string): Tv2BlueprintTimelineObject
   createSplitScreenFrameTimelineObject(frameFilePath: string): Tv2BlueprintTimelineObject
-  createSplitScreenLocatorTimelineObject(): Tv2BlueprintTimelineObject
+  createSplitScreenLocatorTimelineObject(graphicsSetup: GraphicsSetup, splitScreenConfiguration: SplitScreenConfiguration, locatorLabels?: string[]): Tv2BlueprintTimelineObject
```
KvelaGorrrrnio, line 7: Since `locatorLabels` is a string array, then an empty array would probably suffice to describe a split screen without locator.

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
```diff
@@ -30,6 +31,8 @@ enum CasparCgSlot {
   UNKNOWN = 'UNKNOWN'
 }
 
+const HTML_GRAPHICS_INDEX_FILE: string = 'index'
```
KvelaGorrrrnio, line 34: CasparCG supports not having an extension, but for clarity: Should the index filename have `.html` suffixed?
And this is a filename, not a file :)

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
```diff
@@ -204,21 +207,42 @@ export class Tv2CasparCgTimelineObjectFactory implements Tv2GraphicsElementTimel
...
+          slots:
+            {
+              '850_dve': {
+                display: 'program',
+                payload: {
+                  type: 'locators',
+                  style: locatorLabels && locatorLabels.length > 0
+                    ? JSON.parse(splitScreenConfiguration.graphicsTemplateJson)
+                    : '',
+                  ...locatorLabels?.map((label): string => label)
```
KvelaGorrrrnio, line 239: What is the purpose of this map?

LindvedKrvang: Leftover from a previous implementation, but no longer needed. It's been updated.

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
LindvedKrvang: This is meant to be the name of the file that CasparCG should chose to display HTML graphics
image
As far as I'm aware, we are not supposed to send the file type with, but I might be wrong.

I have updated the variable name to indicate that it's the filename and not the file.

src/blueprints/tv2/timeline-object-factories/interfaces/tv2-graphics-split-screen-timeline-object-factory.ts
LindvedKrvang: We could, but then we are forced to add an empty array all the places where we create SplitScreens and we don't have any locators:
```
this.graphicsSplitScreenTimelineObjectFactory.createSplitScreenLocatorTimelineObject(blueprintConfiguration.showStyle.selectedGraphicsSetup, splitScreenConfiguration, [])
```
But I guess we should just not create the locator object in those cases?

KvelaGorrrrnio: By not creating the timeline object, could the current split screen get locators from a lookahead?

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
KvelaGorrrrnio: Fair, no extension then :)

src/blueprints/tv2/timeline-object-factories/interfaces/tv2-graphics-split-screen-timeline-object-factory.ts
LindvedKrvang: Yes, it would take it from lookahead.
So if SplitScreen A is on Air and don't have locators, but SplitScreen B is next and have locators. The locators from B would be applied and showed for A if we don't create the empty object.

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
```diff
@@ -204,21 +207,42 @@ export class Tv2CasparCgTimelineObjectFactory implements Tv2GraphicsElementTimel
...
+        useStopCommand: false,
+        mixer: {
+          opacity: 100
+        },
+        data: {
+          display: 'program',
+          partialUpdate: true,
+          slots:
+            {
+              '850_dve': {
```
andr9528, line 232: I think this should make use of `this.mapTv2GraphicsLayerToHtmlGraphicsSlot(Tv2GraphicsLayer.GRAPHICS_LOCATORS)` to define the used slot similarly to other methods in this class that return `CasparCgTemplateTimelineObject`.

Doing so would also ensure we make use of the `CasparCgSlot` Enum, instead of a magic string.

LindvedKrvang: It's not a magic string.
`850_dve` is defined on `CasparCgTemplateData`.
image
Writing something else result in a compile error:
image

RasmusAlbrektsen (approved):

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
andr9528: I see.

Perhabs it is the other way around then, and where we use `this.mapTv2GraphicsLayerToHtmlGraphicsSlot` we should not, and instead have it defined as done in this method.

KvelaGorrrrnio: Whether or not it is a magic string depends on your definition of a magic string – which there are multiple variations of. But the literal string is typed. Personally I think it would be nice to use an enumeration to indicate what slots we support and give them meaningful names. E.g. instead of `850_dve`, we could have the value `CasparCgSlot.SPLIT_SCREEN`. This way, we also reduce the co-existence of DVE and Split screen in the code.

Currently we mix the typed string literals and the `CasparCGSlot` enum for types and values. The enum is only used (partially) in this file. 
I'll suggest having the enum moved out and use it to define the supported slots on the `CasparCgTemplateData` interface.

KvelaGorrrrnio: In addition, the `display` and `payload.type` attributes currently have the type `string`, but only support specific values. That would also be nice to have specialized.

KvelaGorrrrnio (approved):
I would like to see a common way for defining and using the CasparCG slot types.

PR #56, opened 2023-11-20
Sof 1680/on air timing

KvelaGorrrrnio:
When taking a segment, the new attribute `executedAtEpochTime` is set, to indicate when the segment was put on air.
When taking another part in the same segment, the attribute is kept, but when taking a part in another segment, the attribute is cleared.

Budget duration for segments has been renamed to expected duration.


src/model/entities/rundown.ts
```diff
@@ -350,6 +351,23 @@ export class Rundown extends BasicRundown {
...
+    }
+
+    if (onAirSegment) {
+      onAirSegment.clearExecutedAtEpochTime()
+    }
+
+    if (nextSegment) {
+      nextSegment.setExecutedAtEpochTime(executedAtEpochTime)
+    }
+  }
```
LindvedKrvang, line 369: This seems like something that should be handled by the Segment itself.
When the Segment is put onAir it should set its executedAtTime. When it's taking ofAir, it should clear it. No need to let the Rundown handled that logic.

src/model/entities/segment.ts
```diff
@@ -58,6 +61,18 @@ export class Segment {
...
+    this.executedAtEpochTime ??= executedAtEpochTime
+  }
+
+  public clearExecutedAtEpochTime(): void {
+    this.executedAtEpochTime = undefined
+  }
+
+  public getExecutedAtEpochTime(): number | undefined {
+    return this.executedAtEpochTime
+  }
```
LindvedKrvang, line 74: No need to expose these methods if you just let the Segment handle its own executedAt state in putOnAir and takeOfAir

LindvedKrvang (changes requested):
The Rundown is handling the "state" of the Segments `executedAt`.
The logic for this is basically when the Segment is put onAir, set the `executedAt`. When it's taken offAir, clear the `executedAt`.

There is no reason why the Segment can't handle that in its `putOnAir` and `takeOnAir` methods. We are still using object oriented programming, so lets the objects handle their own state :)

src/model/entities/rundown.ts
KvelaGorrrrnio: You are absolutely right. It is updated :)

src/model/entities/segment.ts
KvelaGorrrrnio: These no longer exist.

src/model/entities/segment.ts
```diff
@@ -55,17 +55,11 @@ export class Segment {
 
   public putOnAir(): void {
     this.isSegmentOnAir = true
+    this.executedAtEpochTime ??= Date.now()
```
LindvedKrvang, line 58: Wouldn't you need to always update the `executedAt` whenever the Segment comes on Air?
I believe this is guarding against if `putOnAir` is called multiple times without being taken off Air. But that sounds more like that `putOnAir` should have an `assertNotOnAir` guard instead of this.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/model/entities/segment.ts
KvelaGorrrrnio: Correct. I have updated both putOnAir and takeOffAir to assert that they are off air and on air respectively when called.

src/model/entities/segment.ts
```diff
@@ -54,15 +56,29 @@ export class Segment {
...
+  private assertOffAir(): void {
+    if (this.isSegmentOnAir) {
+      throw new OffAirException(`Expected segment '${this.name}' with id '${this.id}' to be off air.`)
+    }
   }
 
   public takeOffAir(): void {
+    this.assertOnAir()
     this.isSegmentOnAir = false
     this.executedAtEpochTime = undefined
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Moved to reset.

src/model/entities/segment.ts
```diff
@@ -54,15 +56,29 @@ export class Segment {
...
+  }
+
+  private assertOffAir(): void {
+    if (this.isSegmentOnAir) {
+      throw new OffAirException(`Expected segment '${this.name}' with id '${this.id}' to be off air.`)
+    }
   }
 
   public takeOffAir(): void {
+    this.assertOnAir()
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Removed.

src/model/entities/rundown.ts
```diff
@@ -325,9 +325,11 @@ export class Rundown extends BasicRundown {
   }
 
   private takeNextCursor(): void {
+    const isLeavingOnAirSegment: boolean = this.activeCursor?.segment.id !== this.nextCursor?.segment.id
```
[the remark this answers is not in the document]
KvelaGorrrrnio: Removed.

LindvedKrvang (approved):

andr9528 (approved):

PR #60, opened 2023-11-28
SOF-1702 Fix slow graphics continue

andr9528:


src/business-logic/services/superfly-timeline-builder.ts
```diff
@@ -230,7 +230,7 @@ export class SuperflyTimelineBuilder implements TimelineBuilder {
           : piece.duration
 
     return {
-      start: piece.getStart() + partCalculatedTimings.delayStartOfPiecesDuration,
```
LindvedKrvang, line 233: A unit test to ensure we don't revert this change in the future would be nice

LindvedKrvang: Interesting, it commented on the old code. 
The unittest is offcourse meant to test the change intoduced in this PR

andr9528: Have added the requested test.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #61, opened 2023-11-29
SOF-1705 Able to clear Pieces on different layers when inserting a Piece on Air

LindvedKrvang:
Calling a "Clear Graphics" Actions now stops any Graphic TimelineObjects being played, so they won't be put on Air again.

Breaking Change: The event `InfiniteRundownPieceAddedEvent` has been replaced with `InfinitePiecesUpdatedEvent`


andr9528 (approved):

RasmusAlbrektsen (approved):

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -232,15 +220,17 @@ export class RundownTimelineService implements RundownService {
     await this.saveRundown(rundown)
   }
 
-  public async insertPieceAsOnAir(rundownId: string, piece: Piece): Promise<void> {
+  public async insertPieceAsOnAir(rundownId: string, piece: Piece, layersToStopPiecesOn: string[] = []): Promise<void> {
     const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
+
+    rundown.stopActivePiecesOnLayers(layersToStopPiecesOn)
```
KvelaGorrrrnio, line 226: If `stopActivePiecesOnLayers` returns whether the infinite pieces was affected, we are able to only call `RundownEventEmitter.emitInfinitePiecesUpdatedEvent` when there is an actual change.

KvelaGorrrrnio: That is a fine solution. My thoughts were that the result would be something like:
```typescript
interface StoppedPiecesResult {
  infinitePieceIds: string[]
  pieceIdsFromOnAirPart: string[]
}
```
It allows for more granular event emitting.

KvelaGorrrrnio (approved):

PR #62, opened 2023-11-30
SOF-1703 Add classes='full' to VCP Piece so Ident won't take precedent

LindvedKrvang:
Our Fullscreen Actions for Viz was missing a `classes=['full']`.

The `classes` attribute is used by the Timeline Resolver to determine enables and can be used like `enable: { while: '!.full' }` which means while there is no objects with the class "full", then put this on the Timeline.

All our Idents is ingested by Blueprints with the above enable. For some reason, if the VCP does not have the `classes=['full']` a command is send to Viz to clear the VCP when it's been taken from an Ident. (I am not entirely sure why, but the working theory is that the Ident clear command takes precedent, but having the class "full" somehow 'cancels' that precedent.)

Ideally, we would like to avoid using the `classes` attribute since it's an extra complication in the Timeline Resolving (it does have its usecases though), but that would be a Blueprints change, that we can do when we move the Ingest Part of Blueprints.

### DIFF
--- a/src/blueprints/tv2/action-factories/tv2-graphics-action-factory.ts
+++ b/src/blueprints/tv2/action-factories/tv2-graphics-action-factory.ts
@@ -227,7 +227,7 @@ export class Tv2GraphicsActionFactory {
 
   private createFullscreenGraphicsPartInterface(graphicsData: Tv2FullscreenGraphicsManifestData, blueprintConfiguration: Tv2BlueprintConfiguration): PartInterface {
     return {
-      id: `fullscreenGraphicsPart_${graphicsData.name}`,
+      id: `fullscreenGraphicsPart_${this.stringHashConverter.getHashedValue(graphicsData.name)}`,
       rundownId: '',
       name: `Full ${graphicsData.name}`,
       segmentId: '',
--- a/src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
+++ b/src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
@@ -130,7 +130,8 @@ export class Tv2VizTimelineObjectFactory implements Tv2GraphicsCommandTimelineOb
         noAutoPreloading: false,
         channelName: EngineName.FULLSCREEN,
         ...this.getFullGraphicOutTransitionProperties(blueprintConfiguration)
-      }
+      },
+      classes: ['full'] // Due to how Idents are ingested by Blueprints, if this is omitted VCPs will disappear when being taken from an Ident.
     }
   }
 


src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
```diff
@@ -130,7 +130,8 @@ export class Tv2VizTimelineObjectFactory implements Tv2GraphicsCommandTimelineOb
         noAutoPreloading: false,
         channelName: EngineName.FULLSCREEN,
         ...this.getFullGraphicOutTransitionProperties(blueprintConfiguration)
-      }
+      },
+      classes: ['full'] // Due to how Idents are ingested by Blueprints, if this is omitted VCPs will disappear when being taken from an Ident.
```
andr9528, line 134: I asked you while copying it over, if this value too should be moved other. You said no at that time. Turns out it was important :)

Happy to know the issue has been fixed.

LindvedKrvang: I remember. Do you also remember why I said to remove it?

src/blueprints/tv2/action-factories/tv2-graphics-action-factory.ts
```diff
@@ -227,7 +227,7 @@ export class Tv2GraphicsActionFactory {
 
   private createFullscreenGraphicsPartInterface(graphicsData: Tv2FullscreenGraphicsManifestData, blueprintConfiguration: Tv2BlueprintConfiguration): PartInterface {
     return {
-      id: `fullscreenGraphicsPart_${graphicsData.name}`,
+      id: `fullscreenGraphicsPart_${this.stringHashConverter.getHashedValue(graphicsData.name)}`,
```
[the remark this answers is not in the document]
LindvedKrvang: Those changes will automatically be propogated to here through the PR we are targetting.
This PR won't be merged untill the target PR has been merged.

RasmusAlbrektsen (approved):

src/blueprints/tv2/timeline-object-factories/tv2-viz-timeline-object-factory.ts
andr9528: Yes somewhat. It was something in the lines of your last line in this PR's description.

KvelaGorrrrnio, line 134: This would be nice to have extracted to a constant :)

KvelaGorrrrnio (approved):

PR #63, opened 2023-11-30
Sof 1644 Use Logger instead of Console

andr9528:
Note: The branch is based on the branch for #61, as there was a few log statements added, that i wanted to avoid missing.

Unsure of if we need to add Log Level environment arguments for our start/watch commands. At the moment it is at the very least running with debug log level, perhabs even trace level, as i have seen debug logs being printed.

Just as an example, this is how the console logs look now at startup.
image

Following whats written by Anders(?) in [Common Yarn Scripts](https://tv2cms.atlassian.net/wiki/spaces/SOF/pages/4366958593/Common+Yarn+scripts) running `set LOG_LEVEL=warn; yarn watch` on windows should run with log level set to warning, unless the log level name is incorrect. This however does not work, as debug/info level logs are still displayed.


src/blueprints/tv2/action-factories/test/tv2-transition-effect-action-factory.spec.ts
```diff
@@ -19,12 +20,14 @@ function createTestee(params?: {
   videoMixerTimelineObjectFactory?: Tv2VideoMixerTimelineObjectFactory,
   casparCgTimelineObjectFactory?: Tv2CasparCgTimelineObjectFactory,
   audioTimelineObjectFactory?: Tv2AudioTimelineObjectFactory,
-  assetPathHelper?: Tv2AssetPathHelper
+  assetPathHelper?: Tv2AssetPathHelper,
+  loggerService?: Tv2LoggerService
```
KvelaGorrrrnio, line 24: I strongly prefer the logger is not suffixed with `Service`.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -50,8 +51,10 @@ export class Tv2TransitionEffectActionFactory {
     private readonly videoMixerTimelineObjectFactory: Tv2VideoMixerTimelineObjectFactory,
     private readonly videoClipTimelineObjectFactory: Tv2VideoClipTimelineObjectFactory,
     private readonly audioTimelineObjectFactory: Tv2AudioTimelineObjectFactory,
+    private readonly loggerService: Tv2LoggerService,
     private readonly assetPathHelper: Tv2AssetPathHelper
   ) {
+    loggerService.tag(Tv2TransitionEffectActionFactory.name)
```
KvelaGorrrrnio, line 57: The tag has no effect in this case, since `.tag` creates a new context for the logger. The correct way to set it is:
```typescript
class MyClass {
  private readonly logger: Tv2Logger

  public constructor(logger: Tv2Logger) {
    this.logger = logger.tag(MyClass.name)
  }
}
```

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -201,7 +204,7 @@ export class Tv2TransitionEffectActionFactory {
   private updateTimelineObjectsWithTransitionEffect(action: Tv2TransitionEffectAction, piece: Piece): Tv2TransitionEffectAction {
     const sourceInput: number | undefined = this.videoMixerTimelineObjectFactory.findProgramSourceInputFromPiece(piece)
     if (!sourceInput) {
-      console.log('Can\'t find a Program SourceInput to put the Transition Effect on')
+      this.loggerService.warn('Can\'t find a Program SourceInput to put the Transition Effect on')
```
KvelaGorrrrnio, line 207: In this task, it would be nice if the messages themself are checked as well along with adding potential data. In this case, program, source input and transition effect should not be capitalized. And information about the action and piece would be nice to have included. So this line would be something like:
```typescript
      this.logger.data({ action, piece }).warn('Can\'t find a program source input to put the transition effect on.')
```

src/presentation/facades/controller-facade.ts
```diff
@@ -22,23 +23,23 @@ export class ControllerFacade {
       ServiceFacade.createRundownService(),
       RepositoryFacade.createRundownRepository(),
       ServiceFacade.createIngestService(),
-      new ExpressErrorHandler()
+      new ExpressErrorHandler(new LoggerService())
```
KvelaGorrrrnio, line 26: There are being created new logger instances in a lot of places. I'd like this to be in one place, such that it can be used as a singleton. Currently, we are logging to the console, but if we were to change the LoggerService to write to a file on disk, all the instances would fight for write access to the file.

src/presentation/index.ts
```diff
@@ -49,7 +50,9 @@ function startSofieServer(): void {
 
 function attachExpressServerToPort(port: number): void {
   new SofieServer().server.listen(port, () => {
-    return console.log(`### Express is listening at http://localhost:${port}`)
+    const loggerService: LoggerService = new LoggerService()
+    loggerService.tag('index')
```
KvelaGorrrrnio, line 54: This is the presentation layer index file, not just index.

src/model/services/logger-service.ts
KvelaGorrrrnio, line 1: The model layer should not have services. I'd rather have that the logger lives at `src/logger.ts`.

KvelaGorrrrnio (commented):
I haven't gone through the changed log statements yet, so this is more on the structure.

src/model/services/logger-service.ts
andr9528: Moved as requested.

src/presentation/facades/controller-facade.ts
andr9528: Has been changed to a singleton

src/blueprints/tv2/action-factories/test/tv2-transition-effect-action-factory.spec.ts
andr9528: `Service` has been removed from both the class name and attributes/argument names.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
andr9528: Tagging are now used correctly.

src/presentation/index.ts
andr9528: Have updated the text for the tag.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
andr9528: Will be going through monday making the changes for this, and potentially added some more log statements here and there. 

For example are we currently logging (info) in the `CachedRundownRepository.getRundown` when we retrieve the rundown from the database, but do not have any similar log in the other Cache versions (Segment, Part, Pieces). Think i am going to add similar log statement there, as debug level. Propably going to go through an check if we have any other similar areas where one of a set is logging something, that the others doesn't. Not aiming to add tons of logs, just so similar ones logs at the same area.

andr9528:
Just as an example, this is how the console logs look now at startup.
image

src/blueprints/tv2/tv2-logger.ts
```diff
@@ -0,0 +1,15 @@
...
+  error(message: string, metadata?: object): void
+  warn(message: string, metadata?: object): void
+  info(message: string, metadata?: object): void
+  debug(message: string, metadata?: object): void
+  trace(message: string, metadata?: object): void
+  metadata(metadata: object): Tv2Logger
+  tag(tag: string): Tv2Logger
+  data(data: unknown): Tv2Logger
+  setLevel(level: LogLevel): void
+}
```
LindvedKrvang, line 15: This interface is already created in the logging library that we are using. There is no need to recreate it.
If the interface of the library change, we won't catch it except for a runtime error.

Just use the interface:
```
import { Logger as MediaLogger } from '@tv2media/logger'

export interface Tv2Logger extends MediaLogger {}
```

src/business-logic/services/timeout-callback-scheduler.ts
```diff
@@ -1,27 +1,29 @@
...
-  public static getInstance(): CallbackScheduler {
+  public static getInstance(logger: ConsoleLogger): CallbackScheduler {
     if (!this.instance) {
-      this.instance = new TimeoutCallbackScheduler()
+      this.instance = new TimeoutCallbackScheduler(logger)
     }
     return this.instance
   }
-
+  private readonly logger: Logger
```
LindvedKrvang, line 17: Mising space between the `getInstance` and the next variable

src/business-logic/services/timeout-callback-scheduler.ts
```diff
@@ -1,27 +1,29 @@
...
   private timeoutIdentifier?: NodeJS.Timeout
 
-  private constructor() {
-    return
+  private constructor(logger: ConsoleLogger) {
+    this.logger = logger.tag(TimeoutCallbackScheduler.name)
```
LindvedKrvang, line 21: Instead of setting the intance variable you can just do:
```
private constructor(private readonly logger: Logger) {
    logger.tag(TimeoutCallbackScheduler.name)
}
```

Note: You have made this class dependent on the concrete implementation `ConsoleLogger` instead of the abstraction: `Logger`. That's a no go.

src/business-logic/services/timeout-callback-scheduler.ts
```diff
@@ -1,27 +1,29 @@
...
+import { Logger } from '../../logger'
 
 const EXECUTE_CALLBACK_DURATION_THRESHOLD_IN_MS: number = 1
 const SCHEDULE_RESOLUTION: number = 2
 
 export class TimeoutCallbackScheduler implements CallbackScheduler {
   private static instance: CallbackScheduler
 
-  public static getInstance(): CallbackScheduler {
+  public static getInstance(logger: ConsoleLogger): CallbackScheduler {
```
LindvedKrvang, line 11: Should be dependent on abstraction, not implementation

LindvedKrvang (changes requested):
A lot of things needs to be addressed.

You have redeclared the logger interface from the logging library (twice). Just use the interface provided from the library. (If the Logging library isn't providing an interface, it really should.)

You have made a lot of classes dependent on the concrete implementation of the logger (`ConsoleLogger`) instead of the abstraction (`Logger`). If we change the implementation, we have to update about as many files as there currently is in this PR.
Example of the dependency of the concrete implementation that you have made:
```
class SomeClass {
  private logger: Logger

  constructor(logger: ConsoleLogger) {
    this.logger = logger
  }
}
```
This should just be:
```
class SomeClass {
  constructor(private readonly logger: Logger)
}
```

`ConsoleLogger.getInstance()` is something from the concrete class. Every class that uses this would need to be changed if we change our implementation of the logger. They should not be dependent on anything from the concrete class.

Making the Logger singleton seems unnecessary and we should generally try and avoid using singletons where possible.

Your `ConsoleLogger` has nothing to do with the `Console` and therefore shouldn't be named `ConsoleLogger`.

src/business-logic/services/timeout-callback-scheduler.ts
andr9528: Added new line

src/console-logger.ts
```diff
@@ -0,0 +1,17 @@
+import { DefaultLogger } from '@tv2media/logger/node'
+import { Logger } from './logger'
+
+export class ConsoleLogger extends DefaultLogger implements Logger {
```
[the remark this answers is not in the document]
andr9528: The default implementation of the logger is a console logger. 
Would you rather have it named `LoggerImplementation`?

src/data-access/facades/repository-facade.ts
```diff
@@ -48,21 +48,30 @@ import { IngestedPart } from '../../model/entities/ingested-part'
...
+    return CachedRundownRepository.getInstance(mongoRundownRepository, ConsoleLogger.getInstance())
+  }
+
+  private static createMongoEntityConverter(): MongoEntityConverter {
+    return new MongoEntityConverter(ConsoleLogger.getInstance())
+  }
+
+  private static getMongoDatabaseInstance(): MongoDatabase {
+    return MongoDatabase.getInstance(ConsoleLogger.getInstance())
   }
```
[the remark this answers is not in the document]
andr9528: Are them being private facade method that bad?

It will mean that for use to add additional arguements to the `MongoEntityConverter` and `MongoDatabase`, we will only need to make the change ones instead of multiple times.

src/blueprints/tv2/tv2-logger.ts
andr9528: The message of the method on this interface have the type string, whereas the interface from the library have the type unknown for messages.

Also there is no interface in the library called `MediaLogger`.

src/presentation/facades/controller-facade.ts
andr9528: We might be able to say YAGNI and not have it singleton at the moment, but as soon as we begin to write logs to a file, we would end up with multiple instances fighting for access to the samme file.

This is atleast how i understood it, when i asked Anders.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
KvelaGorrrrnio: As @LindvedKrvang mentions, we don't need a lot of debug logs, since they are for debugging purposes. We want to keep our logging tight – errors, warnings and info logs are what is used in production and these should have the necessary information to help out debugging an incident.

src/business-logic/services/timeout-callback-scheduler.ts
KvelaGorrrrnio: To elaborate on the invalidity. `logger.tag` creates a new logging context, so the suggestion creates a new logging context with the tag associated, and marks it as garbage instantly.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
andr9528: Have not added any extra log statements as requested.

Should i remove the one in `CachedRundownRepository`?
It is a debug level one, which should only ever be written ones per rundown, so i don't see the harm in keeping it.
It has already revealed the frontend bug regarding cleanup, so we know it is well placed and usefull.

src/blueprints/tv2/action-factories/test/tv2-transition-effect-action-factory.spec.ts
```diff
@@ -19,12 +20,8 @@ function createTestee(params?: {
...
+  assetPathHelper?: Tv2AssetPathHelper,
+  logger?: Tv2ConsoleLogger
 }): Tv2TransitionEffectActionFactory {
-  return new Tv2TransitionEffectActionFactory(
-    params?.videoMixerTimelineObjectFactory ?? instance(mock<Tv2VideoMixerTimelineObjectFactory>()),
-    params?.casparCgTimelineObjectFactory ?? instance(mock(Tv2CasparCgTimelineObjectFactory)),
-    params?.audioTimelineObjectFactory ?? instance(mock<Tv2AudioTimelineObjectFactory>()),
-    params?.assetPathHelper ?? instance(mock(Tv2AssetPathHelper))
-  )
+  return new Tv2TransitionEffectActionFactory(params?.videoMixerTimelineObjectFactory ?? instance(mock<Tv2VideoMixerTimelineObjectFactory>()), params?.casparCgTimelineObjectFactory ?? instance(mock(Tv2CasparCgTimelineObjectFactory)), params?.audioTimelineObjectFactory ?? instance(mock<Tv2AudioTimelineObjectFactory>()), params?.assetPathHelper ?? instance(mock(Tv2AssetPathHelper)), params?.logger ?? instance(mock(Tv2ConsoleLogger)))
```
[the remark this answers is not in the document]
andr9528: Accident due to using refactor to move/add new argument. Has been fixed.

src/blueprints/tv2/tv2-console-logger.ts
```diff
@@ -0,0 +1,17 @@
+import { DefaultLogger } from '@tv2media/logger/node'
+import { Tv2Logger } from './tv2-logger'
+
+export class Tv2ConsoleLogger extends DefaultLogger implements Tv2Logger {
```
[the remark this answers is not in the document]
andr9528: You have made a similar comment on the other similar file, which also somewhat applies here.

andr9528:
Question: Do we want to add `"no-console": "error"` linting rule to backend, just as frontend has?

src/blueprints/tv2/tv2-logger.ts
LindvedKrvang: > An advantage of defining the interface ourselves, is that if we change the logger library or the logger library API changes drastically, we only need to change the implementation of the logger instead of updating all the log statements throughout the codebase.

We wouldn't lose that advantage.
I'm not going to die on this hill, so if this is what the majority is for then lets go with it.

src/console-logger.ts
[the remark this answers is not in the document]
LindvedKrvang: No, I would prefer that it was made clear why this has gotten the name `ConsoleLogger`.
Right now it is because of something implicit in `@tv2media`. 
Optimally, `DefaultLogger` was called `ConsoleLogger` to imply it logs to the console. There is nothing stopping the authors of the library to change the implementation of `DefaultLogger` to something else than console and we would never know

src/data-access/facades/repository-facade.ts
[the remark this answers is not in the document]
LindvedKrvang: It feels wrong to have these as methods. Also, it would be nice to limit the usage of static as much as possible.

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -403,7 +410,7 @@ export class MongoEntityConverter {
         return LookaheadMode.WHEN_CLEAR
       }
       default: {
-        console.log(`### Warning: Found unknown number for LookAhead: ${lookAheadNumber}`)
+        this.logger.warn(`Found unknown number for LookAhead: ${lookAheadNumber}`)
         // TODO: Throw error. Currently we have some misconfiguration that uses an outdated lookAhead mode
         // throw new UnsupportedOperation(`Found unknown number for LookAhead: ${lookAheadNumber}`)
```
LindvedKrvang, line 415: I'm tempted to say that we should just let it throw the error now intead of logging. That way we don't need to parse in a logger to this file and we can remove the misconfigguration from the instances that have it.

LindvedKrvang (commented):
All the places where you instantiate the Logger is still dependent on the implementation of the Logger.
You should make a Facade for it, so you only have to update the method of `createLogger()` when changing the Logger implementation instead of all the places that wants to instantiate it.

andr9528:
> You should make a Facade for it, so you only have to update the method of createLogger() when changing the Logger implementation instead of all the places that wants to instantiate it.

Done

src/data-access/facades/repository-facade.ts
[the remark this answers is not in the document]
andr9528: Thinking abit about it, i more or less just applied the same thing you said regarding creating a facade for creating a logger, just for the creation of a `MongoEntityConverter` and a `MongoDatabase` internally in the facade class.

If the signature for the constructor of those classes change, as they did in the case for this PR, then it will now only be one place to edit, instead of a number of places.

src/blueprints/tv2/action-factories/tv2-video-clip-action-factory.ts
```diff
@@ -56,7 +60,7 @@ export class Tv2VideoClipActionFactory {
   private updateVideoClipAction(action: Action, media?: Media): Action {
     const videoClipAction: Tv2VideoClipAction = action as Tv2VideoClipAction
     if (videoClipAction.metadata.contentType !== Tv2ActionContentType.VIDEO_CLIP) {
-      console.error('Can\'t update VideoClipAction. Action is not a VideoClipAction')
+      this.logger.data({ action, media }).warn('Can\'t update VideoClipAction. Action is not a VideoClipAction')
```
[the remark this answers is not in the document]
andr9528: Throws an exception now.

src/data-access/repositories/mongo/mongo-entity-converter.ts
andr9528: Throws the exception now.

src/console-logger.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: If we want it to be more explicit we can extend from the `NodeEnvironmentLogger` instead and set the console vault explicitly like so:

```typescript
export class ConsoleLogger extends NodeEnvironmentLogger implements Logger {
  private constructor() {
    super([
      new ConsoleVault({
        level: Level.TRACE,
        format: new PlainTextFormat(),
        isFormatLocked: false,
      })
    ])
  }
}
```

src/presentation/emitters/web-socket-event-server.ts
```diff
@@ -43,12 +49,12 @@ export class WebSocketEventServer implements EventServer {
     this.webSocketServer = this.createWebSocketServer(port)
 
     this.webSocketServer.on('connection', (webSocket: WebSocket) => {
-      console.log('### WebSocket successfully registered to Server')
+      this.logger.info('WebSocket successfully registered to Server')
```
KvelaGorrrrnio, line 52: `Server` should be `server`.

src/presentation/express-error-handler.ts
```diff
@@ -3,11 +3,18 @@ import { Exception } from '../model/exceptions/exception'
...
+  private readonly logger: Logger
+
+  constructor(logger: Logger) {
+    this.logger = logger.tag(ExpressErrorHandler.name)
+  }
+
   public handleError(response: Response, exception: Exception): void {
-    console.log(`Caught Exception: "${exception.errorCode}". Message: ${exception.message}`)
-    console.log(exception.stack)
+    this.logger.data(exception.stack).error(`Caught Exception: "${exception.errorCode}". Message: ${exception.message}`)
```
KvelaGorrrrnio, line 17: Can't we just give the exception as data? It would be nice if the error message gives insight in what the context was when this happened in a human readable format. One way to achieve this would be to have the `handleError` method take an extra (required) argument that is the error message.

src/data-access/facades/repository-facade.ts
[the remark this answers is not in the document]
andr9528: Okay that counters my previous argument neatly.

Then i'd like to say that having them reduces duplication, and therefore makes it easier to update, should the class being created need new argument. This was the case for these. I have since then removed reasoning for one of them (MongoEntityConverter) to have this helper method, and have locally mvoed the creation of those back out.

There are at the moment 21 lines in the `RepositoryFacade` that call for a `MongoDatabase`. Without this change, should the `MongoDatabase` need an additional argument, as it did now, then it would require 21 lines to be updated, instead of one. Is this amount of duplication really something you'd rather want?

Regarding the `createExpressErrorHandler` in the `ControllerFacade`, the same thought goes, that it lowers duplication. With this one i even knew that there are two parallel PR which both adds arguments to the creation of `ExpressErrorHandler`, That is currently 5 places that would have had the exact same code snippet, which has been lowered down to 1, that would need changes, should the `ExpressErrorHandler` call for more arguments.

Had there only been 2-3 identical places, i'd not argue it would have been as usefull, as in these remaining instances.

src/presentation/facades/controller-facade.ts
```diff
@@ -25,30 +25,34 @@ export class ControllerFacade {
...
       RepositoryFacade.createRundownRepository(),
       ServiceFacade.createIngestService(),
-      new ExpressErrorHandler(ConsoleLogger.getInstance())
+      ControllerFacade.createExpressErrorHandler()
     )
   }
 
+  private static createExpressErrorHandler(): ExpressErrorHandler {
+    return new ExpressErrorHandler(LoggerFacade.createLogger())
+  }
```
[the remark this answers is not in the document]
andr9528: During our chat, we agreed that it was okay to have the helper methods.

src/blueprints/tv2/action-factories/tv2-video-clip-action-factory.ts
```diff
@@ -56,8 +56,7 @@ export class Tv2VideoClipActionFactory {
   private updateVideoClipAction(action: Action, media?: Media): Action {
     const videoClipAction: Tv2VideoClipAction = action as Tv2VideoClipAction
     if (videoClipAction.metadata.contentType !== Tv2ActionContentType.VIDEO_CLIP) {
-      console.error('Can\'t update VideoClipAction. Action is not a VideoClipAction')
-      return action
+      throw new Tv2IncorrectActionException('Can\'t update VideoClipAction. Action is not a VideoClipAction')
```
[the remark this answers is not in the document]
andr9528: Have updated the exception text to the suggested text, and the name as suggested.

I like the idea of the exception being able to take the action as an argument, just not sure how that should be proccesed further up the inheritance.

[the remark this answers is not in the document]
andr9528: Perhabs our base exception could have an optional unknown data argument, which a logger could then extract and add as data, or metadata, if it exists.

src/presentation/emitters/web-socket-event-server.ts
```diff
@@ -43,12 +49,12 @@ export class WebSocketEventServer implements EventServer {
...
       this.addObserversForWebSocket(webSocket)
     })
 
     this.webSocketServer.on('close', () => {
-      console.log('### Server is closed')
+      this.logger.info('Websocket Server has closed')
```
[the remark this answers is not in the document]
andr9528: Fixed

src/presentation/emitters/web-socket-event-server.ts
```diff
@@ -59,7 +65,7 @@ export class WebSocketEventServer implements EventServer {
     const webSocketServer = new WebSocketServer({ server })
 
     server.listen(port, () => {
-      console.log(`### WebSocketServer started on port: ${port}`)
+      this.logger.info(`WebSocketServer started on port: ${port}`)
```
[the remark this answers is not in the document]
andr9528: Fixed

src/presentation/emitters/web-socket-event-server.ts
```diff
@@ -76,10 +82,10 @@ export class WebSocketEventServer implements EventServer {
 
   public killServer(): void {
     if (!this.webSocketServer) {
-      console.log('### Server is already dead')
+      this.logger.info('Websocket Server is already dead')
       return
     }
-    console.log('### Killing Server')
+    this.logger.info('Killing Websocket Server')
     this.webSocketServer.close()
```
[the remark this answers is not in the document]
andr9528: Fixed & renamed to `stopServer `.

src/blueprints/tv2/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
```diff
@@ -270,12 +276,12 @@ export class Tv2AtemVideoMixerTimelineObjectFactory implements Tv2VideoMixerTime
   public findProgramSourceInputFromPiece(piece: Piece): number | undefined {
     const timelineObject: TimelineObject | undefined = piece.timelineObjects.find(timelineObject => timelineObject.layer === Tv2AtemLayer.PROGRAM)
     if (!timelineObject) {
-      console.log(`Can't update Atem Me Input. No TimelineObject for '${Tv2AtemLayer.PROGRAM}' found on Piece '${piece.id}'.`)
+      this.logger.data({ piece }).warn(`Can't update Atem Me Input. No TimelineObject for '${Tv2AtemLayer.PROGRAM}' found on Piece '${piece.id}'.`)
```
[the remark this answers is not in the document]
andr9528: Updated to suggested message.

src/presentation/emitters/web-socket-event-server.ts
andr9528: Fixed

src/console-logger.ts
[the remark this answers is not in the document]
andr9528: Have applied Anders comment, such that we define the `ConsoleVault` ourselves, and extends from `NodeEnvironmentLogger` instead of `DefaultLogger`. This now makes it obvious, why it is called a `ConsoleLogger`.

andr9528:
> There are now three files lying "loose" in the root of src. It would be nice to put them in a folder that could potentially be named logger.

Done

src/presentation/express-error-handler.ts
andr9528: I simply updated the log statement to remain as close to the original as possible.
Has now been updated to have the complete exception in the data.

Not sure about the idea of the extra required message argument.
The method is only ever called from a catch clause, where anything might have happend to end up there.
The message would not be able to add much more detail, not already included by the logged exception. I guess being told where the call comes from does have some merit, as the tag will just say `<ExpressErrorHandler>``as it is currently.

If an extra argument is to be added, then maybe it should specify who the caller is, to add that to the log. Can the logger have multiple Tags?

src/blueprints/tv2/action-factories/tv2-video-clip-action-factory.ts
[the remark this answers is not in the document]
LindvedKrvang: > Perhabs our base exception could have an optional unknown data argument, which a logger could then extract and add as data, or metadata, if it exists.

YAGNI. No need to overthink it.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/blueprints/tv2/action-factories/tv2-video-clip-action-factory.ts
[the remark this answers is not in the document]
andr9528: I did have a talk with Anders friday, where we indeed angreed that it would be YAGNI. So no data argument has been added.

PR #64, opened 2023-11-30
SOF-1698 Able to get, create, update and delete ActionTriggers.

LindvedKrvang:
SofieServer now supports ActionTriggers

This PR includes:
- ActionTriggerController, -Service and -Repository
- ActionTriggerEvents

### DIFF
--- a/src/presentation/interfaces/action-trigger-event-listener.ts
+++ b/src/presentation/interfaces/action-trigger-event-listener.ts
@@ -1,0 +1,5 @@
+import { ActionTriggerEvent } from '../value-objects/action-trigger-event'
+
+export interface ActionTriggerEventListener {
+  listenToActionTriggerEvents(onActionTriggerEventCallback: (actionTriggerEvent: ActionTriggerEvent) => void): void
+}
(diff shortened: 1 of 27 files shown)


andr9528 (approved):

src/presentation/interfaces/action-trigger-event-listener.ts
```diff
@@ -0,0 +1,5 @@
+import { ActionTriggerEvent } from '../value-objects/action-trigger-event'
+
+export interface ActionTriggerEventListener {
+  listenToActionTriggerEvents(onActionTriggerEventCallback: (actionTriggerEvent: ActionTriggerEvent) => void): void
+}
```
KvelaGorrrrnio, line 5: This interface and the implementing class seems to conceptually implement the observer pattern.
Suggestion: The interface could be suffixed with `Observer` instead of `Listener` and the method renamed to `subscribeToActionTriggerEvents`. It would also streamline the naming across frontend and backend.

KvelaGorrrrnio (approved):
Nice work :rocket:

RasmusAlbrektsen (approved):

src/presentation/interfaces/action-trigger-event-listener.ts
LindvedKrvang: No objections here. I have renamed `RundownEventListener` and `ActionTriggerEventListener` to `RundownEventObserver` and `ActionTriggerEventObserver`

PR #66, opened 2023-12-05
SOF-1666 Changes controller responses to match JSend format.

andr9528:
Backend changes for changes in https://github.com/tv2/ng-sofie/pull/74
I will only merge this, once both PR are approved.

This PR chages the format of the responses from the controllers.


src/presentation/jsend-response-formatter.ts
```diff
@@ -0,0 +1,28 @@
...
+export class JsendResponseFormatter implements HttpResponseFormatter{
+  public buildErrorResponse(message: string, errorCode: ErrorCode): object {
+    return {
+      status: 'error',
+      code: errorCode,
+      message: message,
+    }
+  }
+
+  public buildFailResponse(data: object): object {
```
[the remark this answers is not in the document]
andr9528: I was debating whether to make data optional. We don't currently have anywhere, where we don't give it an object.
I say optional, as it is the JSend formatters job to translate the undefined value into a null value.

As requested i will make it optional.

src/presentation/interfaces/http-response-formatter.ts
```diff
@@ -0,0 +1,7 @@
+import { ErrorCode } from '../../model/enums/error-code'
+
+export interface HttpResponseFormatter {
+  buildSuccessResponse(data: object): object
+  buildFailResponse(data: object): object
+  buildErrorResponse(message: string, errorCode: ErrorCode): object
+}
```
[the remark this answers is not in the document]
andr9528: Has been renamed to `format...`

src/presentation/controllers/action-controller.ts
```diff
@@ -13,7 +14,11 @@ interface ExecuteActionRequestBody {
 @RestController('/actions')
 export class ActionController extends BaseController {
 
-  constructor(private readonly actionService: ActionService, private readonly httpErrorHandler: HttpErrorHandler) {
+  constructor(
+    private readonly actionService: ActionService,
+    private readonly httpErrorHandler: HttpErrorHandler,
+    private readonly responseFormatter: HttpResponseFormatter
```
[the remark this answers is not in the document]
andr9528: Done

andr9528:
Anders and i have discussed a bit back and forth if we want responses to have a wrapping object or not.
Having the wrapping object, means we need a bit more code to extract the desired values in the frontend, but means that we follow the JSend format more closely. Having the wrapping object also allow us to easier send multiple different results if a request calls for it, and keeping it would thus keep an uniform response structure, regardless of returned amount of different results.

An example of where we send multiple different results in one request is the currently unused `/configurations/blueprints`.
```ts
res.send(this.httpResponseFormatter.formatSuccessResponse({
        showStyleConfiguration: configuration.showStyle.blueprintConfiguration,
        studioConfiguration: configuration.studio.blueprintConfiguration
      }))
```

The disussion has mostly been aimed at location where we only send one result.
As an example this is how returning actions are done now with the changes.
```ts
response.send(this.httpResponseFormatter.formatSuccessResponse(
  { actions: actions.map(action => new ActionDto(action)) }
))
```

What Anders suggests is for request like the one above to look like this.
```ts
response.send(this.httpResponseFormatter.formatSuccessResponse(
  actions.map(action => new ActionDto(action))
))
```

For responses where we currently are just sending a small message, which is unused by our Angular Frontend, but are very nice to see in Postman, Anders and I are again at odds.

I like how it is setup right now, as an example executing an action looks like the following with new changes.
```ts
response.send(this.httpResponseFormatter.formatSuccessResponse(
  { message: `Successfully executed action: ${actionId} on Rundown: ${rundownId}` }
))
```

Where as Anders suggest that the same response should just say the following.
```ts
response.send(this.httpResponseFormatter.formatSuccessResponse())
```
Note that a null data object is allowed per the JSend standard, which is what his suggestion will do in the current implementation.

With Anders suggestion you'd have to discern from the returned statuscode and the endpoint being called what happend, instead of the human readable message saying it directly.

What are you thought about this @RasmusAlbrektsen @LindvedKrvang ?

LindvedKrvang:
> What are your thoughts about this @RasmusAlbrektsen @LindvedKrvang ?

I'm with Anders on both issues.
There is no need for the wrapping object since it should (hopefully) be explicit in the URL what is returned. It feels like an extra unnecessary layer to go through.

For the ``{ message: `Successfully executed action: ${actionId} on Rundown: ${rundownId}` }`` the message was there to give a success status. Since we are now using JSend, that message is no longer needed.

src/presentation/interfaces/http-response-formatter.ts
```diff
@@ -0,0 +1,7 @@
+import { ErrorCode } from '../../model/enums/error-code'
+
+export interface HttpResponseFormatter {
+  formatSuccessResponse(data?: object): object
+  formatFailResponse(data?: object): object
+  formatErrorResponse(message: string, errorCode: ErrorCode): object
```
LindvedKrvang, line 6: Since this is most likely going to be mostly used with Exceptions it would be nice to have an overload for this method that took our `Exception` class as an argument.

src/presentation/jsend-response-formatter.ts
```diff
@@ -0,0 +1,26 @@
...
+      status: 'error',
+      code: errorCode,
+      message: message,
+    }
+  }
+
+  public formatFailResponse(data?: object): object {
+    return {
+      status: 'fail',
+      data: data ? data : null
```
LindvedKrvang, line 16: Haven't confirmed it, but this `data ?? null` should give the same result

andr9528: It should indeed, so i have updated the code to use it.

src/presentation/jsend-response-formatter.ts
```diff
@@ -0,0 +1,26 @@
+import { HttpResponseFormatter } from './interfaces/http-response-formatter'
+import { ErrorCode } from '../model/enums/error-code'
+
+export class JsendResponseFormatter implements HttpResponseFormatter{
+  public formatErrorResponse(message: string, errorCode: ErrorCode): object {
+    return {
+      status: 'error',
```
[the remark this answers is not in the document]
andr9528: Enum added and used.

src/presentation/controllers/action-controller.ts
```diff
@@ -38,7 +43,7 @@ export class ActionController extends BaseController {
       const rundownId: string = request.params.rundownId
       const body: ExecuteActionRequestBody = request.body
       await this.actionService.executeAction(actionId, rundownId, body.actionArguments)
-      response.send(`Successfully executed action: ${actionId} on Rundown: ${rundownId}`)
+      response.send(this.httpResponseFormatter.formatSuccessResponse({ message: `Successfully executed action: ${actionId} on Rundown: ${rundownId}` }))
```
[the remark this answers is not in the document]
andr9528: Message has been keept, but without wrapping object.

src/presentation/jsend-response-formatter.ts
```diff
@@ -1,26 +1,37 @@
 import { HttpResponseFormatter } from './interfaces/http-response-formatter'
 import { ErrorCode } from '../model/enums/error-code'
+import { Exception } from '../model/exceptions/exception'
+
+enum JSendStatus {
```
LindvedKrvang, line 5: This seems to be the status of the request so a more apt named could be `RequestStatus`.

Sidenote: What's the difference between `ERROR` and `FAIL` and when should I use one over the other? They seem the same to me. Is error equivalent to status codes 500s and fail is for 300 to 499?

src/presentation/jsend-response-formatter.ts
```diff
@@ -1,26 +1,37 @@
...
+  ERROR = 'ERROR',
+  FAIL = 'FAIL',
+  SUCCESS = 'SUCCESS'
+}
 
 export class JsendResponseFormatter implements HttpResponseFormatter{
```
LindvedKrvang, line 11: Missing space before curly bracket

src/presentation/jsend-response-formatter.ts
andr9528: Updated to be named `RequestStatus`.

To me from reading the [JSend document](https://github.com/omniti-labs/jsend), it sounds like error is for when an actual exception of some kind is raised. On the other hand fail sounds like it should be used if a call was successful, but some form of validation failed. The example on the document returns a fail if a form field fails validation.

Where to draw the line between error and fail, does indeed seem a bit blurry. Not sure if that means we should remove fail, or not?

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #67, opened 2023-12-07
SOF-1723: Simple 'Publish Image' workflow.

KvelaGorrrrnio:
Adds GitHub Action workflow for publishing a new Docker image to Docker Hub with `workflow_dispatch` as the trigger.
For the workflow dispatch information, the branch to publish from should be selected along with the version tag to use.

The version tag should be written as a semantic version (e.g.`46.3.1` or `46.3.1-staging`).

The necessary secrets are added to the repository.

### DIFF
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -1,0 +1,35 @@
+name: 'Publish Image'
+
+env:
+  node-version: '18'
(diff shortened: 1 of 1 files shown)


LindvedKrvang (approved):
Have we given any thoughts to what version we want to start SofieServer out on?

PR #69, opened 2023-12-11
SOF 1662/Ingest Events for Rundown View

RasmusAlbrektsen:
This PR is made in connection to the frontend PR: https://github.com/tv2/ng-sofie/pull/80

The purpose is to add properties needed by the frontend to the DTO'S as well as create "unsynced"-events for parts and segments.


src/model/entities/rundown.ts
```diff
@@ -559,22 +559,23 @@ export class Rundown extends BasicRundown {
    * If the Segment is currently OnAir, the Segment is still removed, but an "unsynced copy" is created of the Segment and added to the Rundown in its place.
    */
   public removeSegment(segmentId: string): Segment | undefined {
-    const segmentToRemove: Segment | undefined = this.segments.find(segment => !segment.isUnsynced() && segment.id === segmentId)
+    let segmentToRemove: Segment | undefined = this.segments.find(segment => !segment.isUnsynced() && segment.id === segmentId)
     if (!segmentToRemove) {
       return
     }
 
+    // TODO: Is it okay that i moved filtering up here? Otherwise we crash since we are looking for the unsynced segment id that has yet to exist.
```
RasmusAlbrektsen, line 567: @LindvedKrvang I could use your opinion on this, i don't know if this could cause any issues elsewhere?

andr9528 (approved):

src/business-logic/services/interfaces/rundown-event-emitter.ts
```diff
@@ -21,8 +21,10 @@ export interface RundownEventEmitter {
   emitSegmentCreated(rundown: Rundown, segment: Segment): void
   emitSegmentUpdated(rundown: Rundown, segment: Segment): void
   emitSegmentDeleted(rundown: Rundown, segmentId: string): void
+  emitSegmentUnsynced(rundown: Rundown, unsyncedSegment: Segment, originalSegmentId: string): void
 
   emitPartCreated(rundown: Rundown, part: Part): void
   emitPartUpdated(rundown: Rundown, part: Part): void
-  emitPartDeleted(rundown: Rundown, partId: string): void
+  emitPartDeleted(rundown: Rundown, segmentId: string, partId: string): void
+  emitPartUnsynced(rundown: Rundown, part: Part): void
```
LindvedKrvang, line 29: It's a bit weird that the parameters for `emitSegmentUnsynced` and `emitPartUnsynced` don't follow the same "pattern" since the names of the methods are so similar. 
It would be nice if there was some familiarity between the two to avoid potential confusion.

src/model/entities/rundown.ts
LindvedKrvang: It should be fine, but you could use the parameter `segmentId` as the filter instead of the the returned Segment from `this.unsyncSegment()`.

You could also do the thing we talked about.

src/presentation/services/rundown-event-service.ts
```diff
@@ -114,6 +111,10 @@ export class RundownEventService implements RundownEventEmitter, RundownEventObs
     const event: SegmentDeletedEvent = this.rundownEventBuilder.buildSegmentDeletedEvent(rundown, segmentId)
     this.emitRundownEvent(event)
   }
+  public emitSegmentUnsynced(rundown: Rundown, unsyncedSegment: Segment, originalSegmentId: string): void {
```
LindvedKrvang, line 114: A newline between methods is greatly appreciated.
Goes for the two other methods in this file that are also missing a new line.

src/business-logic/services/database-change-ingest-service.ts
```diff
@@ -355,6 +355,7 @@ export class DatabaseChangeIngestService implements IngestChangeService {
       if (!removedSegment.isUnsynced()) {
         this.eventEmitter.emitSegmentDeleted(rundown, removedSegment.id)
       }
+      this.eventEmitter.emitSegmentUnsynced(rundown, removedSegment, segmentId)
```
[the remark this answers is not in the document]
RasmusAlbrektsen: This has been changed to only emit unsynced events when needed!

src/model/entities/part.ts
```diff
@@ -128,11 +128,12 @@ export class Part {
     this.markAsUnsynced()
   }
 
-  public markAsUnsynced(): void {
+  public markAsUnsynced(): Part {
     this.isPartUnsynced = true
     this.rank = this.rank - 1
     this.pieces.forEach(piece => piece.markAsUnsyncedWithUnsyncedPart())
     this.pieces = this.pieces.map(piece => piece.getUnsyncedCopy())
+    return this
```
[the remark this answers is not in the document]
RasmusAlbrektsen: This was an oversight, it is not needed!

src/business-logic/services/interfaces/rundown-event-emitter.ts
RasmusAlbrektsen: I understand the pattern is off when you look here. 
The issue is the fact that the flow is not the same for segments and parts. Segment are emitted with their unsynced ID. Parts are not. 
I could try and change the flow to fit this if you want me to?

src/presentation/services/rundown-event-service.ts
RasmusAlbrektsen: I think someone else fell into the webstorm-"looks like there is a spaces but its just information above the method"-issue alongside me, almost the whole file had this issue :-)

src/model/entities/rundown.ts
```diff
@@ -564,17 +564,21 @@ export class Rundown extends BasicRundown {
       return
     }
 
+    // We know that 'segmentToRemove' is not undefined at this point
+    this.segments = this.segments.filter(segment => segment.id !== segmentId)
```
andr9528, line 568: You could remove the comment by using the `segmentToRemove.id` to check against. It should have the same outcome.

src/presentation/value-objects/rundown-event.ts
```diff
@@ -4,6 +4,7 @@ import { PartDto } from '../dtos/part-dto'
 import { PieceDto } from '../dtos/piece-dto'
 import { SegmentDto } from '../dtos/segment-dto'
 import { BasicRundownDto } from '../dtos/basic-rundown-dto'
+import {RundownDto} from '../dtos/rundown-dto'
```
andr9528, line 7: Spaces around the imported interface would be appreciated. Really should be something that lint handles for us...

src/business-logic/services/interfaces/rundown-event-emitter.ts
LindvedKrvang: I don't think you need to spend much time on it for now.

src/model/entities/rundown.ts
```diff
@@ -559,16 +559,19 @@ export class Rundown extends BasicRundown {
...
-    let segmentToRemove: Segment | undefined = this.segments.find(segment => !segment.isUnsynced() && segment.id === segmentId)
+    const segmentToRemove: Segment | undefined = this.segments.find(segment => !segment.isUnsynced() && segment.id === segmentId)
     if (!segmentToRemove) {
       return
     }
 
-    // TODO: Is it okay that i moved filtering up here? Otherwise we crash since we are looking for the unsynced segment id that has yet to exist.
     // We know that 'segmentToRemove' is not undefined at this point
-    this.segments = this.segments.filter(segment => segment.id !== segmentToRemove!.id)
+    this.segments = this.segments.filter(segment => segment.id !== segmentId)
```
LindvedKrvang, line 568: You no longer use `segmentToRemove` here, so the comment is outdated and should be deleted.

src/model/entities/rundown.ts
```diff
@@ -559,16 +559,19 @@ export class Rundown extends BasicRundown {
...
+
     if (segmentToRemove.isOnAir()) {
-      segmentToRemove = this.unsyncSegment(segmentToRemove)
+      const unsyncedSegment: Segment = this.unsyncSegment(segmentToRemove)
+      this.updateNextCursor()
+      return unsyncedSegment
+
```
LindvedKrvang, line 574: Unnecessary line

src/presentation/services/event-builder.ts
```diff
@@ -128,7 +129,8 @@ export class EventBuilder implements RundownEventBuilder, ActionTriggerEventBuil
       type: IngestEventType.RUNDOWN_CREATED,
       timestamp: Date.now(),
       rundownId: rundown.id,
-      basicRundown: new BasicRundownDto(rundown)
+      basicRundown: new BasicRundownDto(rundown),
+      rundown: new RundownDto(rundown)
```
LindvedKrvang, line 133: Do you still need the `basicRundown`? All the values that the BasicRundown have is also in the Rundown

src/model/entities/rundown.ts
RasmusAlbrektsen: This has been removed!

src/model/entities/rundown.ts
RasmusAlbrektsen: Removed!

src/presentation/services/event-builder.ts
RasmusAlbrektsen: Nope, i removed it, and added the changes to the frontend as well!

src/model/entities/rundown.ts
RasmusAlbrektsen: It was to avoid a possibly undefined id.

src/presentation/value-objects/rundown-event.ts
RasmusAlbrektsen: I try to keep up with it, but until the lint takes care of it we'll see these issues.

LindvedKrvang (approved):

PR #73, opened 2023-12-15
SOF-1760 Parts are also updated on "Reload iNews data"

LindvedKrvang:
Note: This is dependent on PR https://github.com/tv2/sofie-server/pull/69

Updating a Segment from IngestedSegment now also update the Parts of the Segment, if the Part belongs in the IngestedSegment.


k-primeholding-dim (approved):
verified changes and tests still pass

RasmusAlbrektsen (approved):

PR #74, opened 2023-12-18
Sof 1753/actions knows whether they require arguments

LindvedKrvang:
It is now possible to define on an Action whether it wants arguments.

This PR also includes:
- FadeAudioBed action now tells that it want arguments.
- TransitionEffect actions has been refactored so that Mix and Dip Actions takes an argument for the transition duration instead of having to configure each option in the configuration - this results in the generation of fewer Actions.
- We no longer create Cut transition effect Actions since this is the default behaviour of a Take, so no Action is needed for this.
- The "Transitions" table of the ShowStyle blueprint configuration is now exclusive for definining Breaker effect Actions. NOTE: This means this table needs to have removed all entries previously meant for Mix and Dip Actions


RasmusAlbrektsen (approved):

k-primeholding-dim (approved):
was able to run tests locally with success.

PR #75, opened 2023-12-18
SOF-1727 - Only emit event on change

k-primeholding-dim:
Capturing InfinitePieces before and after an action (in question) might yield a diff - that corresponds to appropriate event being emitted. Introduces lodash.isequal as dev deps used in comparison.


src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -39,12 +40,14 @@ export class RundownTimelineService implements RundownService {
   public async activateRundown(rundownId: string): Promise<void> {
     await this.assertNoRundownIsActive()
     const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
-
+    const snapBefore = rundown.getInfinitePiecesMap()
```
RasmusAlbrektsen, line 43: I would name these constants something along the line of `infinitePiecesBeforeActivation`. Could you add types to these constants as well? :-)

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -39,12 +40,14 @@ export class RundownTimelineService implements RundownService {
...
     rundown.activate()
 
     await this.buildAndPersistTimeline(rundown)
-
-    this.rundownEventEmitter.emitInfinitePiecesUpdatedEvent(rundown)
+    const snapAfter = rundown.getInfinitePiecesMap()
+    if (isDeepStrictEqual(snapBefore, snapAfter)) {
```
RasmusAlbrektsen, line 48: Am i mistaken or do we not want it to be:
`if (!isDeepStrictEqual(snapBefore, snapAfter)` ?

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -103,15 +106,18 @@ export class RundownTimelineService implements RundownService {
...
     rundown.takeNext()
     rundown.getActivePart().setEndState(this.getEndStateForActivePart(rundown))
 
     const timeline: Timeline = await this.buildAndPersistTimeline(rundown)
 
     this.startAutoNext(timeline, rundownId)
 
-    this.rundownEventEmitter.emitInfinitePiecesUpdatedEvent(rundown)
+    const snapAfter = rundown.getInfinitePiecesMap()
+    if (isDeepStrictEqual(snapBefore, snapAfter)) {
```
RasmusAlbrektsen, line 118: Same comment about negation as before, am i mistaken or will we not just get an update event whenever nothing changes?

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -77,6 +82,97 @@ describe(RundownTimelineService.name, () => {
...
+      await testee.activateRundown('aRundown')
+      verify(mockRundownEventEmitter.emitInfinitePiecesUpdatedEvent(aRundown)).once()
+      verify(mockRundownEventEmitter.emitActivateEvent(aRundown)).once()
+      verify(mockRundownEventEmitter.emitSetNextEvent(aRundown)).once()
+    })
+  })
+
+  describe(`${RundownTimelineService.prototype.takeNext.name}`, () => {
+    it('does not emit events unless piecess are changed on takeNext', async () => {
+      const activePiece: Piece = EntityTestFactory.createPiece({ id: 'activePiece' })
```
LindvedKrvang, line 111: Your test is passing, but the implementation in `RundownTimelineService` is checking on the wrong conditional (mentioned in a previous comment). 
The reason your test is still parsing is because a bug (that this test just discovered for us) in `EntityTestFactory.createPiece()`. `createPiece()` does currently not set a default `PieceLifespan` for the Piece. So unless you provide a `PieceLifespan` it will be undefined.
This result in the Rundown thinking that the Piece is an infinite Piece because the Rundown checks `piece.pieceLifespan !== PieceLifspan.WITHIH_PART`, which is true then the lifespan is undefined.

To fix the bug: `pieceLifespan: PieceLifespan.WITHIN_PART` should be added to the `createPiece` method in `EntityTestFactory` so it would look like this:
image

Adding this to `createPiece` will cause your test to fail because now there will be an emitted `infinitePiecesUpdatedEvent` since the current implementation emits when the InfinitePieces array is the same before and after the operation which they will be now, since there is no infinitePieces in the Rundown before activation and there is no infinite Pieces in the Rundown after activation.

LindvedKrvang (commented):
It would be nice if you would name your branch names a bit more than just the Jira number :)

src/business-logic/services/rundown-timeline-service.ts
LindvedKrvang, line 48: You don't need to do a deep equals for this.
We just want to know if the Pieces in the Map before are the same as the Pieces in the Map after.
So if the length differs, we know an update happened and we need to emit.
If the length doesn't differ, then we need to check if it's the same Pieces that are in the Map and for that we can just use the ids.

And example to check for ids could look like this:
image

There is some looping so it is not necessarily great, but it's only looping over small string arrays instead of doing a deep comparison.

k-primeholding-dim: I believe order makes no diff

src/business-logic/services/rundown-timeline-service.ts
k-primeholding-dim: ok, Now I see the 'negation' you are talking about here and above!

nodemon.json
```diff
@@ -2,6 +2,6 @@
   "watch": ["src"],
   "ext": "ts,json",
   "ignore": ["src/**/*.spec.ts"],
-  "exec": "cross-env TS_NODE_PROJECT=tsconfig.development.json node --inspect=0.0.0.0:9229 --require ts-node/register src"
+  "exec": "cross-env TS_NODE_PROJECT=tsconfig.development.json node --dns-result-order=ipv4first --inspect=0.0.0.0:9229 --require ts-node/register src"
```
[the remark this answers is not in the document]
k-primeholding-dim: App binds to [::1] instead of ipv4 stack address.

package.json
```diff
@@ -20,6 +20,7 @@
     "@types/express": "^4.17.17",
     "@types/got": "^9.6.12",
     "@types/jest": "^29.5.4",
+    "@types/lodash.isequal": "^4.5.8",
```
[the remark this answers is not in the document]
k-primeholding-dim: You are correct.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -68,6 +70,17 @@ export class RundownTimelineService implements RundownService {
...
+    if ( first !== null && second != null){
+      if (first.size != second.size) return true
+      first.forEach((firstPiece, firstMapKey)=>{
+        const secondPiece: Piece | undefined = second.get(firstMapKey)
+        if (secondPiece !== undefined && secondPiece.id !== firstPiece.id) return true
+        else return true
+      })
+    }
+    return false
+  }
```
RasmusAlbrektsen, line 83: From my understanding the return will only break out of the inner loops and still return false when we should return true inside the forEach.  So if i have two maps of the same size butwith different content, it would still return false. 
I came up with a suggestion that i think covers the scenarios, please do look it over and see if you agree :-) 

``` 
  private doInfinitePiecesMapsDiffer(first: Map<string, Piece>, second: Map<string, Piece>): boolean {
    if (first === null && second === null) {
      return false
    }

    if (first.size !== second.size) {
      return true
    }

    for (const [firstMapKey, firstPiece] of first) {
      const secondPiece: Piece | undefined = second.get(firstMapKey)
      if (secondPiece === undefined || secondPiece.id !== firstPiece.id) {
        return true
      }
    }
    return false
  }

Comment (KvelaGorrrrnio) [reply src/business-logic/services/rundown-timeline-service.ts]: The type signature of the method does not take a null values, thus the null check can be removed. And instead of using the for-loop, you can you use `Array.some`. For better understanding what the maps are, they can be renamed to something like `firstInfinitePieceMap` and `secondInfinitePieceMap`. An example can be seen below:
```typescript
  private doInfinitePieceMapsDiffer(firstInfinitePieceMap: Map<string, Piece>, secondInfinitePieceMap: Map<string, Piece>): boolean {
    if (firstInfinitePieceMap.size !== secondInfinitePieceMap.size) {
      return true
    }
    return [...firstInfinitePieceMap.entries()].some(([layer, piece]) => secondInfinitePieceMap.get(layer)?.id !== piece.id)
  }
```

RasmusAlbrektsen (approved):

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -68,6 +70,12 @@ export class RundownTimelineService implements RundownService {
     return timeline
   }
 
+  private doInfinitePieceMapsDiffer(firstInfinitePieceMap: Map<string, Piece>, secondInfinitePieceMap: Map<string, Piece>): boolean {
+    if (firstInfinitePieceMap.size !== secondInfinitePieceMap.size) {
+      return true
+    }
+    return [...firstInfinitePieceMap.entries()].some(([layer, piece]) => secondInfinitePieceMap.get(layer)?.id !== piece.id)
+  }
   public async deactivateRundown(rundownId: string): Promise<void> {
```
LindvedKrvang, line 79: A single newline between methods is much appreciated :)

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -103,15 +111,18 @@ export class RundownTimelineService implements RundownService {
     this.stopAutoNext()
 
     const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
-
+    const infinitePiecesBeforeStartAutoNext: Map<string,Piece> = rundown.getInfinitePiecesMap()
```
LindvedKrvang, line 114: `startAutoNext` has nothing to do with updating the infinite Pieces so it shouldn't be in the variable name. It's `rundown.takeNext()` that changes the infinite Pieces of the Rundown.
The `startAutoNext()` just starts a timer, that will execute the `takeNext()` method again which will have its own before and after check for the infinite Pieces.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -39,12 +39,14 @@ export class RundownTimelineService implements RundownService {
...
     await this.assertNoRundownIsActive()
     const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
-
+    const infinitePiecesBeforeActivation: Map<string,Piece> = rundown.getInfinitePiecesMap()
     rundown.activate()
 
     await this.buildAndPersistTimeline(rundown)
-
-    this.rundownEventEmitter.emitInfinitePiecesUpdatedEvent(rundown)
+    const infinitePiecesAfterActivation: Map<string,Piece> = rundown.getInfinitePiecesMap()
```
LindvedKrvang, line 46: Minor thing, but it would be nice to have a space after the comma and before the Piece in `Map<string,Piece>`, so it becomes `Map<string, Piece>`.
It makes it easier to read when everything isn't cramped :)

LindvedKrvang (approved):

k-primeholding-dim:
Closing this PR due to unsigned commits in history, and opening https://github.com/tv2/sofie-server/pull/81 as followup

PR #76, opened 2023-12-20
SOF-1754 - Reset configuration cache with an endpoint request

k-primeholding-dim:
We should be able to reset configuration cache without restarting the application.
For this - here I'm adding endpoint that will set cached configuration to undefined.
[WIP]
Also a button in config page that performs POST request to the reset endpoint.


k-primeholding-dim:
Corresponds with https://github.com/tv2/ng-sofie/pull/84 as frontend implementation of the task.

src/presentation/controllers/configuration-controller.ts
```diff
@@ -43,4 +43,14 @@ export class ConfigurationController extends BaseController {
     }
   }
 
+  @PostRequest('/reset')
```
RasmusAlbrektsen, line 46: A minor note here. `/reset` in the context of configurations could sound like returning what has been configured to defaults rather than clearing the cache.

RasmusAlbrektsen (approved):

src/data-access/repositories/test/mongo-configuration-repository.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+describe(MongoConfigurationRepository.name, () => {
+  describe(MongoConfigurationRepository.prototype.clearConfigurationCache.name, () => {
+    it('throws error when invoked', () => {
+      const mockStudioRepository: StudioRepository = mock<StudioRepository>()
+      const mockShowStyleRepository: ShowStyleRepository = mock<ShowStyleRepository>()
+      const aMongoConfigurationRepository: MongoConfigurationRepository = new MongoConfigurationRepository(
+        instance(mockStudioRepository),
+        instance(mockShowStyleRepository)
+      )
+      expect(()=>aMongoConfigurationRepository.clearConfigurationCache()).toThrow('Method not applicable.')
```
LindvedKrvang, line 15: Would be nice to have some spaces around `()=>` so it becomes `() => aMongoConfigurationRepository...` :)

LindvedKrvang (approved):
Looks good. Mainly minor things :)

k-primeholding-dim:
follow-up PR is https://github.com/tv2/sofie-server/pull/79

PR #80, opened 2024-01-09
Sof 1697 name and description chance in mix and dip transition

vivanov90:
We had 2 mix and dip transitions, they was with same name and description and there was no way to tell which is which in the UI


LindvedKrvang (approved):
Congratulations on your first backend changes in SofieServer :D

KvelaGorrrrnio (approved):
:rocket:

PR #82, opened 2024-01-15
Sof 1719/update segments with mini shelf

LindvedKrvang:
- Segments now exposes `isHidden`
- Segments now exposes a `metadata` field
- Sofie Server now loads all "hidden" Segments and Segments with no Parts and sends them to the frontend so the frontend can decide what to do with them.
  - Sofie Server now supports empty and hidden Segments when executing a Rundown (they are ignored in the execution and they are not possible to be Taken)

### DIFF
--- a/src/business-logic/services/ingested-entity-to-entity-mapper.ts
+++ b/src/business-logic/services/ingested-entity-to-entity-mapper.ts
@@ -53,6 +53,7 @@ export class IngestedEntityToEntityMapper {
       name: ingestedSegment.name,
       rank: ingestedSegment.rank,
       isHidden: ingestedSegment.isHidden,
+      metadata: ingestedSegment.metadata,
       isOnAir: false,
       isNext: false,
       isUnsynced: false,
--- a/src/business-logic/services/ingested-entity-to-entity-mapper.ts
+++ b/src/business-logic/services/ingested-entity-to-entity-mapper.ts
@@ -52,6 +52,7 @@ export class IngestedEntityToEntityMapper {
       rundownId: ingestedSegment.rundownId,
       name: ingestedSegment.name,
       rank: ingestedSegment.rank,
+      isHidden: ingestedSegment.isHidden,
       isOnAir: false,
       isNext: false,
       isUnsynced: false,
--- a/src/model/entities/rundown.ts
+++ b/src/model/entities/rundown.ts
@@ -1,7 +1,5 @@
 import { Segment } from './segment'
 import { Part } from './part'
-import { Exception } from '../exceptions/exception'
-import { ErrorCode } from '../enums/error-code'
 import { LastPartInSegmentException } from '../exceptions/last-part-in-segment-exception'
 import { NotFoundException } from '../exceptions/not-found-exception'
 import { NotActivatedException } from '../exceptions/not-activated-exception'
@@ -18,7 +16,7 @@ import { UnsupportedOperationException } from '../exceptions/unsupported-operati
 import { RundownCursor } from '../value-objects/rundown-cursor'
 import { Owner } from '../enums/owner'
 import { AlreadyExistException } from '../exceptions/already-exist-exception'
-import { LastSegmentInRundown } from '../exceptions/last-segment-in-rundown'
+import { LastSegmentInRundownException } from '../exceptions/last-segment-in-rundown-exception'
 import { NoPartInHistoryException } from '../exceptions/no-part-in-history-exception'
 import { OnAirException } from '../exceptions/on-air-exception'
 import { RundownTiming } from '../value-objects/rundown-timing'
@@ -119,10 +117,7 @@ export class Rundown extends BasicRundown {
 
   private findFirstSegment(): Segment {
     return this.segments
-      .filter(segment => this.isSegmentValidForRundownExecution(segment))
-      .reduce((previousSegment: Segment, currentSegment: Segment) => {
-        return previousSegment.rank < currentSegment.rank ? previousSegment : currentSegment
-      })
+      .filter(segment => this.isSegmentValidForRundownExecution(segment))[0]
   }
 
   private isSegmentValidForRundownExecution(segment: Segment): boolean {
@@ -141,7 +136,7 @@ export class Rundown extends BasicRundown {
       const nextSegment: Segment | undefined = this.segments.find(segment => segment.id === nextPart.getSegmentId())
       this.nextCursor = this.createCursor(this.nextCursor, { segment: nextSegment, part: nextPart, owner })
     } catch (exception) {
-      if ((exception as Exception).errorCode !== ErrorCode.LAST_PART_IN_SEGMENT) {
+      if (!(exception instanceof LastPartInSegmentException)) {
         throw exception
       }
       this.unmarkNextSegment()
@@ -151,7 +146,7 @@ export class Rundown extends BasicRundown {
         this.nextCursor = this.createCursor(this.nextCursor, { segment, part: segment.findFirstPart(), owner })
         this.markNextSegment()
       } catch (err) {
-        if ((err as Exception).errorCode !== ErrorCode.LAST_SEGMENT_IN_RUNDOWN) {
+        if (!(err instanceof LastSegmentInRundownException)) {
           throw err
         }
         // TODO: Notify about last Segment.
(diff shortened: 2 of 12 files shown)


src/model/entities/rundown.ts
```diff
@@ -118,9 +118,15 @@ export class Rundown extends BasicRundown {
...
 
   private findFirstSegment(): Segment {
-    return this.segments.reduce((previousSegment: Segment, currentSegment: Segment) => {
-      return previousSegment.rank < currentSegment.rank ? previousSegment : currentSegment
-    })
+    return this.segments
+      .filter(segment => this.isSegmentValidForRundownExecution(segment))
+      .reduce((previousSegment: Segment, currentSegment: Segment) => {
+        return previousSegment.rank < currentSegment.rank ? previousSegment : currentSegment
+      })
```
KvelaGorrrrnio, line 125: There are other places in the the rundown entity (e.g. `findNextValidSegment`) where we assume that the segments list is ordered correctly – why can't we do that here?

RasmusAlbrektsen (approved):
Based on the comments from @KvelaGorrrrnio being resolved i approve! :-D

src/model/entities/rundown.ts
```diff
@@ -178,15 +189,17 @@ export class Rundown extends BasicRundown {
...
     }
-    if (activeSegmentIndex === this.segments.length + 1) {
-      throw new LastSegmentInRundown(`Segment: ${this.activeCursor?.segment?.id} is the last Segment of Rundown: ${this.id}`)
+    for (let i = activeSegmentIndex; i < this.segments.length - 1; i++) {
+      if (this.isSegmentValidForRundownExecution(this.segments[i + 1])) {
+        return this.segments[i + 1]
+      }
     }
-    return this.segments[activeSegmentIndex + 1]
+    throw new LastSegmentInRundown(`Segment: ${this.activeCursor?.segment?.id} is the last Segment of Rundown: ${this.id}`)
```
KvelaGorrrrnio, line 202: In the exception message, it is stated that it is the last segment in the rundown, but it is the last _valid_ segment in the rundown – minor thing :) Personally, I would like our error messages to be somewhat written out like our construct naming conventions, to ensure a lower level of ambiguity.

The following is just a suggestion. One could use `Array.slice` and `Array.find` instead of the for-loop to find the next valid segment.

```typescript
const nextValidSegment: Segment | undefined = this.segments.slice(activeSegmentIndex + 1).find(segment => this.isSegmentValidForRundownExecution(segment))
if (!nextValidSegment) {
  throw new LastSegmentInRundown(`The segment '${this.activeCursor?.segment.name}' with the id '${this.activeCursor?.segment.id}' is the last valid segment in the Rundown '${this.name}' with the id '${this.id}'.`)
}
return nextValidSegment
```

src/business-logic/services/ingested-entity-to-entity-mapper.ts
```diff
@@ -52,6 +52,8 @@ export class IngestedEntityToEntityMapper {
       rundownId: ingestedSegment.rundownId,
       name: ingestedSegment.name,
       rank: ingestedSegment.rank,
+      isHidden: ingestedSegment.isHidden,
```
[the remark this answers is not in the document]
LindvedKrvang: It's basically whether and not to show it. It's possible for a Segment to be perfectly "valid" and "executable", but still shouldn't be shown nor included in the execution (floating a Segment in iNews for example).

src/model/entities/rundown.ts
```diff
@@ -139,11 +145,16 @@ export class Rundown extends BasicRundown {
...
+      // TODO: Refactor to use less nesting
+      try {
+        const segment: Segment = this.findNextValidSegment()
         this.nextCursor = this.createCursor(this.nextCursor, { segment, part: segment.findFirstPart(), owner })
         this.markNextSegment()
+      } catch (err) {
+        if ((err as Exception).errorCode !== ErrorCode.LAST_SEGMENT_IN_RUNDOWN) {
+          throw err
+        }
+        // TODO: Notify about last Segment.
```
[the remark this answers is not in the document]
LindvedKrvang: Technically yes, but if we don't catch the exception here we are not able to Take the last Segment.
But in general we need better handling around the last Segment, because with this implementation we are not able to leave the Last Segment again...

I'm going to write a bug task on that.

src/model/entities/rundown.ts
```diff
@@ -139,11 +145,16 @@ export class Rundown extends BasicRundown {
...
-      const segment: Segment = this.findNextSegment()
-      // TODO: Handle that we might be on the last Segment
-      if (segment) {
+      // TODO: Refactor to use less nesting
+      try {
+        const segment: Segment = this.findNextValidSegment()
         this.nextCursor = this.createCursor(this.nextCursor, { segment, part: segment.findFirstPart(), owner })
         this.markNextSegment()
+      } catch (err) {
+        if ((err as Exception).errorCode !== ErrorCode.LAST_SEGMENT_IN_RUNDOWN) {
```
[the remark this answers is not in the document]
LindvedKrvang: I have updated the two usages of it here. I will try and remember it in the future

src/model/exceptions/last-segment-in-rundown.ts
```diff
@@ -3,6 +3,6 @@ import { ErrorCode } from '../enums/error-code'
 
 export class LastSegmentInRundown extends Exception {
```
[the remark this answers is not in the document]
LindvedKrvang: Updated

src/model/entities/rundown.ts
LindvedKrvang: I think that is a leftover from the initial implementation and is no longer needed since we keep the Segments sorted.
Updated to 
```
return this.segments.filter(segment => this.isSegmentValidForRundownExecution(segment))[0]
```
which is not great either incase there are no Segments. 
Suggestions are appreciated.

src/model/entities/rundown.ts
LindvedKrvang: Updated to the `Array.slice` and `Array.find` suggestion.

Changed the error message to include "valid Segment".

k-primeholding-dim (approved):

src/business-logic/services/ingested-entity-to-entity-mapper.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: It could also be `isIgnored`. I am not fan of it, since it doesn't properly describe what its purpose is in the backend.
I'll leave it be for now, but I think we should consider a better name for it.

src/model/entities/rundown.ts
```diff
@@ -118,9 +116,12 @@ export class Rundown extends BasicRundown {
   }
 
   private findFirstSegment(): Segment {
-    return this.segments.reduce((previousSegment: Segment, currentSegment: Segment) => {
-      return previousSegment.rank < currentSegment.rank ? previousSegment : currentSegment
-    })
+    return this.segments
+      .filter(segment => this.isSegmentValidForRundownExecution(segment))[0]
```
KvelaGorrrrnio, line 120: This can be rewritten as
```suggestion
    return this.segments
      .find(segment => this.isSegmentValidForRundownExecution(segment))
```

And note that the return type of the method declaration and the implementation are not aligning. The implementation returns `Segment | undefined`.

src/model/entities/rundown.ts
```diff
@@ -135,15 +136,20 @@ export class Rundown extends BasicRundown {
...
-      if (segment) {
+      // TODO: Refactor to use less nesting
+      try {
+        const segment: Segment = this.findNextValidSegment()
         this.nextCursor = this.createCursor(this.nextCursor, { segment, part: segment.findFirstPart(), owner })
         this.markNextSegment()
+      } catch (err) {
+        if (!(err instanceof LastSegmentInRundownException)) {
+          throw err
+        }
```
KvelaGorrrrnio, line 151: It would be nice with `err` written out as `error`.

KvelaGorrrrnio (approved):

src/model/entities/rundown.ts
LindvedKrvang: Updated to use `.find()`.
The return type is still just `Segment`, however, the method throws an exception if it couldn't find a Segment.

PR #83, opened 2024-01-16
Sof 1719/provide mini shelf media information

LindvedKrvang:
All dependent PRs have been merged

- Studio now exposes the `PreviewMediaUrl` needed for rendering thumbnails.
- Exposes endpoints to fetch `Media`

### DIFF
--- a/src/data-access/repositories/interfaces/MediaRepository.ts
+++ b/src/data-access/repositories/interfaces/MediaRepository.ts
@@ -1,5 +1,6 @@
 import { Media } from '../../../model/entities/media'
 
 export interface MediaRepository {
-  getMedia(mediaId: string): Promise<Media | undefined>
+  getMedia(): Promise<Media[]>
+  getMediaById(mediaId: string): Promise<Media | undefined>
 }
--- a/src/data-access/repositories/interfaces/MediaRepository.ts
+++ b/src/data-access/repositories/interfaces/MediaRepository.ts
@@ -2,5 +2,5 @@ import { Media } from '../../../model/entities/media'
 
 export interface MediaRepository {
   getMedia(): Promise<Media[]>
-  getMediaById(mediaId: string): Promise<Media | undefined>
+  getMediaBySourceName(sourceName: string): Promise<Media | undefined>
 }
--- a/src/presentation/controllers/media-controller.ts
+++ b/src/presentation/controllers/media-controller.ts
@@ -1,0 +1,44 @@
+import { BaseController, GetRequest, RestController } from './base-controller'
+import { Request, Response } from 'express'
+import { MediaRepository } from '../../data-access/repositories/interfaces/MediaRepository'
+import { HttpErrorHandler } from '../interfaces/http-error-handler'
+import { HttpResponseFormatter } from '../interfaces/http-response-formatter'
+import { Media } from '../../model/entities/media'
+import { Exception } from '../../model/exceptions/exception'
+import { NotFoundException } from '../../model/exceptions/not-found-exception'
+
+@RestController('/media')
+export class MediaController extends BaseController {
+
+  constructor(
+    private readonly mediaRepository: MediaRepository,
+    private readonly httpErrorHandler: HttpErrorHandler,
+    private readonly httpResponseFormatter: HttpResponseFormatter
+  ) {
+    super()
+  }
+
+  @GetRequest()
+  public async getMedia(_request: Request, response: Response): Promise<void> {
+    try {
+      const media: Media[] = await this.mediaRepository.getMedia()
+      response.send(this.httpResponseFormatter.formatSuccessResponse(media))
+    } catch (error) {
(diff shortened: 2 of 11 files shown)


src/data-access/repositories/interfaces/MediaRepository.ts
```diff
@@ -1,5 +1,6 @@
 import { Media } from '../../../model/entities/media'
 
 export interface MediaRepository {
-  getMedia(mediaId: string): Promise<Media | undefined>
+  getMedia(): Promise<Media[]>
+  getMediaById(mediaId: string): Promise<Media | undefined>
```
RasmusAlbrektsen, line 5: It's a bit of a weird scenario, there is actually a unique ID connected to `mediaObjects` but we also use the term `mediaId` because that is how we identify media by their filename (without the file extension). Technically it would be `getMediaByMediaId`, but i don't mind seeing if we can just make do by abstracting it away like this.

RasmusAlbrektsen (approved):

src/data-access/repositories/interfaces/MediaRepository.ts
LindvedKrvang: What is the difference between `Media.id` and `Media.mediaId`?

RasmusAlbrektsen: It can be seen here, the `media.id` is the collection it stems from suffixed by a unique identifier.
The `media.mediaId` is the filename without the file extension. 

image

RasmusAlbrektsen: Don't mind the `to2` in the second row, that was just me testing some update-change scenarios :-D

LindvedKrvang: It just seems weird that we have an Entity called `Media` that has an attribute that is called `MediaId` but is in fact not the id of the Media...

LindvedKrvang: Is `MediaId` the correct name to use? We do have the option to change it.

RasmusAlbrektsen: After we discussed in a call we enden up deciding to rename `mediaId` to `mediaName`.

vivanov90: Just question:
Even after changing the name of the identify to `mediaName` doesn't the same problem remain, now we have `Media.mediaName` . Is there have any problem to be only `Media.name`?

RasmusAlbrektsen: Also a good point, but in the meantime we renamed it to `sourceName` :-D

LindvedKrvang: I have updated it be using `sourceName` instead `mediaId`, but the change for introducing `sourceName` is in this PR https://github.com/tv2/sofie-server/pull/84

The endpoint still works without the changes of that PR.

src/presentation/dtos/media-dto.ts
```diff
@@ -0,0 +1,12 @@
+import { Media } from '../../model/entities/media'
+
+export class MediaDto {
+
+  public readonly id: string
+  public readonly duration?: number
```
KvelaGorrrrnio, line 6: Why is duration optional?

KvelaGorrrrnio (approved):

src/presentation/controllers/media-controller.ts
```diff
@@ -0,0 +1,45 @@
...
+    private readonly httpResponseFormatter: HttpResponseFormatter
+  ) {
+    super()
+  }
+
+  @GetRequest()
+  public async getMedia(_request: Request, response: Response): Promise<void> {
+    try {
+      const media: Media[] = await this.mediaRepository.getMedia()
+      response.send(this.httpResponseFormatter.formatSuccessResponse(media.map(m => new MediaDto(m))))
```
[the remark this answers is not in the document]
RasmusAlbrektsen: I like the second or third option you provide! Not the first, however, it should not be `medias`.

k-primeholding-dim (approved):

src/presentation/dtos/media-dto.ts
LindvedKrvang: We might not have finished loading the media on the media server and hence not have a duration yet. Atleast that is the idea.
But if Media is also for graphics, then it's not all graphics that have a duration

src/presentation/controllers/media-controller.ts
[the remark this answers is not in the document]
LindvedKrvang: I not a fan of the first option.
Both the second and the third option would be a renaming of our entity which is too large a task for this scope.
I could be up for renaming the `Media` entity, but the reason for doing so has to be a lot better than we use a `m` in an arrow function somewhere :)

PR #85, opened 2024-01-23
SOF-1791 Return correct type and format for Media.duration

LindvedKrvang:
The duration for Media is saved as a string in the database.

Media.duration is now an actual number and the value is in milliseconds when fetching it.

### DIFF
--- a/src/data-access/repositories/mongo/mongo-entity-converter.ts
+++ b/src/data-access/repositories/mongo/mongo-entity-converter.ts
@@ -155,10 +155,12 @@ export interface MongoMedia extends MongoId {
   mediaId: string
   mediainfo?: {
     format?: {
-      duration?: number
+      duration?: string
     }
   }
 }
+
+const MILLISECONDS: number = 1000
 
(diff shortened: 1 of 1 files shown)


k-primeholding-dim (approved):

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -155,11 +155,13 @@ export interface MongoMedia extends MongoId {
   mediaId: string
   mediainfo?: {
     format?: {
-      duration?: number
+      duration?: string
     }
   }
 }
 
+const MILLISECONDS: number = 1000
```
KvelaGorrrrnio, line 163: This could contain two milliseconds. Personally I don't think it is necessary having this in a constant. But otherwise it could say `ONE_SECOND_IN_MS` or `ONE_THOUSAND_MILLISECONDS` or `MILLISECONDS_TO_SECONDS_RATIO`.

KvelaGorrrrnio (approved):

PR #88, opened 2024-01-29
SOF-1792 Add/Remove Pieces when Segments/Parts are being updated.

LindvedKrvang:
Pieces are now being correctly added/removed when we receive updates to Segments and/or Parts.

If we receive and Ingest update event for a Segment or Part that had newly added or removed Pieces, these Pieces were either not added or removed. They are now.


RasmusAlbrektsen (approved):

src/data-access/repositories/interfaces/piece-repository.ts
```diff
@@ -5,6 +5,7 @@ export interface PieceRepository {
   getPiecesFromIds(pieceIds: string[]): Promise<Piece[]>
   savePiece(piece: Piece): Promise<void>
   deletePiecesForPart(partId: string): Promise<void>
+  deletePieces(pieceIdsToBeDeleted: string[]): Promise<void>
```
vivanov90, line 8: When the function name is `deletePieces`, is it imperative that the parameter name be `pieceIdsToBeDeleted`. Isn't it clear from the function name that `pieceIds` will be deleted.

src/data-access/repositories/mongo/mongo-piece-repository.ts
```diff
@@ -56,6 +56,10 @@ export class MongoPieceRepository extends BaseMongoRepository implements PieceRe
     }
   }
 
+  public async deletePieces(pieceIdsToBeDeleted: string[]): Promise<void> {
```
vivanov90, line 59: Same as that above.

vivanov90 (approved):

src/data-access/repositories/interfaces/piece-repository.ts
LindvedKrvang: I would agree. I keep going back and forth on it. It's been updated :)

src/data-access/repositories/mongo/mongo-piece-repository.ts
LindvedKrvang: Same answer as above :)

PR #90, opened 2024-01-30
SOF-1686 Move Robot Actions

LindvedKrvang:
Note: Needs https://github.com/tv2/ng-sofie/pull/96 to be merged for the frontend to work.

Blueprints now creates the CallPreset Robot Action


src/blueprints/timeline-state-resolver-types/telemetrics-type.ts
```diff
@@ -0,0 +1,14 @@
+import { TimelineObject } from '../../model/entities/timeline-object'
+import { DeviceType } from '../../model/enums/device-type'
+
+export interface TelemetricsCallPresetTimelineObject extends TimelineObject {
+  content: {
+    deviceType: DeviceType.TELEMETRICS
+    type: TelemetricsType.CALL_PRESET
+    presetShotIdentifier: number[]
```
vivanov90, line 8: I saw that value is `number[]`, but the name looks like is only one value `number`

src/blueprints/tv2/action-factories/tv2-robot-action-factory.ts
```diff
@@ -0,0 +1,115 @@
...
+      argument: {
+        name: 'Preset',
+        description: 'The preset to be called provided as a number',
+        type: ActionArgumentType.NUMBER
+      }
+    }
+  }
+
+  public isRobotAction(action: Tv2Action): action is Tv2RobotAction {
+    return action.metadata.contentType === Tv2ActionContentType.ROBOT
```
vivanov90, line 55: In my experience, I personally like to pass the smallest unit of interest to the function that will use it. In that case I would send only `contentType` as parameter, not full action. Is there a reason to do it this way?

src/blueprints/tv2/action-factories/tv2-robot-action-factory.ts
```diff
@@ -0,0 +1,115 @@
...
+    robotAction.data.pieceInterface = this.createCallPresetPieceInterface(actionArguments)
+    return robotAction
+  }
+
+  private isCallPresetArgumentInteger(callPresetArgument: unknown): callPresetArgument is number {
+    if (!Number.isInteger(callPresetArgument)) {
+      throw new Tv2MisconfigurationException(`The preset to be called must be an integer. ${callPresetArgument} is not an integer`)
+    }
+    return true
+  }
```
vivanov90, line 89: Suggestion:
It might not be a bad idea to have a check for negative values ​​or ones over 100 when it comes to percentages

src/blueprints/tv2/value-objects/tv2-layers.ts
```diff
@@ -21,6 +21,8 @@ export enum Tv2SourceLayer {
   REPLAY_STUDIO_AUXILIARY = 'studio0_aux_studio_screen',
   REPLAY_VIZ_AUXILIARY = 'studio0_aux_viz_full1',
   VIDEO_CLIP = 'studio0_selected_clip',
+
+  ROBOT_CAMERA = 'studio0_robot_camera'
```
vivanov90, line 25: I don't think there is a logical reason for that new line at the top. Enum values are logically related

vivanov90 (approved):
Mostly questions and suggestions

src/blueprints/timeline-state-resolver-types/telemetrics-type.ts
LindvedKrvang: There is a reason for this and it is currently not obvious in the current state of the code (it is the plan to make it obvious once we get to that part of the refactor).

There is something in the Sofie ecosystem called Timeline State Resolver (TSR for short). TSR is responsible for converting our TimelineObjects into actual commands for the different hardware devices. In order for TSR to do this, we must provide data in a structure that TSR knows. 

The `content` object above is defined in the TSR repository (https://github.com/tv2/tv-automation-state-timeline-resolver/blob/develop/packages/timeline-state-resolver-types/src/telemetrics.ts) and we can't deviate from that. If we do, TSR won't be able to translate our data. 

So since TSR currently define `presetShotIdentifier` as `number[]`, we have to do so as well. (We also have to name it the same).

As mentioned earlier, it is the plan to expose types so we don't have to manually maintain the types. But that is currently out of scope.

src/blueprints/tv2/action-factories/tv2-robot-action-factory.ts
LindvedKrvang: In the end it comes down to personal preference, but we are writing object oriented programming, so we should utilize that we have objects. So we can parse our object around instead of breaking down our objects.

This gives, in my opinion, the benefit that the data we are parsing around always corresponds to our entities. We don't need to remember which entity they are a part of because we have the entity.

Another thing I have noticed when people try to provide the smallest information as possible is, that often end up with multiple parameters that could just have been a single parameter.

src/blueprints/tv2/value-objects/tv2-layers.ts
LindvedKrvang: Updated

src/blueprints/timeline-state-resolver-types/telemetrics-type.ts
vivanov90: I understand

RasmusAlbrektsen (approved):

PR #96, opened 2024-02-06
Sof 1749/create crud endpoints for configuration

LindvedKrvang:
It's possible to fetch the Shelf configuration and update it through endpoints.
Notice that there is only a Get and an Update endpoint. That's because we currently only support having one Shelf, so there will always be exactly one Shelf. (The system will create one if it is missing, for example for a brand new installation.)
This means that we don't need endpoints for deleting and creating Shelves, since those usecases doesn't exists.

This also mean that we only need to emit one event, SHELF_UPDATED, which is being emitted when the Update endpoint is called.


src/business-logic/services/configuration-service-implementation.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class ConfigurationServiceImplementation implements ConfigurationService {
+
+  constructor(
+    private readonly configurationEventEmitter: ConfigurationEventEmitter,
+    private readonly shelfRepository: ShelfRepository
+  ) { }
+
+  public async updateShelf(shelf: Shelf): Promise<Shelf> {
+    const updateShelf: Shelf = await this.shelfRepository.updateShelf(shelf)
```
RasmusAlbrektsen, line 14: I would argue this should be called `const updatedShelf` since it is the returned and updated shelf right?

LindvedKrvang: I agree. It's been updated.

PR #98, opened 2024-02-13
SOF-1785 Minor cleanup for Replay Actions

LindvedKrvang:
Replace '/' with '_' for Replay Action ids. 
Add 'VO' to voice over Replay Action names


k-primeholding-dim (approved):

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -69,12 +69,12 @@ export class Tv2ReplayActionFactory {
     }
   }
 
-  private removeAllWhitespace(value: string): string {
-    return value.replaceAll(' ', '')
+  private sanitizeString(value: string): string {
+    return value.replaceAll(' ', '').replaceAll('/', '_')
   }
```
KvelaGorrrrnio, line 74: From `sanitizeString` I don't know what to expect. A suggestion for an alternative name is `sanitizeIdentifier` or `sanitizeSourceName`.

LindvedKrvang: I didn't chose your naming suggestions because there is nothing in the body of the function that indicates that it has anything to do with either an `Identifier` or a `SourceName`. So including those in the name would be naming the method after _how we intend to use it_ (which is outside the scope of the method) and not _what does the method do_.

I agree, the name could be more specific. I will try and come up with something.

LindvedKrvang: I have made an update :)

LindvedKrvang: I have changed the parameter from `string` to `Tv2SourceMappingWithSound`.
This allows us to call it `getSanitizedName()`.

src/blueprints/tv2/action-factories/tv2-replay-action-factory.ts
```diff
@@ -69,18 +69,18 @@ export class Tv2ReplayActionFactory {
     }
   }
 
-  private removeAllWhitespace(value: string): string {
-    return value.replaceAll(' ', '')
+  private getSanitizedName(source: Tv2SourceMappingWithSound): string {
+    return source.name.replaceAll(' ', '').replaceAll('/', '_')
   }
```
vivanov90, line 75: I don't know how useful this logic is, but in order not to repeat it somewhere, we can export what to replace and with what in the parameters of the function.

```
 private stringReplace(value: string, valueForChange: string, changeTo: string): string {
    return value.replaceAll(valueForChange, changeTo)
  }
```
`valueForChange: string, changeTo: string` can be string[] if we need to cover change multiple parts of the string.

SIDE NODE: that way the function witch call `stringReplace` will know what exactly what will be change it, without following the function.

vivanov90 (approved):

KvelaGorrrrnio (approved):

PR #101, opened 2024-02-22
Sof 1559/notify about device status

LindvedKrvang:
SofieServer now listen for Device statuses and emits a StatusMessage if a Device status has changed.

SofieServer also exposes an endpoint to fetch all current StatusMessages in the system.

Note: A Device in a good state does not save a StatusMessage. SofieServer will stil emit an event when a Device goes back to the good status.

### DIFF
--- a/src/business-logic/services/device-changed-service.ts
+++ b/src/business-logic/services/device-changed-service.ts
@@ -1,0 +1,102 @@
+import { DatabaseChangeService } from './interfaces/database-change-service'
+import { StatusMessageEventEmitter } from './interfaces/status-message-event-emitter'
+import { DataChangedListener } from '../../data-access/repositories/interfaces/data-changed-listener'
+import { Device } from '../../model/entities/device'
+import { StatusMessage } from '../../model/entities/status-message'
+import { StatusMessageRepository } from '../../data-access/repositories/interfaces/status-message-repository'
+import { NotFoundException } from '../../model/exceptions/not-found-exception'
+import { StatusCode } from '../../model/enums/status-code'
+
+export class DeviceChangedService implements DatabaseChangeService {
+  private static instance: DatabaseChangeService
+
+  public static getInstance(
+    statusMessageEventEmitter: StatusMessageEventEmitter,
+    statusMessageRepository: StatusMessageRepository,
+    deviceChangedListener: DataChangedListener<Device>
+  ): DatabaseChangeService {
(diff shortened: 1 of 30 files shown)


src/data-access/repositories/mongo/mongo-device-changed-listener.ts
```diff
@@ -0,0 +1,93 @@
...
+        const updateChange: ChangeStreamUpdateDocument<MongoDevice> = change as ChangeStreamUpdateDocument<MongoDevice>
+        const mongoDevice: MongoDevice | undefined = updateChange.fullDocument
+        if (!mongoDevice) {
+          return
+        }
+        this.onUpdatedCallback(this.convertMongoDeviceToDevice(mongoDevice))
+        return
+      }
+    }
+  }
```
vivanov90, line 44: Here I get it we listen for the changes and react it to change, but if I don't see the model from the names on to the methods I can't get it what we expect to change and what is react to that change. Suggestions for names: `listenForStreamDocumentChanges `and `onStreamDocumentChange`

vivanov90 (commented):
I went through all the files and everything looks fine to me. As well as I pretty much understood how some of the back-end logic works, but I won't approve this PR because my approval doesn't bring much value regarding back-end logic.

k-primeholding-dim (approved):

src/business-logic/services/device-changed-service.ts
```diff
@@ -0,0 +1,102 @@
...
+    deviceChangedListener: DataChangedListener<Device>
+  ) {
+    this.listenForStatusMessageChanges(deviceChangedListener)
+  }
+
+  private listenForStatusMessageChanges(deviceChangedListener: DataChangedListener<Device>): void {
+    deviceChangedListener.onUpdated(device => void this.onDeviceUpdated(device))
+  }
+
+  private async onDeviceUpdated(device: Device): Promise<void> {
```
RasmusAlbrektsen, line 40: At a glance, this method seems rather large and full of nesting logic. But i tried tampering a bit myself and i see why it is this way. So i don't have any suggestions to improve either :-D So i'm only commenting to make a note in case you had any ideas yourself, otherwise it doesn't change my overall opinion of the PR.

RasmusAlbrektsen (approved):
Looks good!

k-primeholding-dim (approved):

PR #102, opened 2024-02-28
SOF-1813 Update Device statusses on startup

LindvedKrvang:
SofieServer now updates the SystemMessage from the devices on startup. 
Devices not connected now also get a system message


vivanov90 (commented):
I check it and don't have any comments, but will not approved, because is back-end related and I don't have the expertise for approve.

RasmusAlbrektsen (approved):

k-primeholding-dim (approved):

PR #104, opened 2024-03-07
SOF-1748 Add a UUID to ActionPanels when saving if it is missing

LindvedKrvang:


k-primeholding-dim (approved):

src/data-access/repositories/mongo/mongo-shelf-repository.ts
```diff
@@ -38,7 +39,18 @@ export class MongoShelfRepository extends BaseMongoRepository implements ShelfCo
...
     return shelfConfiguration
   }
+
+  private applyMissingActionPanelIds(shelfConfiguration: ShelfConfiguration): ShelfConfiguration {
+    shelfConfiguration.actionPanelConfigurations = shelfConfiguration.actionPanelConfigurations.map(actionPanel => {
+      if (!actionPanel.id || actionPanel.id.length <= 0) {
+        actionPanel.id = this.uuidGenerator.generateUuid()
+      }
+      return actionPanel
+    })
```
KvelaGorrrrnio, line 53: Condensing the condition to `!actionPanel.id` should give the same behavior since an empty string is considered a falsey value.

If you want to use a bit of fancy syntax you can condense this down to:
```typescript
    shelfConfiguration.actionPanelConfigurations = shelfConfiguration.actionPanelConfigurations.map(actionPanel => {
      actionPanel.id ||= this.uuidGenerator.generateUuid()
      return actionPanel
    })
```
or via a functional approach:
```typescript
    shelfConfiguration.actionPanelConfigurations = shelfConfiguration.actionPanelConfigurations.map(actionPanel => ({
      ...actionPanel,
      id: actionPanel.id || this.uuidGenerator.generateUuid()
    }))
```
or have the entire method use functional composition, like the below snippet. I have changed `actionPanel` to `actionPanelConfiguration`.
```typescript
    return {
      ..shelfConfiguration,
      actionPanelConfigurations: shelfConfiguration.actionPanelConfigurations.map(
        actionPanelConfiguration => ({
          ...actionPanelConfiguration,
          id: actionPanel.id || this.uuidGenerator.generateUuid()
        })
      )
    }
```

KvelaGorrrrnio (approved):
I assume that this is to have backward compatibility for shelfs created during the development of the feature?

RasmusAlbrektsen (approved):

LindvedKrvang:
> I assume that this is to have backward compatibility for shelfs created during the development of the feature?

Not quite. It's because the frontend needs ids on the ShelfActionPanelConfigurations which is not really a good argument, because we shouldn't make changes to the backend because of a frontend :p 
But as we talked about, it's not a bad idea to have id's on our entities.

src/data-access/repositories/mongo/mongo-shelf-repository.ts
LindvedKrvang: I have changed the conditional to just be `!actionPanel.id`.
I like the simplicity of that approach compared to some of your functional-approach suggestions.

PR #107, opened 2024-03-18
SOF-1821 - Implement locking mechanism for rundown actions for 500 ms after action arrival

vivanov90:


src/business-logic/facades/service-facade.ts
```diff
@@ -39,15 +40,13 @@ export class ServiceFacade {
...
   }
 
   public static createTimelineBuilder(): TimelineBuilder {
     const superflyTimelineBuilder: TimelineBuilder = new SuperflyTimelineBuilder(new JsonObjectCloner())
-    return new BlueprintTimelineBuilder(
-      superflyTimelineBuilder,
-      RepositoryFacade.createConfigurationRepository(),
-      BlueprintsFacade.createBlueprint()
-    )
+    return new BlueprintTimelineBuilder(superflyTimelineBuilder, RepositoryFacade.createConfigurationRepository(), BlueprintsFacade.createBlueprint())
```
LindvedKrvang, line 49: Was this a linter that changed this?
Because this formatting makes the line way too long.

Comment applies for the other places in this file as well.

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,125 @@
+import { Part } from '../../model/entities/part'
+import { Piece } from '../../model/entities/piece'
+import { Owner } from '../../model/enums/owner'
+import { InTransition } from '../../model/value-objects/in-transition'
+import { RundownService } from './interfaces/rundown-service'
+
+const RUNDOWN_LOCK_INTERVAL = 500
+const RUNDOWN_LOCK_ERROR_TEXT = 'Unable to do action. An action was already executed less than 500ms ago'
```
LindvedKrvang, line 8: Would be nice to add types to these :)

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,125 @@
+import { Part } from '../../model/entities/part'
+import { Piece } from '../../model/entities/piece'
+import { Owner } from '../../model/enums/owner'
+import { InTransition } from '../../model/value-objects/in-transition'
+import { RundownService } from './interfaces/rundown-service'
+
+const RUNDOWN_LOCK_INTERVAL = 500
```
LindvedKrvang, line 7: There is no indication whether this is in miliseconds, seconds, minutes, hours etc.
Adding `_MS` would fix that :)

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,125 @@
...
+const RUNDOWN_LOCK_ERROR_TEXT = 'Unable to do action. An action was already executed less than 500ms ago'
+
+export class RundownLockService implements RundownService {
+  private static instance: RundownService
+  private hasEnoughTimePassed: boolean = true
```
LindvedKrvang, line 12: You don't need a boolean for this (nor be using a `timeout` which you call in every function).

You can simply store the timestamp of the last executed action and then, when you call an action, you check if `Date.now()` is greater than `timestamp + lock_delay`.
It will look something like this:

```
RundownLockService ... {
  private lastActionCalledTimestamp: number

  public takeNext(...): ... {
    this.assertEnoughTimeHasPassed()
    this.rundownService.takeNext(...)
  }

  private assertEnoughTimeHasPassed(): void {
    const now: number = Date.now()
    if (now < (this.lastActionCalledTimestamp + RUNDOWN_LOCK_INTERVAL_MS) {
      throw new RundownIsLockedException('...')
    }
    this.lastActionCalledTimestamp = now
  }
}
```

Then all you have to in each method that needs to be locked, is to add the line `this.assertEnoughTimeHasPassed()`

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  constructor(private readonly rundownService: RundownService) {}
+
+  public static getInstance(rundownService: RundownService): RundownService {
+    if (!this.instance) {
+      this.instance = new RundownLockService(rundownService)
+    }
+    return this.instance
+  }
```
LindvedKrvang, line 21: The order variables and static methods is not ordered after our conventions.
For singletons we order it in this order:
- private instance
- public getInstance
- instance variables
- constructor

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,125 @@
...
+    setTimeout(() => (this.hasEnoughTimePassed = true), RUNDOWN_LOCK_INTERVAL)
+    return this.rundownService.takeNext(rundownId)
+  }
+
+  public deleteRundown(rundownId: string): Promise<void> {
+    if (!this.hasEnoughTimePassed) {
+      throw new Error(RUNDOWN_LOCK_ERROR_TEXT)
+    }
+    this.hasEnoughTimePassed = false
+    setTimeout(() => (this.hasEnoughTimePassed = true), RUNDOWN_LOCK_INTERVAL)
```
LindvedKrvang, line 37: You are duplicating these lines in all methods that you want a lock for.
Whenever you start duplicating yourself, there should be something inside you telling you, that you are doing something suboptimally.

When you find yourself duplicating the same thing, that should be a clear indicator, that you atleast could extract it to a method.
```
private assertEnoughTimeHasPassed(): void {
  if (!this.hasEnoughTimePassed) {
    throw new Error(RUNDOWN_LOCK_ERROR_TEXT)
  }
  this.hasEnoughTimePassed = false
  setTimeout(() => (this.hasEnoughTimePassed = true), RUNDOWN_LOCK_INTERVAL)
}
```
The above snippet does the exact same thing as you implemented, but now you just need to call the method instead of rewriting it, everytime you need it.

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,125 @@
...
+  public static getInstance(rundownService: RundownService): RundownService {
+    if (!this.instance) {
+      this.instance = new RundownLockService(rundownService)
+    }
+    return this.instance
+  }
+
+  public takeNext(rundownId: string): Promise<void> {
+    if (!this.hasEnoughTimePassed) {
+      throw new Error(RUNDOWN_LOCK_ERROR_TEXT)
```
LindvedKrvang, line 25: We try to avoid to throw `Error`, but instead have more specific errors like `UnsupportedOperationException` or `AlreadyActivatedException`.
Take a look in `/model/exceptions` to see if you can find an exception that fits your need. If not, feel free to create a new one (hint: We need a new for this :) ).

This also gives us the oppertunity to provide an `ErrorCode` to our exception. I can see we already have `ErrorCode.RUNDOWN_IS_LOCKED`, but it has yet to be used for an exception.

Alternatively, you can also just write:
```
throw new Exception(ErrorCode.RUNDOWN_IS_LOCKED, {yourErrorMessage})
```

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,125 @@
...
+    this.hasEnoughTimePassed = false
+    setTimeout(() => (this.hasEnoughTimePassed = true), RUNDOWN_LOCK_INTERVAL)
+    return this.rundownService.enterRehearsal(rundownId)
+  }
+
+  public deactivateRundown(rundownId: string): Promise<void> {
+    return this.rundownService.deactivateRundown(rundownId)
+  }
+
+  public setNext(rundownId: string, segmentId: string, partId: string, owner?: Owner): Promise<void> {
```
LindvedKrvang, line 63: `SetNext` is not supposed to be behind the lock

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,125 @@
...
+  public resetRundown(rundownId: string): Promise<void> {
+    if (!this.hasEnoughTimePassed) {
+      throw new Error(RUNDOWN_LOCK_ERROR_TEXT)
+    }
+    this.hasEnoughTimePassed = false
+    setTimeout(() => (this.hasEnoughTimePassed = true), RUNDOWN_LOCK_INTERVAL)
+    return this.rundownService.resetRundown(rundownId)
+  }
+
+  public insertPartAsOnAir(rundownId: string, part: Part): Promise<void> {
```
LindvedKrvang, line 81: `insertPartAsOnAir`, `insertPartAsNext`, `insertPieceAsNext`, `insertPieceAsOnAir` and `replacePieceOnAirOnNextPart` are not supposed to be behind the lock.
Read the acceptance criteria for the things that should be behind the lock.

LindvedKrvang (commented):
Tests are missing

src/business-logic/services/rundown-lock-service.ts
vivanov90: Fix

src/business-logic/services/rundown-lock-service.ts
vivanov90: Fix

src/business-logic/services/rundown-lock-service.ts
vivanov90: I create new `LockRundownException`

src/business-logic/services/rundown-lock-service.ts
vivanov90: That logic is rework in previous comment

src/business-logic/services/rundown-lock-service.ts
vivanov90: I rework the logic to be that way

src/business-logic/services/rundown-lock-service.ts
vivanov90: Done

src/business-logic/services/rundown-lock-service.ts
vivanov90: Done

src/business-logic/facades/service-facade.ts
vivanov90: Return in to previous formation

vivanov90:
@LindvedKrvang I add some tests

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,85 @@
+import { Part } from '../../model/entities/part'
+import { Piece } from '../../model/entities/piece'
+import { Owner } from '../../model/enums/owner'
+import { LockRundownException } from '../../model/exceptions/lock-rundown-exception'
+import { InTransition } from '../../model/value-objects/in-transition'
+import { RundownService } from './interfaces/rundown-service'
+
+const RUNDOWN_LOCK_INTERVAL_MS: number = 500
+const RUNDOWN_LOCK_ERROR_TEXT: string = 'Unable to do action. An action was already executed less than 500ms ago'
```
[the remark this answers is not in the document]
k-primeholding-dim: Should it be more fuzzy - like 'half a second' or even 'less than a second' ago instead of engineering precision in three places after the decimal millis?

src/business-logic/services/test/rundown-lock-service.spec.ts
```diff
@@ -0,0 +1,118 @@
+import { RundownLockService } from './../rundown-lock-service'
+import { instance, mock, verify, when } from '@typestrong/ts-mockito'
+import { Rundown, RundownInterface } from '../../../model/entities/rundown'
+import { LockRundownException } from '../../../model/exceptions/lock-rundown-exception'
+
+describe(RundownLockService.name, () => {
+  describe('Rundown lock throw error cases', () => {
```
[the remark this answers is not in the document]
vivanov90: I reworked them, hopefully now is better

src/business-logic/services/test/rundown-lock-service.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      expect(result).toThrow(LockRundownException)
+    })
+
+    it('Throw error if we call 2 different actions with 499 ms time between', async () => {
+      const mockRepo: RundownLockService = mock<RundownLockService>()
+      const randomRundownId: string = 'randomRundownId'
+      const randomRundown: Rundown = new Rundown({ id: randomRundownId } as RundownInterface)
+      const now: number = Date.now()
+
+      jest.useFakeTimers().setSystemTime(now)
```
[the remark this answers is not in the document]
vivanov90: Fixed

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,85 @@
...
+  private static instance: RundownService
+
+  public static getInstance(rundownService: RundownService): RundownService {
+    if (!this.instance) {
+      this.instance = new RundownLockService(rundownService)
+    }
+    return this.instance
+  }
+
+  private lastActionCalledTimestamp: number
```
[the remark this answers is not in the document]
vivanov90: Renamed to `lastOperationTakenEpochTimestamp`

src/business-logic/services/rundown-lock-service.ts
```diff
@@ -0,0 +1,85 @@
...
+import { Piece } from '../../model/entities/piece'
+import { Owner } from '../../model/enums/owner'
+import { LockRundownException } from '../../model/exceptions/lock-rundown-exception'
+import { InTransition } from '../../model/value-objects/in-transition'
+import { RundownService } from './interfaces/rundown-service'
+
+const RUNDOWN_LOCK_INTERVAL_MS: number = 500
+const RUNDOWN_LOCK_ERROR_TEXT: string = 'Unable to do action. An action was already executed less than 500ms ago'
+
+export class RundownLockService implements RundownService {
```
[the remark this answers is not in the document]
vivanov90: Renamed

src/business-logic/services/rundown-lock-service.ts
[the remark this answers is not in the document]
vivanov90: @KvelaGorrrrnio I fix your comment, @k-primeholding-dim we can do it but I don't see much benefits of that. My idea for this error is to tell the user, please don't spam. If we want we can calculate seconds from this MS and the message will looks like:
`Unable to do action. An action was already executed less than 0.5 seconds ago`

src/business-logic/services/rundown-lock-service.ts
RasmusAlbrektsen, line 1: The name of this file no longer fits the class name change :-)

src/business-logic/services/test/rundown-lock-service.spec.ts
RasmusAlbrektsen, line 1: The name of this fille no longer fits due to my previous comment.

src/model/exceptions/lock-rundown-exception.ts
RasmusAlbrektsen, line 1: This exception still refers to the `lock`-naming.

src/business-logic/services/rundown-lock-service.ts
vivanov90: Fixed

src/business-logic/services/test/rundown-lock-service.spec.ts
vivanov90: Fixed

src/model/exceptions/lock-rundown-exception.ts
vivanov90: Fixed

src/model/exceptions/throttled-rundown-exception.ts
```diff
@@ -0,0 +1,8 @@
+import { ErrorCode } from '../enums/error-code'
+import { Exception } from './exception'
+
+export class ThrottledRundownException extends Exception {
+  constructor(message: string) {
+    super(ErrorCode.RUNDOWN_IS_LOCKED, message)
```
RasmusAlbrektsen, line 6: The error is still `RUNDOWN_IS_LOCKED` though i see it wasn't created by you in this PR. 
From what i can see this is the only place it is used so maybe that should be changed to fit the naming as well :-)

RasmusAlbrektsen (approved):

src/model/exceptions/throttled-rundown-exception.ts
vivanov90: That is not create in this PR. I can change it, but can we have confirmation to that? @LindvedKrvang @KvelaGorrrrnio
I can rename it to:  `RUNDOWN_IS_THROTTLET`

KvelaGorrrrnio: Confirming that we can change it to `RUNDOWN_IS_THROTTLED` :)

src/business-logic/services/test/throttled-rundown-service.spec.ts
```diff
@@ -0,0 +1,124 @@
+import { ThrottledRundownService } from '../throttled-rundown-service'
+import { instance, mock, verify, when } from '@typestrong/ts-mockito'
+import { Rundown, RundownInterface } from '../../../model/entities/rundown'
+import { ThrottledRundownException } from '../../../model/exceptions/throttled-rundown-exception'
+
+describe(ThrottledRundownService.name, () => {
+  describe('when two operations are performed within the locked interval', () => {
+    beforeEach(() => jest.useFakeTimers())
+    afterEach(() => jest.useRealTimers())
```
KvelaGorrrrnio, line 9: Since these are used for all tests, also outside of this describe, you can move this to the outermost describe and remove the duplication :)

src/business-logic/services/test/throttled-rundown-service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    describe('throws an error to second operation', () => {
+      it('throws error when two take nexts are performed with 0ms delay', async () => {
+        const mockRepo: ThrottledRundownService = mock<ThrottledRundownService>()
```
KvelaGorrrrnio, line 12: You are mocking the testee type here. It would be nice with a more descriptive name and another type. I assume that this is the rundown service that should be passed in to the throttled rundown service. If that is the case, then I will suggest:
```typescript
const mockedRundownService: RundownService = mock<RundownService>()
```

And since you have the same mock setup for this in all tests, I would extract the mock setup out into a separate function.

KvelaGorrrrnio (commented):
Looks good :rocket:

k-primeholding-dim (approved):

src/business-logic/services/test/throttled-rundown-service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        verify(mockRepo.takeNext(randomRundownId)).once()
+
+        const result: () => Promise<void> = () => testee.takeNext(randomRundown.id)
+
+        expect(result).toThrow(ThrottledRundownException)
+      })
+
+      it('throws error when two different actions are performed with 0ms delay', async () => {
+        const mockRepo: ThrottledRundownService = mock<ThrottledRundownService>()
```
LindvedKrvang, line 30: What does `mockRepo` mean? `repo` is useally short for `repository` which is what connects and talks to databases, but you are mocking `ThrottledRundownService` which is not a repository.

Also, the file you want to test is `ThrottledRundownService`. You should not be mocking the file under test.

I see why you have this mock (so you can give it to your testee), but your testee expects a `RundownService` so you can just mock the `RundownService` instead of the `ThrottledRundownSerivece`:
```
const mockRundownService: RundownService = mock<RundownService>()
```

This comment goes for all your tests.

LindvedKrvang (approved):
Looks good :)

Only a few things with some naming in the tests

src/business-logic/services/test/throttled-rundown-service.spec.ts
vivanov90: Fixed and add some more test cover logic out side of throttled mechanism

src/business-logic/services/test/throttled-rundown-service.spec.ts
vivanov90: Fixed

src/business-logic/services/test/throttled-rundown-service.spec.ts
vivanov90: First is fixed, the second I simplify the test and in reality now we don't need any external mock setup because we make the setup very small, when we remove unnecessary values.

src/business-logic/services/test/throttled-rundown-service.spec.ts
```diff
@@ -0,0 +1,125 @@
+import { ThrottledRundownService } from '../throttled-rundown-service'
+import { anything, instance, mock, verify } from '@typestrong/ts-mockito'
+import { ThrottledRundownException } from '../../../model/exceptions/throttled-rundown-exception'
+import { RundownService } from '../interfaces/rundown-service'
+
+describe(ThrottledRundownService.name, () => {
+  beforeEach(() => jest.useFakeTimers())
+  afterEach(() => jest.useRealTimers())
+  describe('when two operations are performed within the throttled interval', () => {
```
KvelaGorrrrnio, line 9: It would be nice with a space between the `afterEach` and the `describe`.

src/business-logic/services/test/throttled-rundown-service.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    })
+  })
+
+  describe('when performed operations witch should not have throttled mechanism', () => {
+    it('performed two insert part as next operations without any time between', async () => {
+      const mockRundownService: RundownService = mock<RundownService>()
+
+      const testee: ThrottledRundownService = new ThrottledRundownService(instance(mockRundownService))
+
+      await testee.insertPartAsNext(anything(), anything())
```
KvelaGorrrrnio, line 95: `anything()` is a matcher object for mocking, not a value that should be used on real instances. Instead, we should use correctly typed values. You can use the `TestEntityFactory` if you need to create entity data for the tests.

vivanov90: All anything() was removed.

src/business-logic/services/test/throttled-rundown-service.spec.ts
vivanov90: Added

KvelaGorrrrnio (approved):
Nice work! 💪

PR #108, opened 2024-03-19
SOF 1828 Resetting in Rehearsal Should Return Rundown to Rehearsal

RasmusAlbrektsen:
This PR fixes an issue where we didn't take into account that we could reset the rundown from rehearsal mode. I therefore made the reset method remember which mode the rundown had prior to deactivation.


vivanov90 (approved):

k-primeholding-dim (approved):

src/model/entities/rundown.ts
```diff
@@ -676,7 +676,12 @@ export class Rundown extends BasicRundown {
   }
 
   public reset(): void {
+    const rundownModePriorToDeactivation: RundownMode = this.mode
     this.deactivate()
+    if (rundownModePriorToDeactivation === RundownMode.REHEARSAL) {
+      this.enterRehearsal()
+      return
+    }
     this.activate()
```
LindvedKrvang, line 685: You can now use void functions in ternaries, so you could write it like this:
```
const isRundownInRehearsalBeforeResetting: boolean = this.mode === RundownMode.REHEARSAL
this.deactivate()
isRundownInRehearsalBeforeResetting ? this.enterRehearsal() : this.activate()
```

LindvedKrvang (approved):

src/model/entities/rundown.ts
KvelaGorrrrnio: Using void in ternaries should not be something new?

KvelaGorrrrnio (approved):

PR #117, opened 2024-03-29
SOF-1596 - Be able to run mongo independent of meteor

k-primeholding-dim:
With this PS a missing cli tool is added to the project - own implementation of several Docker API interactions, aiming to provide portability when running MongoDB.
Added:
db/manager/mongo.ts - cli app with self-descriptive options
```bash
~/TV2/sofie-server SOF-1596/mongo-off-meteor ❯ npx ts-node db/manager/mongo.ts --help                                                       16:01:37
Usage: mongo-manager [option]
Options:
  -h, --help    Display this help message
  --start       Start the mongo service
  --stop        Stop the mongo service
  --restart     Restart the mongo service
  --initrs      Initialize the replica set
  --seed        Seed the database with initial data
  --drop        Stop then drop the database container and its volumes
  --spy         Spy on database events
  --dump        Dump all from mongodb://gateway.docker.internal:3001/ to ./db/dumps/meteor
~/TV2/sofie-server SOF-1596/mongo-off-meteor ❯
```
The above cli tool is later used in package scripts section:
```json
    "start-database": "ts-node ./database/manager/mongo.ts --start",
    "stop-database": "ts-node ./database/manager/mongo.ts --stop",
    "restart-database": "ts-node ./database/manager/mongo.ts --restart",
    "drop-database": "ts-node ./database/manager/mongo.ts --drop",
    "init-replica-set": "ts-node ./database/manager/mongo.ts --initrs",
    "seed-database": "ts-node ./database/manager/mongo.ts --seed",
    "spy-database": "ts-node ./database/manager/mongo.ts --spy",
    "dump-database": "ts-node ./database/manager/mongo.ts --dump"
```

A guided tour on usage is composed in `db/README.md`


package.json
```diff
@@ -13,7 +13,15 @@
...
-    "reset": "rimraf --glob dist node_modules coverage *.log"
+    "reset": "rimraf --glob dist node_modules coverage *.log",
+    "start-database": "npx ts-node ./db/manager/mongo.ts --start",
+    "stop-database": "npx ts-node ./db/manager/mongo.ts --stop",
+    "restart-database": "npx ts-node ./db/manager/mongo.ts --restart",
+    "drop-database": "npx ts-node ./db/manager/mongo.ts --drop",
+    "init-replica-set": "npx ts-node ./db/manager/mongo.ts --initrs",
+    "seed-database": "npx ts-node ./db/manager/mongo.ts --seed",
+    "spy-database": "npx ts-node ./db/manager/mongo.ts --spy",
+    "dump-database": "npx ts-node ./db/manager/mongo.ts --dump"
```
KvelaGorrrrnio: `npx ` should be removed :)

db/manager/mongo.ts
```diff
@@ -0,0 +1,255 @@
...
+  seed: string | boolean | undefined
+  drop: string | boolean | undefined
+  spy: string | boolean | undefined
+  dump: string | boolean | undefined
+}): void {
+  switch (true) {
+    case values.start && !values.stop && !values.restart && !values.initrs && !values.seed && !values.drop && !values.spy && !values.dump:
+      console.log('starting mongo...')
+      startMongoContainer()
+      break
```
KvelaGorrrrnio: Instead of checking if the flag is set and all other flags are not set, this can be divided into 2 phases. It will make the code more manageable.

```typescript
// Phase 1: Collect set arguments
const arguments: string[] = Object.entries(values)
  .filter(([, isSet]) => isSet)
  .map(([argument]) => argument)
  
if (arguments.length !== 1) {
  console.log('Please provide a valid option. Try --help for more information.')
  if (arguments > 0) {
    console.log('Unsupported: ', arguments)
  }
  return
}

// Phase 2: Use a switch matcher on first argument name
switch (arguments[0]) {
  case 'start':
    console.log('starting mongo...')
    startMongoContainer()
    break
  // ...
}
```
Disclaimer: I haven't tested the code.

db/manager/mongo.ts
```diff
@@ -0,0 +1,255 @@
...
+    '"rs.initiate({\'_id\':\'rs0\',members:[{\'_id\':0,\'host\':\'127.0.0.1:3001\'}]})"'
+  ])
+}
+
+function startMongoContainer(): void {
+  if (/^win/i.test(process.platform)) {
+    docker([
+      'run', '--rm',
+      '--name', 'sofie-mongodb',
+      '-v', '.\\db\\dumps:/dumps',
```
KvelaGorrrrnio: Is it necessary to use backslash on windows systems?

db/manager/mongo.ts
```diff
@@ -0,0 +1,255 @@
...
+      '-v', 'sofie-mongodb-data:/data/db',
+      '-v', 'sofie-mongodb-config:/data/configdb',
+      '-p', '3001:3001',
+      // requires mongotools installed on Windows
+      // '--health-cmd', '"test', '\\"echo', '\'db.stats().ok\'', '|', 'mongosh', '\\"mongodb://127.0.0.1:3001/sofie?replicaSet=rs0\\"', '--quiet\\""',
+      '--health-interval=1s',
+      '-d', 'mongo:6.0.1',
```
KvelaGorrrrnio: It would be nice with a newline between these two arguments.

db/manager/mongo.ts
```diff
@@ -0,0 +1,255 @@
...
+    docker([
+      'run', '--rm',
+      '--name', 'sofie-mongodb',
+      '-v', './db/dumps:/dumps',
+      '-v', 'sofie-mongodb-data:/data/db',
+      '-v', 'sofie-mongodb-config:/data/configdb',
+      '-p', '3001:3001',
+      '--health-cmd', '"test', '\\"echo', '\'db.stats().ok\'', '|', 'mongosh', '\\"mongodb://127.0.0.1:3001/sofie?replicaSet=rs0\\"', '--quiet\\""',
+      '--health-interval=1s',
+      '-d', 'mongo:6.0.1',
```
KvelaGorrrrnio: It would be nice with a newline between these two arguments.

db/README.md
```diff
@@ -0,0 +1,148 @@
...
+- `docker` installed
+- `ts-node` installed ( `yarn global add ts-node typescript` )
+- mongodb-tools for running cli commands (optional)
+
+Steps
+===
+
+*note*: short example at the bottom of the page
+
+1.Obtain dump from tv-automation-server-core
```
KvelaGorrrrnio: Missing space after `1.`.

db/README.md
KvelaGorrrrnio, line 1: It would be nice if the folder is called database.

.gitignore
```diff
@@ -10,4 +10,8 @@ dist/
 .python-version
 
 # Ignore VSCode specifics
-.vscode
\ No newline at end of file
+.vscode
+
+jest.config.js*
```
KvelaGorrrrnio: Why is this added?

db/README.md
```diff
@@ -0,0 +1,148 @@
...
+```bash
+cd sofie-server/db/dumps
+mongodump "mongodb://localhost:3001" --oplog --out=./meteor
+```
+
+this will create a dump of the database in the `meteor` sub-folder (git excluded!). In the next steps we'll use this to restore the database in our containerized MongoDB. Once restored - you can safely remove the `meteor` sub-folder.
+
+2. Stop `tv-automation-server-core` - we don't need it anymore.
+
+3. Open third terminal: :
```
KvelaGorrrrnio: The second `:` can be removed.

db/manager/mongo.ts
k-primeholding-dim: Technically executing the command in Powershell does not require double-backslashes, however I've added them for linter to be happy, since I'm not writing `\d` (escaped d) character sequence.

LindvedKrvang (commented):
What is the reason for creating an entire CLI tool?
What we need is:
- Able to start a MongoDB outside of Sofie Core
- Able to stop that MongoDB
- Have Core use that MongoDB instead of the one in Meteor

db/manager/mongo.ts
KvelaGorrrrnio: To me, it looks like the only difference between this and the non-Windows version is the backslashes. So if the forwards slashes works, then we won't have to check the platform.

k-primeholding-dim: Forward slash does not trigger error, but also does not attach the needed volume.
image

package.json
```diff
@@ -13,7 +13,15 @@
     "lint-fix": "eslint --fix --ext .ts .",
     "validate": "yarn audit --level moderate --groups dependencies --recursive",
     "clean": "rimraf dist",
-    "reset": "rimraf --glob dist node_modules coverage *.log"
+    "reset": "rimraf --glob dist node_modules coverage *.log",
+    "start-database": "ts-node-script ./database/manager/mongo.ts --start",
```
KvelaGorrrrnio: Why is it changed from `ts-node` to `ts-node-script`?

k-primeholding-dim: Should not be of a difference see https://github.com/TypeStrong/ts-node/blob/main/package.json#L41 , https://github.com/TypeStrong/ts-node/blob/master/src/bin-script.ts aiming portability since it locates a tsconfig relative to the location of the script, not to cwd.

KvelaGorrrrnio: If there is no difference, then I prefer just having `ts-node`. I have been unable to find documentation on the `--script-mode flag`.

.gitignore
```diff
@@ -11,3 +11,7 @@ dist/
 
 # Ignore VSCode specifics
 .vscode
+
+# Ignore database manager js*
+database/dumps/meteor
+database/manager/*.js*
```
KvelaGorrrrnio: One could consider to move this to a new `.gitignore` file under the database directory. That way, if we at some point change the script to another solution, e.g. docker compose setup. Then we have a logic grouping of all the necessary logic and config for the database  manager.
I would also say that it is the entire dumps folder that should be ignored :)

Do we generate js files for the database manager script?

database/README.md
```diff
@@ -0,0 +1,148 @@
...
+export MONGO_URL='mongodb://127.0.0.1:3001/meteor?replicaSet=rs0'
+export MONGO_OPLOG_URL='mongodb://127.0.0.1:3001/local?replicaSet=rs0'
+yarn dev
+```
+
+For Windows users, the `MONGO_URL` and `MONGO_OPLOG_URL` can be set by running the following commands in the terminal:
+
+```bash
+set MONGO_URL='mongodb://127.0.0.1:3001/meteor?replicaSet=rs0'
+set MONGO_OPLOG_URL='mongodb://127.0.0.1:3001/local?replicaSet=rs0'
```
KvelaGorrrrnio, line 136: The replica set should not be necessary to set :)

.gitignore
k-primeholding-dim: By ignoring a specific folder in dumps, I can deliver the folder (as structure in fs tree) upon checking out the code, and since is specifically in use for the purpose of data migration between the Mongo servers I treat it differently.

There are by-products of the ts script invocation (*.js, *.js.map).

KvelaGorrrrnio: Is there an issue with the dumps folder not being in version control? As I see it, the dumps folder will be generated when I extract the data from the meteor database instance, and should only be created when that is done.

> There are by-products of the ts script invocation (*.js, *.js.map).

But we don't have a workflow, as far as I know, where these are generated.

k-primeholding-dim: Will drop a tsconfig.json in folder to not emit nor create maps. Also dumop destination folder is automagically created, thus gitignoring it too.

KvelaGorrrrnio (approved):

LindvedKrvang (approved):

PR #120, opened 2024-04-09
SOF-1878 Blueprint configuration now maps VideoMixerType

LindvedKrvang:
Blueprint configuration now maps VideoMixerType.

Blueprints now supports creating TimelineObjects for TriCaster.
Note: The TriCasterTypes has been "refactored" to a hopefully more understandable format. However, the attributes of the types still _must_ match what's in TSR. If not, TSR has no idea of what to do with it. So there may be some naming that might seem weird, but to change it we would also need to change it in TSR (which we prefer not to do at the moment).

PR #122, opened 2024-04-17
Sof 1883/switch video mixer timeline object factory

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-server/pull/120
Update: All dependent PRs has been merged.

When we change the VideoMixerType in the configuration we now update the Action- and TimelineObjectFactories so we are creating the correct TimelineObjects.

### DIFF
--- a/src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
+++ b/src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
@@ -91,16 +91,14 @@ export class Tv2ActionFactoryProvider {
     shouldFactoryBeRecreatedCallback: (configuration: Tv2BlueprintConfiguration) => boolean,
     configuration: Tv2BlueprintConfiguration
   ): ActionFactoryInstance<T> {
-    if (!!actionFactoryInstance && !actionFactoryInstance.shouldFactoryBeRecreated(configuration)) {
+    if (actionFactoryInstance && !actionFactoryInstance.shouldFactoryBeRecreated(configuration)) {
       return actionFactoryInstance
     }
 
-    actionFactoryInstance = {
+    return {
       factory: createFactoryCallback(),
       shouldFactoryBeRecreated: shouldFactoryBeRecreatedCallback
     }
-
-    return actionFactoryInstance
   }
 
   private didVideoMixerTypeChange(oldConfiguration: Tv2BlueprintConfiguration, newConfiguration: Tv2BlueprintConfiguration): boolean {
--- a/src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
+++ b/src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
@@ -1,0 +1,180 @@
+import { Tv2BlueprintConfiguration } from '../value-objects/tv2-blueprint-configuration'
+import { Tv2CameraActionFactory } from './tv2-camera-action-factory'
+import { Tv2RemoteActionFactory } from './tv2-remote-action-factory'
+import { Tv2TransitionEffectActionFactory } from './tv2-transition-effect-action-factory'
+import { Tv2AudioActionFactory } from './tv2-audio-action-factory'
+import { Tv2GraphicsActionFactory } from './tv2-graphics-action-factory'
+import { Tv2VideoClipActionFactory } from './tv2-video-clip-action-factory'
+import { Tv2VideoMixerConfigurationActionFactory } from './tv2-video-mixer-configuration-action-factory'
+import { Tv2SplitScreenActionFactory } from './tv2-split-screen-action-factory'
+import { Tv2ReplayActionFactory } from './tv2-replay-action-factory'
+import { Tv2RobotActionFactory } from './tv2-robot-action-factory'
+import {
+  Tv2AtemVideoMixerTimelineObjectFactory
+} from '../timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory'
+import {
+  Tv2VideoMixerTimelineObjectFactory
+} from '../timeline-object-factories/interfaces/tv2-video-mixer-timeline-object-factory'
+import {
+  Tv2AudioTimelineObjectFactory
+} from '../timeline-object-factories/interfaces/tv2-audio-timeline-object-factory'
+import { DeviceType } from '../../../model/enums/device-type'
+import { Tv2LoggerFacade } from '../tv2-logger-facade'
+import {
+  Tv2TriCasterVideoMixerTimelineObjectFactory
+} from '../timeline-object-factories/tv2-tri-caster-video-mixer-timeline-object-factory'
+import { AtemToTriCasterSplitScreenConverter } from '../helpers/atem-to-tricaster-split-screen-converter'
+import { Tv2MisconfigurationException } from '../exceptions/tv2-misconfiguration-exception'
+import {
+  Tv2SisyfosAudioTimelineObjectFactory
+} from '../timeline-object-factories/tv2-sisyfos-audio-timeline-object-factory'
+import {
+  Tv2VideoClipTimelineObjectFactory
+} from '../timeline-object-factories/interfaces/tv2-video-clip-timeline-object-factory'
+import { Tv2AssetPathHelper } from '../helpers/tv2-asset-path-helper'
+import { Tv2CasparCgTimelineObjectFactory } from '../timeline-object-factories/tv2-caspar-cg-timeline-object-factory'
+import { Tv2ActionManifestMapper } from '../helpers/tv2-action-manifest-mapper'
+import {
+  Tv2GraphicsTimelineObjectFactoryFactory
+} from '../timeline-object-factories/tv2-graphics-timeline-object-factory-factory'
+import { Tv2StringHashConverter } from '../helpers/tv2-string-hash-converter'
+import {
+  Tv2GraphicsSplitScreenTimelineObjectFactory
+} from '../timeline-object-factories/interfaces/tv2-graphics-split-screen-timeline-object-factory'
+import {
+  Tv2RobotTimelineObjectFactory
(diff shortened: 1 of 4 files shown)


src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
```diff
@@ -0,0 +1,328 @@
...
+  public createCameraActionFactory(configuration: Tv2BlueprintConfiguration): Tv2CameraActionFactory {
+    this.cameraActionFactoryInstance = this.getUpdatedActionFactoryInstance(
+      this.cameraActionFactoryInstance,
+      () => {
+        return new Tv2CameraActionFactory(
+          this.createVideoMixerTimelineObjectFactory(configuration),
+          this.createAudioTimelineObjectFactory()
+        )
+      },
+      (c: Tv2BlueprintConfiguration): boolean => {
```
RasmusAlbrektsen, line 79: Is this the case where we would be alright with just using `c` since we refer to a configuration in the context where we already have a configuration?

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    actionFactoryInstance = {
+      factory: createFactoryCallback(),
+      shouldFactoryBeRecreated: shouldFactoryBeRecreatedCallback
+    }
+
+    return actionFactoryInstance
+  }
+
+  private didVideoMixerTypeChange(oldConfiguration: Tv2BlueprintConfiguration, newConfiguration: Tv2BlueprintConfiguration): boolean {
```
RasmusAlbrektsen, line 104: Don't we usually do it from a past perspective and say `hasVideoMixerTypeChanged` in these cases? In any case i actually don't dislike this, we usually just don't see it, i just want to make sure.

RasmusAlbrektsen (dismissed):
Very code! I must say! 
All i had was a couple of questions really :-)

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
LindvedKrvang: `did` is also past tense, but you are right, we normally have been using `has`.
I think both works, so if there is a desire to change it to `has`, I don't mind :)

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
LindvedKrvang: `c` isn't ideal. It's not uncommon though to name variables in arrow functions like these and as you point out, we are doing it here since we already have a variable called `configuration` within the scope.

The only other names I can come up with is to name `c` `oldConfiguration` and `configuration` `newConfiguration` which I am also not that much of a fan of...

RasmusAlbrektsen (approved):

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
```diff
@@ -0,0 +1,328 @@
...
+  private cameraActionFactoryInstance: ActionFactoryInstance<Tv2CameraActionFactory>
+  private remoteActionFactoryInstance: ActionFactoryInstance<Tv2RemoteActionFactory>
+  private transitionEffectActionFactoryInstance: ActionFactoryInstance<Tv2TransitionEffectActionFactory>
+  private audioActionFactoryInstance: ActionFactoryInstance<Tv2AudioActionFactory>
+  private graphicsActionFactoryInstance: ActionFactoryInstance<Tv2GraphicsActionFactory>
+  private videoClipActionFactoryInstance: ActionFactoryInstance<Tv2VideoClipActionFactory>
+  private videoMixerConfigurationActionFactoryInstance: ActionFactoryInstance<Tv2VideoMixerConfigurationActionFactory>
+  private splitScreenActionFactoryInstance: ActionFactoryInstance<Tv2SplitScreenActionFactory>
+  private replayActionFactoryInstance: ActionFactoryInstance<Tv2ReplayActionFactory>
+  private robotActionFactoryInstance: ActionFactoryInstance<Tv2RobotActionFactory>
```
KvelaGorrrrnio, line 68: These might be undefined :)

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
KvelaGorrrrnio: Isn't it `configuration` that is the old configuration? (except for first call where it acts as both.

And personally I prefer `newConfiguration` over `c`, e.g. due to the potential confusion of what is the original and what is the changed :)

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
```diff
@@ -0,0 +1,328 @@
...
+      () => {
+        return new Tv2RobotActionFactory(
+          this.createRobotTimelineObjectFactory()
+        )
+      },
+      (): boolean => {
+        return false // Currently we have no configuration changes that warrant a new RobotActionFactory.
+      },
+      configuration
+    )
```
KvelaGorrrrnio, line 317: Personal opinion incoming: When we have lambda functions that have a simple expression that is returned right away, the curly brackets and `return` is unnecessary. For me it becomes more cluttered to read, e.g. since it leads to more indentation in the code. I find the below code a bit easier to read :)
```typescript
    this.robotActionFactoryInstance = this.getUpdatedActionFactoryInstance(
      this.robotActionFactoryInstance,
      () => new Tv2RobotActionFactory(this.createRobotTimelineObjectFactory()),
      (): boolean => false, // Currently we have no configuration changes that warrant a new RobotActionFactory.
      configuration
    )
```

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
```diff
@@ -0,0 +1,328 @@
...
+    if (!!actionFactoryInstance && !actionFactoryInstance.shouldFactoryBeRecreated(configuration)) {
+      return actionFactoryInstance
+    }
+
+    actionFactoryInstance = {
+      factory: createFactoryCallback(),
+      shouldFactoryBeRecreated: shouldFactoryBeRecreatedCallback
+    }
+
+    return actionFactoryInstance
```
KvelaGorrrrnio: We should not overwrite parameters. Instead the created `ActionFactoryInstance` should be returned directly:
```typescript
    return {
      factory: createFactoryCallback(),
      shouldFactoryBeRecreated: shouldFactoryBeRecreatedCallback
    }
```

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
```diff
@@ -0,0 +1,328 @@
...
+      configuration
+    )
+
+    return this.cameraActionFactoryInstance.factory
+  }
+
+  private getUpdatedActionFactoryInstance<T>(
+    actionFactoryInstance: ActionFactoryInstance<T> | undefined,
+    createFactoryCallback: () => T,
+    shouldFactoryBeRecreatedCallback: (configuration: Tv2BlueprintConfiguration) => boolean,
```
KvelaGorrrrnio, line 91: Might just be a personal opinion, but the `Callback` suffix is not needed :)

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
```diff
@@ -0,0 +1,328 @@
...
+    configuration: Tv2BlueprintConfiguration
+  ): ActionFactoryInstance<T> {
+    if (!!actionFactoryInstance && !actionFactoryInstance.shouldFactoryBeRecreated(configuration)) {
```
KvelaGorrrrnio: `!!` is unnecessary :)

KvelaGorrrrnio (approved):

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
LindvedKrvang: Only at instantiation of the class. They aren't allow to be assigned undefined.
If we should change anything we should have a constructor that initializes them

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
LindvedKrvang: No, `configuration` is the new configuration since that is the newest configuration provided when the `createXActionFactory()` is being called.

But `configuration` becomes the old configuration the _next_ time the `createXActionFactory()` is called.

This is kinda why I am, in this case, more for just keeping them as `c` and `configuration`.

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
LindvedKrvang: Personal opinion, but I like the `Callback` suffix when the parameter is a function. Makes it clear that it is a function that has been provided from outside of the scope.

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
LindvedKrvang: Personally opinion, but I do prefer them there. Makes it clear that we are interested in the boolean value of it being there or not.
However, I have removed them.

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
LindvedKrvang: Updated

src/blueprints/tv2/action-factories/tv2-action-factory-provider.ts
KvelaGorrrrnio: Also after instantiation of the class. You would have to call all methods (that can cache a factory) for them to not be undefined.

I like the idea to have them initialized in a constructor.

PR #130, opened 2024-04-23
Sof 1897/transition effect actions lacks information

LindvedKrvang:
TransitionEffect Actions now clearly distincts between if it's a OnNext or TakeWithTransition Action.
Note: `Mix` and `Dip` Actions should not be prefixed with `Effect`.

### DIFF
--- a/src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
+++ b/src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
@@ -204,7 +204,8 @@ export class Tv2TransitionEffectActionFactory {
   }
 
   private getEffectNamePrefix(effectName: string): string {
-    if ([SpecialEffectName.MIX.toString(), SpecialEffectName.DIP.toString()].includes(effectName)) {
+    const effectNamesWithoutPrefix: string[] = [SpecialEffectName.MIX, SpecialEffectName.DIP]
+    if (effectNamesWithoutPrefix.includes(effectName)) {
       return ''
     }
     return 'Effect '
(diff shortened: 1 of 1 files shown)


RasmusAlbrektsen (approved):

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -186,20 +187,59 @@ export class Tv2TransitionEffectActionFactory {
...
+      case PieceActionType.REPLACE_PIECE:
+      case PieceActionType.INSERT_PIECE_AS_ON_AIR:
+      default: {
+        throw new Tv2MisconfigurationException(`ActionType ${actionType} is currently not supported.`)
+      }
+    }
+  }
+
+  private getEffectNamePrefix(effectName: string): string {
+    if ([SpecialEffectName.MIX.toString(), SpecialEffectName.DIP.toString()].includes(effectName)) {
```
KvelaGorrrrnio: The array can be extracted out into a `string[]` constant make it clearer why these are used and to remove the need for the `toString()` calls.  I'll go with a name that describes what is happening here, but feel free to give it a better name:
```typescript
const SPECIAL_EFFECTS_WITHOUT_EFFECT_PREFIX: string[] = [SpecialEffectName.MIX, SpecialEffectName.DIP]


class Tv2TransitionEffectActionFactory {
  private getEffectNamePrefix(effectName: string): string {
    if (SPECIAL_EFFECTS_WITHOUT_EFFECT_PREFIX.includes(effectName)) {
      return ''
    }
  }
}
```

An alternative can be to use a switch case.

KvelaGorrrrnio (approved):

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
LindvedKrvang: The idea of the enum `SpecialEffect` was to convey what you want to convey with `SPECIAL_EFFECTS_WITHOUT_EFFECT_PREFIX`.
Introducing both the enum and the constant for the same thing is overkill in my opinion.

I've updated it so your constant is now a local variable called `effectNamesWithoutPrefix` and my `toString()` methods are gone.

KvelaGorrrrnio (approved):

PR #133, opened 2024-04-24
SOF-1902 - fix split screen actions to start from 1 not from 0 index

vivanov90:
Fix split screen actions count start from one not from zero

### DIFF
--- a/src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
+++ b/src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
@@ -53,7 +53,7 @@ import { Tv2ActionManifestMapper } from '../helpers/tv2-action-manifest-mapper'
 import { Tv2ActionManifest } from '../value-objects/tv2-action-manifest'
 import { Tv2PieceInterface } from '../entities/tv2-piece-interface'
 
-const NUMBER_OF_SPLIT_SCREEN_BOXES: number = 5
+const NUMBER_OF_SPLIT_SCREEN_BOXES: number = 4
 
 // The "Layout" priority must be lower than the "Insert" priority for the inserted sources to "persist" through a Take.
 const LAYOUT_TIMELINE_OBJECT_PRIORITY: number = 0.5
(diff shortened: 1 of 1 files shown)


src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -247,7 +247,7 @@ export class Tv2SplitScreenActionFactory {
 
   private createInsertToInputActionsForSources(blueprintConfiguration: Tv2BlueprintConfiguration, sources: Tv2SourceMappingWithSound[], name: string, audioMode: Tv2AudioMode = Tv2AudioMode.FULL): Tv2SplitScreenInsertSourceInputAction[] {
     const actions: Tv2SplitScreenInsertSourceInputAction[] = []
-    for (let inputIndex = 0; inputIndex < NUMBER_OF_SPLIT_SCREEN_BOXES; inputIndex++) {
+    for (let inputIndex = 1; inputIndex < NUMBER_OF_SPLIT_SCREEN_BOXES; inputIndex++) {
```
LindvedKrvang: This doesn't actually fix the issue.
The issue is that what the producers thinks is box number one, is actually box number zero, because the input boxes are _also_ zero-indexed. 
By starting the loop on one, you don't change that. You just don't have an Action that can insert into the first box (box zero) anymore. The Action that says it inserts into box one still inserts into box two.

The solution is to keep the index as is, but just "bump" the number when we display it.
Or do `inputIndex: inputIndex -1` when you create the `metadata` for this Action.

LindvedKrvang (changes requested):

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -520,8 +524,8 @@ export class Tv2SplitScreenActionFactory {
   private createInsertLastVideoClipToInputActions(blueprintConfiguration: Tv2BlueprintConfiguration): Tv2SplitScreenInsertLastVideoClipInputAction[] {
     const actions: Tv2SplitScreenInsertLastVideoClipInputAction[] = []
     for (let inputIndex = 0; inputIndex < NUMBER_OF_SPLIT_SCREEN_BOXES; inputIndex++) {
-      actions.push(this.createInsertLastVideoClipToInputAction(blueprintConfiguration, inputIndex, Tv2AudioMode.FULL))
-      actions.push(this.createInsertLastVideoClipToInputAction(blueprintConfiguration, inputIndex, Tv2AudioMode.VOICE_OVER))
+      actions.push(this.createInsertLastVideoClipToInputAction(blueprintConfiguration, this.createDisplayIndex(inputIndex), Tv2AudioMode.FULL))
+      actions.push(this.createInsertLastVideoClipToInputAction(blueprintConfiguration, this.createDisplayIndex(inputIndex), Tv2AudioMode.VOICE_OVER))
```
LindvedKrvang: This doesn't solve it for the `InsertLastVideoClipAction`.
You have to call `createDisplayIndex` within `createInsertLastVideoClipToInputAction` method since the parameter `inputIndex` in that method is both given to the metadata and as the name. And as we talked about, the name should be one higher than the one given to the metadata :)

LindvedKrvang (approved):

KvelaGorrrrnio (approved):

PR #138, opened 2024-04-26
Sof 1922/Create Piece Replaced Event

RasmusAlbrektsen:
This PR is to help the issue of us replacing a piece on the backend but not describing this to the frontend other than sending a new piece that we insert. 
To do this the PieceReplaced event is emitted instead!


src/presentation/value-objects/rundown-event.ts
```diff
@@ -55,6 +55,12 @@ export interface PieceInsertedEvent extends PartEvent {
   piece: PieceDto
 }
 
+export interface PieceReplacedEvent extends PartEvent {
+  type: RundownEventType.PIECE_REPLACED
+  replacedPiece: PieceDto
+  newPiece: PieceDto
```
LindvedKrvang: Do we need to send the entire Piece that was replaced? Wouldn't it be enough to just send the id?

LindvedKrvang (approved):

src/presentation/value-objects/rundown-event.ts
RasmusAlbrektsen: A good point, i'll try that out :-)

KvelaGorrrrnio (approved):

src/presentation/value-objects/rundown-event.ts
RasmusAlbrektsen: I tried it out and liked it a lot.

PR #139, opened 2024-04-26
SOF-1916 Generate Actions on Ingest

LindvedKrvang:
Actions are now generated at the end of Ingest and an ActionUpdated event is emitted


src/business-logic/services/ingest-data-changed-service.ts
```diff
@@ -254,6 +277,8 @@ export class IngestDataChangedService implements DataChangeService {
     }
     const eventCallback: (() => Promise<void>) | undefined = this.getEventToExecute()
     if (!eventCallback) {
+      void this.generateActions()
+
```
RasmusAlbrektsen: Superfluous newline here, in my opinion

RasmusAlbrektsen (approved):
LGTM

src/business-logic/services/ingest-data-changed-service.ts
LindvedKrvang: Since you've put it so nicely, I have removed it ;)

src/business-logic/services/ingest-data-changed-service.ts
```diff
@@ -254,6 +277,7 @@ export class IngestDataChangedService implements DataChangeService {
     }
     const eventCallback: (() => Promise<void>) | undefined = this.getEventToExecute()
     if (!eventCallback) {
+      void this.generateActions()
```
KvelaGorrrrnio: We should not use void. If generateAction fails, then the entire server will crash.
Instead add a `.catch` to the call and log the error. E.g.:
```typescript
this.generateActions().catch(error => this.logger.data(error).error('Failed generating actions for ingest batch.'))
```

src/business-logic/services/ingest-data-changed-service.ts
```diff
@@ -66,7 +83,8 @@ export class IngestDataChangedService implements DataChangeService {
   private readonly logger: Logger
   private isExecutingEvent: boolean = false
   private lastBulkExecutionStartTimestamp: number = 0
-  private readonly rundownIdsToBuild: Set<string> = new Set<string>()
+  private readonly rundownIdsToBuild: Set<string> = new Set()
```
KvelaGorrrrnio: How does one build a rundown id?
If it is the timeline, then the same syntax as `rundownIdsToGenerateActionsFor` could be used.

src/business-logic/services/ingest-data-changed-service.ts
```diff
@@ -275,6 +299,23 @@ export class IngestDataChangedService implements DataChangeService {
     return events?.shift()
   }
 
+  private async generateActions(): Promise<void> {
+    for (const rundownId of this.rundownIdsToGenerateActionsFor) {
+      await this.generateActionsForRundown(rundownId)
+    }
+    this.rundownIdsToGenerateActionsFor.clear()
```
KvelaGorrrrnio: We should note that the sequential nature of this can become an issue, if there is at least one bad rundown.

An alternative could be to await a `Promise.allSettled` and remove the ids of the settled ones. E.g.:
```typescript
    await Promise.allSettled(
      [...this.rundownIdsToGenerateActionsFor].map(async (rundownId) => {
        await this.generateActionsForRundown(rundownId)
        this.rundownIdsToGenerateActionsFor.delete(rundownId)
      })
    )
```
This way the rundowns that can generate actions will still be generated even if there is a bad rundown in the set.

src/data-access/repositories/mongo/mongo-action-repository.ts
```diff
@@ -27,6 +27,18 @@ export class MongoActionRepository extends BaseMongoRepository implements Action
     return action
   }
 
+  public async getActions(): Promise<Action[]> {
+    this.assertDatabaseConnection(this.getActions.name)
+    return await this.getCollection().find<Action>({ rundownId: { $exists: false } }).toArray()
```
KvelaGorrrrnio: Since there is no error handling in this method, the `await` is redundant.

src/business-logic/services/ingest-data-changed-service.ts
LindvedKrvang: Updated

src/data-access/repositories/mongo/mongo-action-repository.ts
LindvedKrvang: Updated

src/business-logic/services/ingest-data-changed-service.ts
LindvedKrvang: Updated

src/business-logic/services/ingest-data-changed-service.ts
LindvedKrvang: Updated

KvelaGorrrrnio (approved):

PR #149, opened 2024-05-08
Mongo migration demo

LindvedKrvang:
Small proof of concept demo using the Migrate-mongo library https://github.com/seppevs/migrate-mongo.

The migration utilizes the Mongo DB object so writing our migration would be no different than writing a mongo query.
The migration does also receives a MongoClient object as a second parameter (not shown in the example migrations here).

Migrate-mongo does not support Typescript out of the box, but this [answer](https://github.com/seppevs/migrate-mongo/issues/79#issuecomment-689834972) makes it work with Typescript so we have typesafety.
Only issue is that the `migrate:create` does not work on Windows, but it is possible to create the migrations manually.

This migration framwork works on a `up` and `down` principle. The `up` is our actually migration and the `down` is to roll back said migration.


package.json
```diff
@@ -21,7 +21,11 @@
     "init-replica-set": "ts-node ./database/manager/mongo.ts --initrs",
     "seed-database": "ts-node ./database/manager/mongo.ts --seed",
     "spy-database": "ts-node ./database/manager/mongo.ts --spy",
-    "dump-database": "ts-node ./database/manager/mongo.ts --dump"
+    "dump-database": "ts-node ./database/manager/mongo.ts --dump",
+    "migrate:create": "sh -c 'ts-node node_modules/migrate-mongo/bin/migrate-mongo create $0 -f src/data-access/migrations/mongo/migrate-mongo-config.ts'",
+    "migrate:status": "ts-node node_modules/migrate-mongo/bin/migrate-mongo status -f src/data-access/migrations/mongo/migrate-mongo-config.ts",
+    "migrate:undo-last": "ts-node node_modules/migrate-mongo/bin/migrate-mongo down -f src/data-access/migrations/mongo/migrate-mongo-config.ts && npm run migrate:status",
+    "migrate": "ts-node node_modules/migrate-mongo/bin/migrate-mongo up -f src/data-access/migrations/mongo/migrate-mongo-config.ts"
```
KvelaGorrrrnio: it would nice if these used same naming convention with hyphens as word separators. Whether the word placement should be reversed can be up for debate. I suggest the following:
```json
{
  "migrate": "..."
  "undo-last-migration": "...",
  "migration-status": "...",
  "create-migration": "..."
}
```

src/data-access/migrations/mongo/mongo-migrations/migration-template.ts
KvelaGorrrrnio, line 1: Is this used for anything?

src/data-access/migrations/mongo/migrate-mongo-config.ts
```diff
@@ -0,0 +1,43 @@
...
+  // Enable the algorithm to create a checksum of the file contents and use that in the comparison to determine
+  // if the file should be run.  Requires that scripts are coded to be run multiple times.
+  useFileHash: false,
+
+  // Don't change this, unless you know what you're doing
+  moduleSystem: 'commonjs',
+}
+
+// eslint-disable-next-line no-undef
+module.exports = config
```
KvelaGorrrrnio, line 36: Is it necessary to use CJS syntax for exporting? Can't we default export the config with ESM syntax?

src/data-access/migrations/mongo/migrate-mongo-config.ts
```diff
@@ -0,0 +1,43 @@
...
+  mongodb: {
+    url: MONGO_CONNECTION_STRING,
+
+    databaseName: MONGO_DB_NAME,
+
+    options: {
+      // useNewUrlParser: true, // removes a deprecation warning when connecting
+      // useUnifiedTopology: true, // removes a deprecating warning when connecting
+      //   connectTimeoutMS: 3600000, // increase connection timeout to 1 hour
+      //   socketTimeoutMS: 3600000, // increase socket timeout to 1 hour
```
KvelaGorrrrnio: Are there any of these we should use?

src/data-access/migrations/mongo/mongo-migrations/migration-template.ts
LindvedKrvang: It's a template so you can see how to write a migration. It's written in the file.

src/data-access/migrations/mongo/migrate-mongo-config.ts
LindvedKrvang: I suppose not. This is what it defaults to when creating the config

src/data-access/migrations/mongo/migrate-mongo-config.ts
LindvedKrvang: I don't think so. I have removed them

package.json
LindvedKrvang: I have no preferences. I have updated to your suggestion

RasmusAlbrektsen (approved):

src/data-access/migrations/mongo/mongo-migrations/migration-template.ts
KvelaGorrrrnio: The TODOs in up and down, are there anything to do there? To me they just look like regular comments.

KvelaGorrrrnio (dismissed):
We should register the migration process with the production deployment, e.g. in the Dockerfile.

LindvedKrvang:
@KvelaGorrrrnio Would be awesome if you gave some context
image

KvelaGorrrrnio:
> @KvelaGorrrrnio Would be awesome if you gave some context image

Sure. I was unable to run the migrations due to the template file being there. Failed saying that it expected exported functions `up` and `down`.

LindvedKrvang:
> > @KvelaGorrrrnio Would be awesome if you gave some context image
> 
> Sure. I was unable to run the migrations due to the template file being there. Failed saying that it expected exported functions `up` and `down`.

I have moved the migration-template out of the migration folder.

src/data-access/migrations/mongo/migration-template.ts
KvelaGorrrrnio, line 1: As we talked about, it could be an idea to move the code from the migration template to a README in the mongo-migrations folder. The README could then contain the code snippet and a description of how the migration setup works. That way the workflow is documented and the code template is more visible. A potential starting point:
````md
# MongoDB migrations

Sofie server uses the library [migrate-mongo](https://www.npmjs.com/package/migrate-mongo) for running the migrations.
A migration consists of `up` and `down` methods, where `up` applies the migration and `down` rolls back the changes from `up`.

```typescript
import * as mongodb from 'mongodb'

export async function up(database: mongodb.Db, client: mongodb.MongoClient): Promise<void> {
  // TODO: Write the migration.
}

export async function down(database: mongodb.Db, client: mongodb.MongoClient): Promise<void> {
 // TODO: Write the rollback procedure for the migration.
}
```

The migrations are applied in alphanumeric order, so the migration name has the format `<timestamp>-<descriptive-name>.ts`.
A new migration can be created with `yarn create-migration <descriptive-name>`.
````

LindvedKrvang: Updated

KvelaGorrrrnio (approved):

PR #150, opened 2024-05-10
SOF-1975 Introduce abstact ActionFactory class. Sanitize ActionIds

LindvedKrvang:
- Introduce abstract ActionFactory to keep common sanitize method
- Use sanitize method in ActionFactories

### DIFF
--- a/src/blueprints/tv2/action-factories/ActionFactory.ts
+++ b/src/blueprints/tv2/action-factories/ActionFactory.ts
@@ -1,0 +1,5 @@
+export abstract class ActionFactory {
+  protected sanitizeStringForId(value: string): string {
+    return value.replaceAll(/\s/g, '_').replaceAll('/', '_')
+  }
+}
--- a/src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
+++ b/src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
@@ -22,15 +22,18 @@ import { Tv2PieceType } from '../enums/tv2-piece-type'
 import {
   Tv2VideoClipTimelineObjectFactory
 } from '../timeline-object-factories/interfaces/tv2-video-clip-timeline-object-factory'
+import { ActionFactory } from './ActionFactory'
 
 const FRAME_RATE: number = 25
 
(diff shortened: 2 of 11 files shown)


src/blueprints/tv2/action-factories/ActionFactory.ts
```diff
@@ -0,0 +1,5 @@
+export abstract class ActionFactory {
+  protected sanitizeStringForId(value: string): string {
+    return value.replaceAll(/\s/g, '_').replaceAll('/', '_')
```
KvelaGorrrrnio, line 3: Isn't the `g` in the regular expression unnecessary, since `replaceAll` is used?

src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
```diff
@@ -22,15 +22,18 @@ import { Tv2PieceType } from '../enums/tv2-piece-type'
 import {
   Tv2VideoClipTimelineObjectFactory
 } from '../timeline-object-factories/interfaces/tv2-video-clip-timeline-object-factory'
+import { ActionFactory } from './ActionFactory'
 
 const FRAME_RATE: number = 25
 
-export class Tv2AudioActionFactory {
+export class Tv2AudioActionFactory extends ActionFactory {
```
KvelaGorrrrnio, line 29: What is the reason for using inheritance over composition here? :)

KvelaGorrrrnio (approved):

src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
LindvedKrvang: Less dependencies to inject. I don't want to end up in a place where we have to inject tons of "helper" classes.

src/blueprints/tv2/action-factories/ActionFactory.ts
LindvedKrvang: I think you are right. I just copied the regex from one of the ActionFactories

RasmusAlbrektsen (approved):

PR #152, opened 2024-05-14
Sof 1979/add invalidity to segment

LindvedKrvang:
Note: This PR is dependent on https://github.com/tv2/sofie-blueprints-inews/pull/243
Update: All dependent PRs have been merged.

Segment now has the `invalidity` attribute.
Invalid Segments are not including in "rundown execution"


src/data-access/repositories/mongo/mongo-ingested-entity-converter.ts
```diff
@@ -57,6 +57,7 @@ export interface MongoIngestedSegment extends MongoId {
   identifier?: string
   metaData?: unknown // This is the current spelling in the database from Core... TOD: Update when we control Ingest
   budgetDuration?: number
+  invalidity: { reason: string }
```
RasmusAlbrektsen: I'm probably not considering something, so mostly for curiosity:
Why is this not optional on ingested entities?

RasmusAlbrektsen (approved):
Just a question!

src/data-access/repositories/mongo/mongo-ingested-entity-converter.ts
LindvedKrvang: Because of oversight.
It is now optional :)

src/model/entities/rundown.ts
```diff
@@ -495,6 +496,10 @@ export class Rundown extends BasicRundown {
     const nextSegment: Segment = this.findSegment(segmentId)
     const nextPart: Part = nextSegment.findPart(partId)
 
+    if (nextSegment.invalidity) {
+      throw new InvalidSegmentException(`Can't set Segment ${nextSegment.name}. The Segment is invalid.`)
```
KvelaGorrrrnio: Seems like something is missing in the error message about what can't be set. Suggestion:
```typescript
`Unable to set segment '${nextSegment.name}' as next, since it is invalid.`
```
And should we include the invalidity reason in the message?

src/model/entities/segment.ts
```diff
@@ -63,10 +68,18 @@ export class Segment {
   }
 
   public putOnAir(): void {
+    this.assertNotInvalid(this.putOnAir.name)
     this.isSegmentOnAir = true
     this.executedAtEpochTime ??= Date.now()
   }
 
+  private assertNotInvalid(operationName: string): void {
```
KvelaGorrrrnio: Suggestion: Instead of the negation, the method can the assert validity, e.g. `assertValidity`.

src/model/entities/segment.ts
```diff
@@ -63,10 +68,18 @@ export class Segment {
...
+    if (!this.invalidity) {
+      return
+    }
+    throw new InvalidSegmentException(`Segment "${this.name}" is invalid. Unable to do "${operationName}".`)
```
KvelaGorrrrnio: I suggest that we reverse the order of the points, such that the effect is stated first and then the cause. e.g.
```typescript
`Unable to do '${operationName}', since segment '${this.name}' is invalid.`
```
Should we add the invalidity reason to here?

src/model/entities/test/rundown.spec.ts
```diff
@@ -3459,6 +3461,77 @@ describe(Rundown.name, () => {
         expect(result).toThrow(OnAirException)
       })
     })
+
+    describe('next Segment is invalid', () => {
+      it('does not call setNext on the Segment', () => {
+        const activePart: Part = EntityMockFactory.createPart({ id: 'active-part-id', isOnAir: true })
+        const activeSegment: Segment = EntityTestFactory.createSegment({ id: 'active-segment-id', parts: [activePart] })
```
KvelaGorrrrnio, line 3468: Just a reminder of the mixed use in the codebase of `onAirPart` and `activePart` for the same thing. Same for segment.

src/model/value-objects/invalidity.ts
```diff
@@ -0,0 +1,3 @@
+export interface Invalidity {
+  reason: string
```
KvelaGorrrrnio: The reason for invalidity in an instance should not change, so this can be made readonly.

src/model/value-objects/invalidity.ts
```diff
@@ -0,0 +1,3 @@
+export interface Invalidity {
```
KvelaGorrrrnio, line 1: Do we expect to use this for other things than segments?

KvelaGorrrrnio (approved):

src/data-access/repositories/mongo/mongo-ingested-entity-converter.ts
RasmusAlbrektsen: Well that makes sense then! :-D

src/model/value-objects/invalidity.ts
LindvedKrvang: Currently no, but the concept of invalidity is not unique to Segments so I could see it happen.
That being said, the main reason I created the separate interface/file was so that I could use the `Invalidity` as a return type.

src/model/value-objects/invalidity.ts
LindvedKrvang: Updated

src/model/entities/segment.ts
LindvedKrvang: Updated the message.

Also not sure if we should include the reason...

src/model/entities/segment.ts
LindvedKrvang: Renamed the method

src/model/entities/rundown.ts
LindvedKrvang: Updated the message.

I'm not sure whether we should include the reason in the error thrown...

src/model/entities/rundown.ts
```diff
@@ -495,6 +496,10 @@ export class Rundown extends BasicRundown {
     const nextSegment: Segment = this.findSegment(segmentId)
     const nextPart: Part = nextSegment.findPart(partId)
 
+    if (nextSegment.invalidity) {
+      throw new InvalidSegmentException(`Unable to set segment ${nextSegment.name} as next, since it is invalid.`)
```
KvelaGorrrrnio: Should `${nextSegment.name}` be in quotation marks?

KvelaGorrrrnio (approved):

src/model/entities/rundown.ts
LindvedKrvang: Sure. It has been updated.

KvelaGorrrrnio (approved):

PR #155, opened 2024-05-29
Sof 1999/unplanned parts becomes unplanned on server restart

LindvedKrvang:
No longer mark unplanned Parts as unsynced. For a Part to be unsynced it needs to be unsynced from something. Unplanned Parts aren't included from the Newsroom so they technically can't be unsynced.

Also in this PR:
- Parts inserted into a Rundown will now get its rank updated to reflect its position in the Segment it is being inserted into.

### DIFF
--- a/src/model/entities/rundown.ts
+++ b/src/model/entities/rundown.ts
@@ -707,18 +707,23 @@ export class Rundown extends BasicRundown {
   }
 
   private updateRankFromOnAirAndNextParts(partToBeUpdated: Part): void {
-    const onAirRank: number = this.activeCursor?.part.getRank() ?? 0
-    const nextRank: number = this.nextCursor?.part.getRank() ?? 0
-
-    if (this.activeCursor?.segment.id !== this.nextCursor?.segment.id) {
-      // The new rank is onAir Parts rank times 1.5
-      partToBeUpdated.updateRank(onAirRank * 1.5)
+    if (!this.activeCursor) {
       return
     }
 
-    // The new rank is half the 'distance' between the onAir and next Parts plus the rank of the onAir Part.
-    const newRank: number = (nextRank - onAirRank) / 2 + onAirRank
-    partToBeUpdated.updateRank(newRank)
+    const onAirPart: Part = this.activeCursor.part
+    const onAirSegment: Segment = this.activeCursor.segment
+
+    try {
+      const partAfterOnAirPart: Part = onAirSegment.findNextPart(onAirPart)
+      const newRank: number = (partAfterOnAirPart.getRank() - onAirPart.getRank()) / 2 + onAirPart.getRank()
+      partToBeUpdated.updateRank(newRank)
+    } catch (error) {
+      if (!(error instanceof LastPartInSegmentException)) {
+        throw error
+      }
+      partToBeUpdated.updateRank(onAirPart.getRank() + 0.1)
+    }
   }
 
   public stopActivePiecesOnLayers(layers: string[]): void {
--- a/src/model/entities/rundown.ts
+++ b/src/model/entities/rundown.ts
@@ -701,8 +701,24 @@ export class Rundown extends BasicRundown {
     this.assertActive(this.insertPartAsNext.name)
     this.assertNotUndefined(this.activeCursor, 'active Segment')
 
+    this.updateRankFromOnAirAndNextParts(part)
     this.activeCursor.segment.insertPartAfterActivePart(part)
     this.setNext(this.activeCursor.segment.id, part.id)
+  }
+
+  private updateRankFromOnAirAndNextParts(partToBeUpdated: Part): void {
+    const onAirRank: number = this.activeCursor?.part.getRank() ?? 0
+    const nextRank: number = this.nextCursor?.part.getRank() ?? 0
+
+    if (this.activeCursor?.segment.id !== this.nextCursor?.segment.id) {
+      // The new rank is onAir Parts rank times 1.5
+      partToBeUpdated.updateRank(onAirRank * 1.5)
+      return
+    }
+
+    // The new rank is half the 'distance' between the onAir and next Parts plus the rank of the onAir Part.
+    const newRank: number = (nextRank - onAirRank) / 2 + onAirRank
+    partToBeUpdated.updateRank(newRank)
   }
 
   public stopActivePiecesOnLayers(layers: string[]): void {
--- a/src/model/entities/rundown.ts
+++ b/src/model/entities/rundown.ts
@@ -701,12 +701,12 @@ export class Rundown extends BasicRundown {
     this.assertActive(this.insertPartAsNext.name)
     this.assertNotUndefined(this.activeCursor, 'active Segment')
 
-    this.updateRankFromOnAirAndNextParts(part)
+    this.updateRankFromOnAirPart(part)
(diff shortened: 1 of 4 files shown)


RasmusAlbrektsen (approved):

src/model/entities/rundown.ts
```diff
@@ -701,10 +701,26 @@ export class Rundown extends BasicRundown {
...
+    const nextRank: number = this.nextCursor?.part.getRank() ?? 0
+
+    if (this.activeCursor?.segment.id !== this.nextCursor?.segment.id) {
+      // The new rank is onAir Parts rank times 1.5
+      partToBeUpdated.updateRank(onAirRank * 1.5)
+      return
+    }
+
+    // The new rank is half the 'distance' between the onAir and next Parts plus the rank of the onAir Part.
+    const newRank: number = (nextRank - onAirRank) / 2 + onAirRank
```
KvelaGorrrrnio: If the next part has a lower rank than the on air part, then the partToBeUpdated will be placed before the on air part.

src/model/entities/rundown.ts
```diff
@@ -701,10 +701,26 @@ export class Rundown extends BasicRundown {
...
     this.setNext(this.activeCursor.segment.id, part.id)
   }
 
+  private updateRankFromOnAirAndNextParts(partToBeUpdated: Part): void {
+    const onAirRank: number = this.activeCursor?.part.getRank() ?? 0
+    const nextRank: number = this.nextCursor?.part.getRank() ?? 0
+
+    if (this.activeCursor?.segment.id !== this.nextCursor?.segment.id) {
+      // The new rank is onAir Parts rank times 1.5
+      partToBeUpdated.updateRank(onAirRank * 1.5)
```
KvelaGorrrrnio: Why is the onAirRank multiplied by 1.5?

src/model/entities/rundown.ts
```diff
@@ -701,10 +701,26 @@ export class Rundown extends BasicRundown {
...
+      return
+    }
+
+    // The new rank is half the 'distance' between the onAir and next Parts plus the rank of the onAir Part.
+    const newRank: number = (nextRank - onAirRank) / 2 + onAirRank
+    partToBeUpdated.updateRank(newRank)
+  }
+
```
KvelaGorrrrnio: Both comments describe what the code does, not why.

src/model/entities/test/rundown.spec.ts
```diff
@@ -3675,6 +3675,92 @@ describe(Rundown.name, () => {
...
+
+    describe('there is an onAir Part', () => {
+      describe('next Part is in the same Segment as the onAir Part', () => {
+        it('sets the rank of the inserted Part to be between the onAir and next Parts', () => {
+          const partToBeInserted: Part = EntityTestFactory.createPart({ id: 'partToBeInserted', rank: -1, ingestedPart: undefined })
+          const onAirPart: Part = EntityTestFactory.createPart({ isOnAir: true, rank: 5 })
+          const nextPart: Part = EntityTestFactory.createPart({ isNext: true, rank: 10 })
+          const onAirSegment: Segment = EntityTestFactory.createSegment({ parts: [onAirPart, nextPart] })
+
+          // The new rank is half the 'distance' between the onAir and next Parts plus the rank of the onAir Part.
```
KvelaGorrrrnio: The comment describes the steps that the code takes not what it results in nor why.

src/model/entities/rundown.ts
```diff
@@ -701,10 +701,26 @@ export class Rundown extends BasicRundown {
     this.assertActive(this.insertPartAsNext.name)
     this.assertNotUndefined(this.activeCursor, 'active Segment')
 
+    this.updateRankFromOnAirAndNextParts(part)
     this.activeCursor.segment.insertPartAfterActivePart(part)
     this.setNext(this.activeCursor.segment.id, part.id)
   }
 
+  private updateRankFromOnAirAndNextParts(partToBeUpdated: Part): void {
```
KvelaGorrrrnio: An alternative solution can be to find the closest rank (if any) after the on air part and take find a rank in between. If there are no parts after the on air part in the segment then the "found" rank is on air rank + 1.
Pseudocode:
```typescript
private updateRankFromOnAirPart(part: Part): void {
  if (!this.activeCursor) {
    return
  }
  const onAirRank: number = this.activeCursor.part.getRank()
  const onAirPartIndex: number = this.activeCursor.segment.parts.findIndex(part => part.id === this.activeCursor.part.id)
  if (onAirPartIndex < 0) {
    return
  }
  const nextTakenRank: number = this.activeCursor.segment.parts[onAirPartIndex + 1]?.rank ?? (onAirRank + 1)
  const partRank: number = (nextTakenRank - onAirRank) / 2 + onAirRank
  part.updateRank(partRank)
}
```

KvelaGorrrrnio: Update with a bit more clean approach (based on your suggestion on slack):
```typescript
private updateRankFromOnAirPart(part: Part): void {
  if (!this.activeCursor) {
    return
  }
  const onAirRank: number = this.activeCursor.part.getRank()
  const nextTakenRank: number = this.getRankAfter(this.activeCursor.part)
  const partRank: number = (nextTakenRank - onAirRank) / 2 + onAirRank
  part.updateRank(partRank)
}

private getRankAfter(part: Part): number {
  try {
    return this.getPartAfter(part).rank
  } catch {
    return part.rank + 1
  }
}
```

src/model/entities/rundown.ts
LindvedKrvang: Nice catch. I have updated it to no longer use the next part, but use the part after the onAir part instead.

src/model/entities/rundown.ts
LindvedKrvang: It was an abitrary number. It has been updated to just add `0.1` instead now.

src/model/entities/rundown.ts
LindvedKrvang: Both comments are gone

src/model/entities/test/rundown.spec.ts
LindvedKrvang: It techincally is a why. The why and the how is just basically the same in this case.
But, it is gone.

src/model/entities/rundown.ts
LindvedKrvang: I have updated the calculation.
I am not using your suggestion since manipulating the Parts of the Segments from the Rundown is dancing on the border of the law of demeter.

LindvedKrvang: `getPartAfter` returns the next valid Part to be taken. This Part _could_ be on another Segment. So it's quite possible that the rank of the onAir Part is higher than `nextTakenRank`.

KvelaGorrrrnio: I assume that you are referring to the first suggestion with the comment about the law of demeter.
Per my second comment, I suggest that getting the rank is made its own responsibility by separating it, so the catch doesn't have to handle potential failures of the other logic in the update rank method.
The name of the method should be changed, since it no longer uses the next part.
If `0.1` is kept, then it would be nice to have a descriptive constant of what that value represents.

LindvedKrvang: Method name has been updated.

GitHub hadn't updated, so when I posted my first comment on this thread, I could only see your first comment.

KvelaGorrrrnio (approved):

PR #158, opened 2024-05-31
Rundown reset emits set next event.

KvelaGorrrrnio:
To indicate that a part is set as next when resetting the rundown, a set next event is now emitted on rundown reset.
The PR adds a few tests for the `RundownTimelineService.resetRundown` method and changes the linting setup to allow unused variables if they start with an underscore.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,7 +22,14 @@
     "semi": ["error", "never"],
     "no-void": ["error"],
     "no-unused-vars": "off",
-    "@typescript-eslint/no-unused-vars": "error",
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      {
+        "argsIgnorePattern": "^_",
+        "varsIgnorePattern": "^_",
+        "caughtErrorsIgnorePattern": "^_"
+      }
+    ],
     "@typescript-eslint/explicit-member-accessibility": [
       "error",
       {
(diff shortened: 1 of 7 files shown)


LindvedKrvang:
Why is this needed?

KvelaGorrrrnio:
> Why is this needed?

The intention is being able to have a thinner client that doesn't injects phantom events into itself to observe the behaviour of the backend.

One example is in the rundown view for our web client, when a reset event is emitted, it is up to the client to generate a phantom event such that its services and components that listen for set next events are informed. With this change, the backend gives a semantic description of what reset is. An alternative could also be to embed the information into the reset event, but I feel that gives conflicting responsibilities and can end up making the event fragile, if reset at some point needs to do something extra (e.g. put something on air).

LindvedKrvang:
> > Why is this needed?
> 
> The intention is being able to have a thinner client that doesn't injects phantom events into itself to observe the behaviour of the backend.
> 
> One example is in the rundown view for our web client, when a reset event is emitted, it is up to the client to generate a phantom event such that its services and components that listen for set next events are informed. With this change, the backend gives a semantic description of what reset is. An alternative could also be to embed the information into the reset event, but I feel that gives conflicting responsibilities and can end up making the event fragile, if reset at some point needs to do something extra (e.g. put something on air).

Since we send a SET_NEXT event on Activate, then I can't really argue against doing it on Reset as well.

LindvedKrvang (approved):
Would be nice with a test ;)

RasmusAlbrektsen (approved):

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #161, opened 2024-06-06
SOF-2007 SplitScreens blink when Taken with VideoClip

LindvedKrvang:
No longer set various duration of Piece to zero when inserting a VideoClip into a SplitScreen.

Inserting a VideoClip into a SplitScreen would set the various durations to zero. In the case of SplitScreen, we needed a `preRollDuration`, but since we overrode it to zero, we had a quarter of a second where we would show the "default" source which would give the illusion of a "blink":

### DIFF
--- a/src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
+++ b/src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
@@ -213,7 +213,7 @@ export class Tv2SplitScreenActionFactory extends ActionFactory {
 
   private createSplitScreenPieceInterface(partId: string, name: string, metadata: Tv2PieceMetadata, timelineObjects: Tv2BlueprintTimelineObject[]): Tv2PieceInterface {
     return {
-      id: `${partId}_piece`,
+      id: `${partId}_piece_${Date.now()}`,
       partId,
       name,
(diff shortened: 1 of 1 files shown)


RasmusAlbrektsen (approved):

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -231,6 +231,26 @@ export class Tv2SplitScreenActionFactory extends ActionFactory {
     }
   }
 
+  private createSplitScreenPieceInterfaceFromPiece(piece: Piece, metadata: Tv2PieceMetadata, timelineObjects: Tv2BlueprintTimelineObject[]): Tv2PieceInterface {
+    return {
+      id: `${piece.getPartId()}_piece`,
```
KvelaGorrrrnio: Won't this give multiple pieces with the same id if I e.g. change both input 1 and 2 in a split screen?

LindvedKrvang: It would. Updated to also include the timestamp of `Date.now()` to ensure uniqueness.

KvelaGorrrrnio (approved):

PR #163, opened 2024-06-10
Sof 2012/graphics actions get wrong graphics applied

LindvedKrvang:
Our Blueprints did not take ShowStyleVariants into account. So whenever we would ask for the selected graphics setup, we would always get the default and never the one matching our ShowStyleVariant. Blueprints now take ShowStyleVariants into account when mapping Cores blueprintConfiugration into TV2s.

Added validation to ShowStyleVariants to check whether they have a GfxDefaults.

### DIFF
--- a/src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
+++ b/src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
@@ -12,7 +12,7 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
 
 
   public validateConfiguration(configuration: Configuration): StatusMessage[] {
-    const tv2BlueprintConfiguration: Tv2BlueprintConfiguration = this.configurationMapper.mapBlueprintConfiguration(configuration)
+    const tv2BlueprintConfiguration: Tv2BlueprintConfiguration = this.configurationMapper.mapBlueprintConfiguration(configuration, '')
     // Add validation as needed.
     return [
       ...this.validateShowStyleConfiguration(tv2BlueprintConfiguration.showStyle)
--- a/src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
+++ b/src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
@@ -45,7 +45,7 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
 
   private validateShowStyleVariants(showStyleVariants: ShowStyleVariant[]): StatusMessage[] {
     return showStyleVariants
-      .filter(variant => !(variant.blueprintConfiguration as CoreShowStyleVariantBlueprintConfiguration).GfxDefaults)
+      .filter(variant => !Array.isArray((variant.blueprintConfiguration as CoreShowStyleVariantBlueprintConfiguration).GfxDefaults))
       .map(variant => {
         return {
           id: `${variant.id}_noGraphicsDefault`,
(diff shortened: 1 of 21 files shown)


RasmusAlbrektsen (approved):

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
```diff
@@ -39,4 +42,17 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
         })
     })
   }
+
+  private validateShowStyleVariants(showStyleVariants: ShowStyleVariant[]): StatusMessage[] {
+    return showStyleVariants
+      .filter(variant => !(variant.blueprintConfiguration as CoreShowStyleVariantBlueprintConfiguration).GfxDefaults)
```
KvelaGorrrrnio: In addition to `GfxDefaults` not being a defined attribute, shouldn't blueprints generate a status message if the `GfxDefaults` is an empty array?

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -447,12 +447,17 @@ export class MongoEntityConverter {
     }
   }
 
-  public convertShowStyle(mongoShowStyle: MongoShowStyle): ShowStyle {
+  public convertShowStyle(mongoShowStyle: MongoShowStyle, showStyleVariants: ShowStyleVariant[] = []): ShowStyle {
```
KvelaGorrrrnio: Why is `showStyleVariants` optional?

src/data-access/repositories/mongo/mongo-show-style-variant-configuration-listener.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  private listenForChanges(): void {
+    const options: ChangeStreamOptions = { fullDocument: 'updateLookup' }
+    const changeStream: ChangeStream = this.getCollection().watch<ShowStyleVariant, ChangeStreamDocument<ShowStyleVariant>>([], options)
+    changeStream.on('change', () => this.onChange())
+    this.logger.debug('Listening for ShowStyleVariantConfiguration collection changes...')
+  }
+
+  private onChange(): void {
+    // We just want to notify that a changed has been made
```
KvelaGorrrrnio: `changed` => `change`.
And why is it that we only care about the event type?

KvelaGorrrrnio (commented):
Nice work!

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
LindvedKrvang: We could also take it all the way to validating that you have picked at valid combination of Schema/design depending on the Graphic setup. But I would prefer to do the complex validation for once the settings are defined by SofieWebClient/SofieServer.

I can add the extra check for empty array now if you want.

src/data-access/repositories/mongo/mongo-show-style-variant-configuration-listener.ts
LindvedKrvang: Updated.

Because we have just need to notify that a change has happened, so we can update the Configuration. We update the entire configuration, so we don't need special treatment for the exact thing that was updated.

src/data-access/repositories/mongo/mongo-entity-converter.ts
LindvedKrvang: No particular reason. It is no longer optional.

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
KvelaGorrrrnio: As long as the checks we make ensures that the value conforms to the type, then I am satisfied.
I can see that the length is checked where it is used, so I'll leave it up to you to add it or not. A minor thing is that we only check `GfxDefaults` to be a truthy value, but not that it is an array. A suggestion here could be to check if it is an Array, e.g.
```typescript
variant => !Array.isArray((variant.blueprintConfiguration as CoreShowStyleVariantBlueprintConfiguration).GfxDefaults)
```

KvelaGorrrrnio (approved):
I added the comment: https://github.com/tv2/sofie-server/pull/163#discussion_r1634758066.

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
LindvedKrvang: Updated to check that it is an array.

KvelaGorrrrnio (approved):

PR #165, opened 2024-06-12
SOF-2011: Missing newlines in lower third graphics generated by actions

KvelaGorrrrnio:
The structured template data for graphics are located in the `content` property under the `expectedPlayoutItems` property on `AdlibPieces` coming from Sofie Core. This property is currently not mapped to `ActionManifests` and since the display name for the adlib pieces contain the necessary information, this is still kept as the base for extracting the template data.
This should be reworked when the ingest flow is rewritten for Sofie server.

This has only been tested against the test tools in Sofie Core to see that the single lines of the template data are split correctly up into multiple lines.

### DIFF
--- a/src/blueprints/tv2/timeline-state-resolver-types/tv2-caspar-cg-types.ts
+++ b/src/blueprints/tv2/timeline-state-resolver-types/tv2-caspar-cg-types.ts
@@ -14,7 +14,7 @@ export interface Tv2CasparCgTemplateData {
       display: Tv2CasparCgTemplateDisplayMode
       payload: {
         type: string
-        0: string
+        [locatorIndex: number]: string
       }
     }
     [Tv2CasparCgTemplateSlotType.IDENT]?: {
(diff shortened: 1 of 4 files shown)


RasmusAlbrektsen (approved):

src/blueprints/tv2/timeline-state-resolver-types/tv2-caspar-cg-types.ts
```diff
@@ -14,7 +14,7 @@ export interface Tv2CasparCgTemplateData {
       display: Tv2CasparCgTemplateDisplayMode
       payload: {
         type: string
-        0: string
+        [locatorIndex: number]: string
```
LindvedKrvang, line 25: I see what you did here, but I do have a slight concern that since this is the type used by TSR we should keep it "identical" to TSR as well. And in TSR the type is `0: string, 1: string`.
I know this is a bad type, but with your suggestion one could think that `2: string, 3: string` etc also are valid options.

LindvedKrvang (approved):
Tested by the power of Windows on a CasparCG ;)

https://github.com/tv2/sofie-server/assets/23049647/2d1168c1-b971-416f-a53e-996a51aee58f

src/blueprints/tv2/timeline-state-resolver-types/tv2-caspar-cg-types.ts
KvelaGorrrrnio: In the HTML graphics project that we use there is a lower third (Headline) that uses 0, 1 and 2. If we want to have the specific entries, we can do something like:
```typescript
type LowerThirdLocators =
  | { 0: string }
  | { 0: string, 1: string }
  | { 0: string, 1: string, 2: string }
```
or add all supported graphic templates from the project to the types. Personally, I think both will be a bit too cumbersome to manage – unless we decide that we want to validate the data before sending it to CasparCG.

LindvedKrvang: That's also not great. I am okay with your current solution. I just wanted to raise a little flag.

PR #167, opened 2024-06-12
SOF-2004 Implementation of guard clauses and GFX defaults validator. …

aardhuis:
…To test navigate to Sofie core -> settings -> Show Styles - Blueprint Config -> and delete values


src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
```diff
@@ -19,10 +19,16 @@ interface CoreShowStyleBlueprintConfiguration {
   GfxSchemaTemplates: CoreGraphicsSchema[]
   DVEStyles: CoreSplitScreenConfiguration[]
   BreakerConfig: CoreBreaker[]
-  Transitions: { _id: string, Transition: string }[]
+  Transitions: Transition[]
   ShowstyleTransition: string
 }
 
+// SOF-2004
```
LindvedKrvang: What is the purpose of this comment? 
If it is to tell us which task this change is from, then that is in your commit message since we put our Jira Ids in those :)

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
```diff
@@ -81,14 +87,32 @@ export class Tv2ShowStyleBlueprintConfigurationMapper {
...
       splitScreenConfigurations: this.mapSplitScreenConfigurations(coreConfiguration.DVEStyles),
-      breakerTransitionEffectConfigurations: this.mapTransitionEffectConfigurations([
-        ...coreConfiguration.Transitions.map(transition => transition.Transition),
-        coreConfiguration.ShowstyleTransition
-      ]),
+      // SOF-2004 - Need to handle undefined values from Sofie core
+      // breakerTransitionEffectConfigurations: this.mapTransitionEffectConfigurations([
+      //   ...coreConfiguration.Transitions.map(transition => transition.Transition),
+      //   coreConfiguration.ShowstyleTransition
+      // ]),
```
LindvedKrvang: Commented code should be deleted. We use a VCS so if we need to check an older iteration, we can check the history.

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
```diff
@@ -81,14 +87,32 @@ export class Tv2ShowStyleBlueprintConfigurationMapper {
...
     }
   }
 
+  private mapBreakerTransitionEffectConfigurations(coreTransitions: Transition[], showstyleTransition: string) : BreakerTransitionEffect[] {
+    //SOF-2004 insert guard clause to protect against undefined values
+    if (!coreTransitions) {
+      return []
+    }
+
+    const transitions = coreTransitions.map(transition => transition.Transition).filter(transition => transition !== undefined)
```
LindvedKrvang: It's not clear from this line what the type of `transitions` is, so a type declaration should be added.

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
```diff
@@ -81,14 +87,32 @@ export class Tv2ShowStyleBlueprintConfigurationMapper {
...
+
+    // If no undefined value is found, proceed with mapping
+    const breakerTransitionEffectConfigurations = this.mapTransitionEffectConfigurations([...transitions, showstyleTransition])
+    
+    return breakerTransitionEffectConfigurations
```
LindvedKrvang: Same comment as for `transitions`. It's not clear what the type of `breakerTransitionEffectConfiugrations` is, so a type should be added.

In this scenario you could just return the result of `this.mapTransitionEffectConfigurations(...)` directly so you don't have to create a local variable that "isn't used".

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
```diff
@@ -39,4 +40,50 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
...
+    if (!showStyleConfiguration.graphicsDefault) {
+      validationErrors.push({
+        id: 'GraphicsDefaultMissing',
+        title: 'Graphics Default Configuration',
+        message: 'Graphics default configuration is missing',
+        statusCode: StatusCode.BAD
+      })
+    } else {
+      // Check if DefaultSetupName is empty
+      if (!showStyleConfiguration.graphicsDefault.setupName?.value) {
```
LindvedKrvang: We want to reduce nesting as much as possible. Instead of doing the nested if-statements you could return early from your first if-statement:
```
private validateGraphicsDefault(...) {
  if (!showStyleConfiguration.graphicsDefault) {
    return [{
      id: '',
      ...
    }]
  }

  const validationErrors: StatusMessage[] = []
  // Rest of your if-statements from here.
}
```

We useally refer to returning early as "defensive programming" (I am aware that that term covers more than that)

LindvedKrvang (commented):
The code looks good :)

The reason I didn't approve yet is because you have a lot of comments explaining `what` the code does. Comments should explain `why` the code does as it does.
Comments that don't explain `why` should generally be deleted. (This also goes for the comments saying which task introduced the change)

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
```diff
@@ -81,14 +87,32 @@ export class Tv2ShowStyleBlueprintConfigurationMapper {
...
+        coreConfiguration.ShowstyleTransition),
       breakers: this.mapToBreakers(coreConfiguration.BreakerConfig)
     }
   }
 
+  private mapBreakerTransitionEffectConfigurations(coreTransitions: Transition[], showstyleTransition: string) : BreakerTransitionEffect[] {
+    //SOF-2004 insert guard clause to protect against undefined values
+    if (!coreTransitions) {
+      return []
+    }
```
KvelaGorrrrnio: There is a small inconsistency between the described types and the way they are used.
The type for `coreTransitions` states that it is always an array. So from the type-systems perspective, the expression `!coreTransitions` will never evaluate to true, since an empty array is also a truthy value in JavaScript land :) When working with TypeScript, there is a greater emphasis on the difference between runtime and compile-time data, than perhaps other languages.

From the comment, I assume that the guard is to add the default/fallback value as described in the task. In that case, I would alter the type to reflect that the value can be undefined. By doing that, the types model the real world/runtime data, ensuring that the programmers and the program can make more informed decisions on how to handle said data. It also makes the types tell what the comment currently describes, making the comment redundant. The same goes for the other entries :)

src/blueprints/tv2/.DS_Store
KvelaGorrrrnio, line 1: `.DS_Store` files should not be committed to the repository. I suggest that this is added to your [global gitignore](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) :)

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
aardhuis: Ah, my bad. A lot of the software houses I worked in required me to comment on in association with what task ID something changed and who changed it. Since that isn't a thing here I will remove and not use that style of commenting.

LindvedKrvang: No worries :)
My comment wasn't meant negatively and I hope it wasn't received that way.

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
aardhuis: I debated with the guys whether to use coreTransitions: Transition[] | undefined or coreTransitions?: Transition[]. 
After some debate we agreed upon the latter approach

aardhuis:
Anders and Lindved are away, so I added you guys to help this review make it through before tomorrow.

RasmusAlbrektsen (approved):

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
```diff
@@ -39,4 +40,48 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
         })
     })
   }
-}
+
+  private validateGraphicsDefault(showStyleConfiguration: Tv2ShowStyleBlueprintConfiguration): StatusMessage[] {  
+    // Check if showStyleConfiguration.graphicsDefault exists
```
MadsThy:
```suggestion
```

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
```diff
@@ -39,4 +40,48 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
...
+      return [{
+        id: 'GraphicsDefaultMissing',
+        title: 'Graphics Default Configuration',
+        message: 'Graphics default configuration is missing',
+        statusCode: StatusCode.BAD
+      }]
+    }
+
+
+    // Check if DefaultSetupName is empty
```
MadsThy:
```suggestion
```

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
```diff
@@ -39,4 +40,48 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
...
+    if (!showStyleConfiguration.graphicsDefault.setupName?.value) {
+      return [{
+        id: 'DefaultSetupName',
+        title: 'Default Setup Name Configuration',
+        message: 'The Default Setup Name is missing or empty',
+        statusCode: StatusCode.BAD
+      }]
+    }
+  
+    // Check if DefaultSchema is empty
```
MadsThy:
```suggestion
```

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
```diff
@@ -39,4 +40,48 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
...
+    if (!showStyleConfiguration.graphicsDefault.schema?.value) {
+      return [{
+        id: 'DefaultSchema',
+        title: 'Default Schema Configuration',
+        message: 'The Default Schema is missing or empty',
+        statusCode: StatusCode.BAD
+      }]
+    }
+  
+    // Check if DefaultDesign is empty
```
MadsThy:
```suggestion
```

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
```diff
@@ -21,7 +21,8 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
 
   private validateShowStyleConfiguration(showStyleConfiguration: Tv2ShowStyleBlueprintConfiguration): StatusMessage[] {
     return [
-      ...this.validateGraphicsSchemas(showStyleConfiguration)
+      ...this.validateGraphicsSchemas(showStyleConfiguration),
+      ...this.validateGraphicsDefault(showStyleConfiguration)
```
MadsThy:
```suggestion
      ...this.validateGraphicsDefaults(showStyleConfiguration)
```

src/blueprints/tv2/tv2-blueprint-configuration-validator.ts
```diff
@@ -39,4 +40,48 @@ export class Tv2BlueprintConfigurationValidator implements BlueprintValidateConf
         })
     })
   }
-}
+
+  private validateGraphicsDefault(showStyleConfiguration: Tv2ShowStyleBlueprintConfiguration): StatusMessage[] {
```
MadsThy:
```suggestion
  private validateGraphicsDefaults(showStyleConfiguration: Tv2ShowStyleBlueprintConfiguration): StatusMessage[] {  
```

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
```diff
@@ -81,14 +87,24 @@ export class Tv2ShowStyleBlueprintConfigurationMapper {
...
   }
 
+  private mapBreakerTransitionEffectConfigurations(showstyleTransition: string, coreTransitions?: Transition[]) : BreakerTransitionEffect[] {
+    if (!coreTransitions) {
+      return []
+    }
+
+    const transitions: string[] = coreTransitions.map(transition => transition.Transition).filter(transition => transition !== undefined)
+
+    // If no undefined value is found, proceed with mapping
```
MadsThy:
```suggestion
```

MadsThy (approved):

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
MadsThy: Approved with a few suggestions :)

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
RasmusAlbrektsen: This comment still survived it all :-D

RasmusAlbrektsen (approved):
The one little comment with `// SOF-2004` is still alive and well, we should get that out before merging :-)

RasmusAlbrektsen (approved):

MadsThy:
The Yarn.lock file has been removed! 🫨 We can't have that! Please put it back 😄

RasmusAlbrektsen (approved):

PR #168, opened 2024-06-13
SOF-2016: Expose invalidity for parts

KvelaGorrrrnio:
Depends on: https://github.com/tv2/sofie-blueprints-inews/pull/245.

Adds the invalidity attribute to parts (in various forms) and exposes it to the REST endpoints via the `PartDto`.

Update: The logic for ensuring that we can't take an invalid part will be made in a separate PR.


RasmusAlbrektsen (approved):

LindvedKrvang (changes requested):
It's still possible to Take an invalid Part. That should not be possible.

LindvedKrvang (approved):
Allright, you are making the it impossible to take an invalid Part in a different PR, so you get an approve anyway ;)

PR #169, opened 2024-06-13
SOF-2013 - Reingest data upon show style variant change

RasmusAlbrektsen:
This PR depends on: https://github.com/tv2/sofie-blueprints-inews/pull/246

Whenever a segment has the new property ```definesShowStyleVariant```set to true, we reingest the rundown data to ensure that all pieces have changes timeline objects that fit the new show style.

### DIFF
--- a/src/business-logic/services/ingested-entity-to-entity-mapper.ts
+++ b/src/business-logic/services/ingested-entity-to-entity-mapper.ts
@@ -60,7 +60,7 @@ export class IngestedEntityToEntityMapper {
       isNext: false,
       isUnsynced: false,
       expectedDurationInMs: ingestedSegment.budgetDuration,
-      definesShowStyleVariant: ingestedSegment.definesShowStyleVariant,
+      definesShowStyleVariant: ingestedSegment.definesShowStyleVariant ?? false,
       parts: [],
     })
   }
--- a/src/business-logic/services/ingested-entity-to-entity-mapper.ts
+++ b/src/business-logic/services/ingested-entity-to-entity-mapper.ts
@@ -60,6 +60,7 @@ export class IngestedEntityToEntityMapper {
       isNext: false,
       isUnsynced: false,
       expectedDurationInMs: ingestedSegment.budgetDuration,
+      definesShowStyleVariant: ingestedSegment.definesShowStyleVariant,
       parts: [],
     })
   }
--- a/src/business-logic/services/rundown-timeline-service.ts
+++ b/src/business-logic/services/rundown-timeline-service.ts
@@ -156,14 +156,13 @@ export class RundownTimelineService implements RundownService {
     this.rundownEventEmitter.emitSetNextEvent(rundown)
     this.startAutoNext(timeline, rundownId)
 
+    await this.deleteUnsyncedPreviousPart(rundown)
+    await this.deleteUnsyncedSegments(rundown)
+    await this.saveRundown(rundown)
 
     if (rundown.getActiveSegment().definesShowStyleVariant) {
       await this.ingestService.reloadIngestData(rundown.id)
     }
-
-    await this.deleteUnsyncedPreviousPart(rundown)
-    await this.deleteUnsyncedSegments(rundown)
-    await this.saveRundown(rundown)
   }
 
   private async deleteUnsyncedPreviousPart(rundown: Rundown): Promise<void> {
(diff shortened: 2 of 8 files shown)


src/business-logic/services/ingested-entity-to-entity-mapper.ts
```diff
@@ -60,6 +60,7 @@ export class IngestedEntityToEntityMapper {
       isNext: false,
       isUnsynced: false,
       expectedDurationInMs: ingestedSegment.budgetDuration,
+      definesShowStyleVariant: ingestedSegment.definesShowStyleVariant,
```
KvelaGorrrrnio: To remove the optionality of the attribute on the `Segment` type, a default value of `false` could be specified here.

KvelaGorrrrnio (approved):

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -154,6 +156,11 @@ export class RundownTimelineService implements RundownService {
...
     this.startAutoNext(timeline, rundownId)
 
+
+    if (rundown.getActiveSegment().definesShowStyleVariant) {
+      await this.ingestService.reloadIngestData(rundown.id)
+    }
+
     await this.deleteUnsyncedPreviousPart(rundown)
     await this.deleteUnsyncedSegments(rundown)
     await this.saveRundown(rundown)
```
LindvedKrvang: Perhaps we should start the reload flow after we are done with our own manipulations of the data and have saved that. If the reload flow is extremely fast then there is the option for a race condition here.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -154,6 +156,11 @@ export class RundownTimelineService implements RundownService {
     this.rundownEventEmitter.emitSetNextEvent(rundown)
     this.startAutoNext(timeline, rundownId)
 
+
+    if (rundown.getActiveSegment().definesShowStyleVariant) {
```
LindvedKrvang, line 163: Missing tests

src/data-access/repositories/mongo/mongo-ingested-entity-converter.ts
```diff
@@ -58,6 +58,7 @@ export interface MongoIngestedSegment extends MongoId {
   metaData?: unknown // This is the current spelling in the database from Core... TOD: Update when we control Ingest
   budgetDuration?: number
   invalidity?: { reason: string }
+  definesShowStyleVariant?: boolean
```
LindvedKrvang: I agree with Anders that this should just default to false instead of being optional.

LindvedKrvang: Goes for all the other places this is defined as well.

aardhuis (approved):

src/data-access/repositories/mongo/mongo-ingested-entity-converter.ts
RasmusAlbrektsen: It is no longer optional and now defaults to false!

src/business-logic/services/rundown-timeline-service.ts
RasmusAlbrektsen: A test has been added that verifies that the reingest is called if a segment has the new property set to true.

src/business-logic/services/rundown-timeline-service.ts
RasmusAlbrektsen: I have moved it to be after the data manipulation, good point!

src/business-logic/services/ingested-entity-to-entity-mapper.ts
RasmusAlbrektsen: It has been done so!

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -242,6 +242,56 @@ describe(RundownTimelineService.name, () => {
...
+      })
+      const mockTimelineBuilder: TimelineBuilder = mock<TimelineBuilder>()
+      when(mockTimelineBuilder.buildTimeline(aRundown)).thenResolve(mockTimeline)
+
+      const testee: RundownTimelineService = createTestee({
+        rundownEventEmitter: instance(mockRundownEventEmitter),
+        rundownRepository: instance(aRundownRepo),
+        timelineBuilder: instance(mockTimelineBuilder),
+        ingestService: instance(mockIngestService)
+      })
```
LindvedKrvang: All this is way overkill.
All you need is a test that verifies that `ingestService.reloadIngestData` is called when `segment.definesShowStyleVariant` is true. And another test that verifies that it isn't called, when it is false.

LindvedKrvang (approved):

src/business-logic/services/rundown-timeline-service.ts
LindvedKrvang: You have an if-statement. That means atleast two tests. One for going into the if-statment and one for not going into the if-statement. 
I made a comment on your test.

MadsThy (approved):
LGTM

PR #171, opened 2024-06-21
Update docker image and packages

MadsThy:
This PR:

- Gets rid of all yarn package vulnerabilities, and minimizes the vulnerabilities from the docker image we use.
- Cuts down on the image size, so we are now at 161.65 MB insterad of 275.73 MB
- Updates the check script with newest version for readability

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,38 +1,34 @@
-name: 'Check'
-run-name: 'Test, build, lint and validate'
+name: "Check"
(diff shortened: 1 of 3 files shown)


aardhuis (approved):

LindvedKrvang (approved):

MadsThy:
I forgot I had to reverify my machine again after I got the new one :) Will do this now, and hope people are ok with me merging even though I committed from a non-verified machine :)

PR #178, opened 2024-07-08
Sof 1955 multiple story items plus refactor

aardhuis:
Early push of a refactor that aims to differentiate Sofie core devices and Sofie new devices, plus the initial post of the CR (but not UD) actions in the controller/service/repository for story SOF-1955.

PR #180, opened 2024-07-11
Sof 1967 able to update device

aardhuis:
Added a controller endpoint for device updates and emitted an updated event


LindvedKrvang:
Pro tip:
This PR is dependent on https://github.com/tv2/sofie-server/pull/178. Instead of pointing this PR at staging, which shows all the changes from the dependent PR and makes it difficult to see what the actual changes for this PR is, you can have the target branch be the dependent PR.
This way we only see the relevant changes for this PR.

Also, I just noticed that you are targeting staging. Your current task is "new feautures" so they should be going to develop.

src/business-logic/services/device-service-implementation.ts
```diff
@@ -22,4 +21,9 @@ export class DeviceServiceImplementation implements DeviceService {
     await this.deviceRepository.save(device)
     this.deviceEventEmitter.emitDeviceCreatedEvent(device)
   }
-}
+
+  public async update(device: Device): Promise<void> {
+    await this.deviceRepository.update(device)
+    this.deviceEventEmitter.emitDeviceUpdatedEvent(device) // Change to updated
```
LindvedKrvang: I'm a little confused about this comment. It looks like that it might be a note for you to do later?
In any case, this comment doesn't provide much value for a maintenance developer, so it would be nice if it was deleted :)

LindvedKrvang (approved):

src/business-logic/services/device-service-implementation.ts
aardhuis: Yes, it was a note to self that got implemented then I forgot to update.
Nice catch.

src/data-access/repositories/mongo/mongo-core-device-repository.ts
```diff
@@ -12,6 +12,11 @@ export class MongoCoreDeviceRepository extends BaseMongoRepository implements De
   constructor(mongoDatabase: MongoDatabase, private readonly mongoEntityConverter: MongoEntityConverter) {
     super(mongoDatabase)
   }
+  
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
KvelaGorrrrnio, line 16: Since this is targeting develop, then this shouldn't be necessary, since `_` prefixed parameters are now allowed there :)

src/presentation/controllers/device-controller.ts
```diff
@@ -1,12 +1,14 @@
...
+import { BaseController, GetRequest, PostRequest, PutRequest, RestController } from './base-controller'
+import { Request, Response } from 'express'
+import { DeviceService } from '../../business-logic/services/interfaces/device-service'
+import { HttpErrorHandler } from '../interfaces/http-error-handler'
+import { Exception } from '../../model/exceptions/exception'
+import { HttpResponseFormatter } from '../interfaces/http-response-formatter'
+import { Device, INewsDevice, TelemetricsDevice } from '../../model/entities/device'
+import { DeviceType } from '../../model/enums/device-type'
+import { DeviceDtoInterface, INewsDeviceDto, TelemetricsDeviceDto } from '../dtos/device-dto'
+import { UnprocessableEntityException } from '../../model/exceptions/unprocessable-entity-exception'
```
KvelaGorrrrnio: This line looks like it can be removed.

KvelaGorrrrnio (approved):
Minor comments added. It also looks like linting and build failed. Approve is assumed that these are fixed :) Nice work 🚀

LindvedKrvang (approved):

src/presentation/controllers/device-controller.ts
MadsThy:
```suggestion
```

src/presentation/controllers/device-controller.ts
```diff
@@ -33,6 +35,24 @@ export class DeviceController extends BaseController {
...
 
+  private toDeviceDto(device: Device): DeviceDtoInterface | undefined {
+    if (this.isINewsDevice(device)) {
+      return new INewsDeviceDto(device)
+    }
+    if (this.isTelemetricsDevice(device)) {
+      return new TelemetricsDeviceDto(device)
+    }
+    return undefined
+  }
```
MadsThy: What do you think of this style? 😄 

```suggestion
  private toDeviceDto(device: Device): DeviceDtoInterface | undefined {
    if (this.isINewsDevice(device)) return new INewsDeviceDto(device)
    if (this.isTelemetricsDevice(device)) return new TelemetricsDeviceDto(device)
    return undefined
  }
```

src/presentation/controllers/device-controller.ts
```diff
@@ -66,21 +86,18 @@ export class DeviceController extends BaseController {
...
+  @PutRequest('/:deviceId')
+  public async updateDevice(request: Request, response: Response): Promise<void> {
+    try{
+      const device: Device = await request.body
+      const { deviceId } = request.params
 
-  private isINewsDevice(device: Device): device is INewsDevice {
-    return device.type === DeviceType.INEWS
-  }
+      if(device.id !== deviceId) throw new ConflictException('Conflict: The ID in the URL path does not match the ID in the request body.')  //to be in compliance with the REST convention (it is not a strict rule) we pass in the ID, and then we validate it against the body ID to stop IDOR attacks
```
MadsThy: Suggest to move the comment on top of the function, so it's more visible 😄 

```suggestion
      /* To be in compliance with the REST convention (it is not a strict rule) we pass in the ID,
       and then we validate it against the body ID to stop IDOR attacks */
      if (device.id !== deviceId) throw new ConflictException('Conflict: The ID in the URL path does not match the ID in the request body.')
```

src/presentation/express-error-handler.ts
```diff
@@ -38,6 +38,9 @@ export class ExpressErrorHandler implements HttpErrorHandler {
       case ErrorCode.UNPROCESSABLE_ENTITY: {
         return HttpStatusCode.UNPROCESSABLE_CONTENT
       }
+      case ErrorCode.CONFLICT:{
```
MadsThy:
```suggestion
      case ErrorCode.CONFLICT: {
```

MadsThy (approved):

PR #181, opened 2024-07-15
Fix prevent multiple autonext at the same time

LindvedKrvang:
If you to took multiple Parts with autonext right after each other, all the autonext would still be in effect which would mean that all of a sudden Sofie would do a lot of Takes.


src/business-logic/services/timeout-callback-scheduler.ts
```diff
@@ -39,6 +39,7 @@ export class TimeoutCallbackScheduler implements CallbackScheduler {
     }
     const durationToNextSchedulingInMs: number = durationToExecuteCallbackInMs / SCHEDULE_RESOLUTION
     this.timeoutIdentifier = setTimeout(() => {
+      this.stop()
```
KvelaGorrrrnio: This should be outside of the setTimeout callback, before the timeoutIdentifier value is overwritten, since this can lead to incorrect behaviour. As I understand it, if a new callback is scheduled, the old timeout should be cleared and the one should be kept.
Since `this.stop()` clears the timer tied to `this.timeoutIdentifier`, then it will always be the latest scheduled callback that gets cleared. It is not an issue in the RundownTimelineService, since we manually trigger the stop method when calling takeNext.

Since this can be a bit tricky to see through, it would be nice with a few tests for this case :)
E.g. if the first logarithmic step from the second scheduled callback triggers before the first scheduled callback, then both callbacks will reschedule and none of the callbacks will be stopped.

LindvedKrvang: Updated

KvelaGorrrrnio (approved):

aardhuis (approved):

MadsThy (approved):

PR #184, opened 2024-07-16
Sof 1955 general improvements

aardhuis:
- Fixed the body parser to return JSON on 400 bad request
- Made the toDeviceDto method return device and not undefined if it could not map it to one of our approved devices. (this avoids null in returned array)


LindvedKrvang:
It looks like you are targetting the wrong branch.
Atleast, I don't think we are ready to merge to master yet :D

src/presentation/index.ts
```diff
@@ -19,20 +20,30 @@ class SofieServer {
   public server: Express
 
   constructor() {
+    this.server = express()
```
LindvedKrvang: What's the reason for moving this line up here and not keep it in `configureServer`?

aardhuis: Stack Overflow. In an issue the approved answer changed from express.json() to body parser.json() and then moved the this.server = express() up in the constructor. I assumed it had something to do with the lifecycle events and making sure express was instantiated early enough in the loop.

src/presentation/index.ts
```diff
@@ -1,12 +1,13 @@
...
-import { RepositoryFacade } from '../data-access/facades/repository-facade'
+import express, {Express, NextFunction, Request, Response, Router} from 'express'
+import {BaseController} from './controllers/base-controller'
+import {ControllerFacade} from './facades/controller-facade'
+import {EventServerFacade} from './facades/event-server-facade'
+import {ServiceFacade} from '../business-logic/facades/service-facade'
+import {Logger} from '../logger/logger'
+import {LoggerFacade} from '../logger/logger-facade'
+import {RepositoryFacade} from '../data-access/facades/repository-facade'
+import bodyparser from 'body-parser'
```
KvelaGorrrrnio: Minor thing: Since body parser is two words, the import should use camelCasing: `bodyParser` :)

KvelaGorrrrnio (approved):

src/presentation/index.ts
LindvedKrvang: There shouldn't be any difference from having it directly in the constructor than having it as the first thing in the first method of the constructor.

Personally, I would prefer that it got back into the method it was in, so we don't mix "abstraction levels". The constructor was nice and clean with just a single level before :)

LindvedKrvang (approved):

src/presentation/index.ts
MadsThy: Yeah there is no difference in this instance :) But I like that you thought about lifecycle events! 😄 those are usually very important!

I agree with @LindvedKrvang about it just going back to configureServer here :) But good to know that you had a thought about this! I had to investigate it myself, so... good to be challenged on 😄

MadsThy (approved):

src/presentation/index.ts
aardhuis: Yep, you are completely right and had I taken a moment to watch the final layout of the code after moving stuff around I would have noticed that.

Alas, I did not. Moved it back now though.

src/data-access/repositories/mongo/mongo-device-repository.ts
```diff
@@ -32,17 +31,23 @@ export class MongoDeviceRepository extends BaseMongoRepository implements Device
     return device
   }
 
-  public async save(device: Device): Promise<void> {
+  public async save(device: Device | Omit<Device, 'id'>): Promise<void> {
```
LindvedKrvang, line 34: We use `save` other places in the code for both "create" and "update" functionality. 
This save function will always override the `id` of the Device, which is not wrong, but we should be aware that a maintenance developer might call this in the future thinking it can be used to save an existing Device.

I would suggest one of two solutions:
- Change the name from `save` to `create`.
- Merge the save an update method into one.

PR #185, opened 2024-07-17
Sof 1968 delete endpoint

aardhuis:
## SOF-1968 - Able to delete device
We need an endpoint that can delete a Device. 
Before the deletion happens we should disconnect from the Device, then delete the Device in our database.

We should emit a `DEVICE_DELETED` event when the Device is deleted


src/business-logic/services/test/device-service.spec.ts
```diff
@@ -1,41 +1,111 @@
...
-    deviceRepository = instance(mockedValue)
-    deviceEventEmitter = mock<DeviceEventEmitter>()
+import { DeviceServiceImplementation } from '../device-service-implementation'
+import { Device } from '../../../model/entities/device'
+import { DeviceRepository } from '../../../data-access/repositories/interfaces/device-repository'
+import { instance, mock, verify, when } from '@typestrong/ts-mockito'
+import { DeviceEventEmitter } from '../interfaces/device-event-emitter'
+import { EntityTestFactory } from '../../../model/entities/test/entity-test-factory'
+
+describe(DeviceServiceImplementation.name, () => {
```
LindvedKrvang, line 8: Since this file is testing `DeviceServiceImplementation` we should update the name of the file to be `device-service-implementation.spec.ts` :)

src/data-access/repositories/interfaces/device-repository.ts
```diff
@@ -1,8 +1,13 @@
...
 export interface DeviceRepository {
   getDevices(): Promise<Device[]>
+
   getDevice(deviceId: string): Promise<Device>
+
   save(device: Device): Promise<void>
+
   update(device: Device): Promise<void>
+
+  delete(deviceId: string): Promise<void>
```
LindvedKrvang: I think this might be your IDE settings, but it would be nice if it didn't add newlines to all our interfaces :)

src/presentation/controllers/device-controller.ts
```diff
@@ -94,4 +94,14 @@ export class DeviceController extends BaseController {
       this.httpErrorHandler.handleError(response, error as Exception)
     }
   }
+
+  @DeleteRequest('/:deviceId')
+  public async deleteDevice(request: Request, response: Response): Promise<void> {
+    try {
+      const {deviceId} = request.params
```
LindvedKrvang: Would be nice with a syntax similar to `const deviceId: string = request.params.deviceId`.

LindvedKrvang (approved):

src/business-logic/services/test/device-service.spec.ts
aardhuis: Are there any WebStorm tools that checks consistency between filenames and the classes within?

src/data-access/repositories/interfaces/device-repository.ts
aardhuis: Hard to find this setting. Working on it!

aardhuis: Does these settings match yours?

Skærmbillede 2024-07-18 kl  08 52 20

src/data-access/repositories/mongo/mongo-device-repository.ts
```diff
@@ -35,15 +35,24 @@ export class MongoDeviceRepository extends BaseMongoRepository implements Device
...
   public async save(device: Device): Promise<void> {
     this.assertDatabaseConnection(MongoDeviceRepository.prototype.save.name)
+    let pureDevice: Device = {} as Device
     if (!device.id) {
-      device.id = this.uuidGenerator.generateUuid() // TODO: Remove side effect
+      pureDevice = {
+        ...device,
+        id: this.uuidGenerator.generateUuid(),
+      }
     }
```
KvelaGorrrrnio: From the method type definition, the given device does have an `id`. If the method should support getting a device without id, the method definition should model it such that the method type becomes (e.g.) `save(device: Device | Omit<Device, 'id'>): Promise<void>`.

By doing this we know what variations of the data the method should work with. I would suggest that instead of using the check to make a imperative update, to compose the object directly. This also avoids the `{} as Device`.
```typescript
class MongoDeviceRepository {
  public save(device: Device | Omit<Device, 'id'>): Promise<void> {
    this.assertDatabaseConnection(MongoDeviceRepository.prototype.save.name)
    const deviceWithId: Device = {
      ...device,
      id: device.id ?? this.uuidGenerator.generateUuid(),
    }
    await this.getCollection().updateOne({ id: deviceWithId.id }, { $set: deviceWithId }, { upsert: true })
  }
}
```

src/business-logic/services/test/device-service.spec.ts
MadsThy: @aardhuis i have not found any linting or other type of rule that can check for this kind of consistency ☹️

src/data-access/repositories/interfaces/device-repository.ts
LindvedKrvang: Almost. Mine looks like this
image

LindvedKrvang (approved):

src/data-access/repositories/mongo/mongo-device-repository.ts
KvelaGorrrrnio: I might have overlooked something, but to me it looks like if the device has an id, then the empty pureDevice is used, such that the given device is not saved to the collection?

aardhuis: I did not notice this comment and am sorry about that. It seems like a sound best practice and I will include it.

MadsThy (approved):
Except for Lindveds comments, i approve of this PR :)

src/presentation/controllers/device-controller.ts
aardhuis: To address your comment about type safety I did as you suggested and moved from destructuring assignment to type assertion

KvelaGorrrrnio (approved):

PR #187, opened 2024-07-23
Sof 1965 connect to devices

aardhuis:
The draft PR with:
- initial test setup (tests not successful atm)
- DeviceConnectionFactory w/ factory logic + mapper Method
- DeviceConnectionService w/ logic to create, remove, disconnect, list, check if connection exists and get connection status
- DeviceConnection interface and telemetrics/inewsgateway implementation classes (clumsily added inline below the DeviceConnection interface... I know)

The connection logic itself is not added, just the general boilerplate to setup outbound connections of whatever type.

To be added later:
- Connection and disconnection logic with circuit breaker pattern (or perhaps a partial circuit breaker like in inewsgateway)


src/business-logic/device-factories/device-connection-factory.ts
```diff
@@ -0,0 +1,21 @@
+import { DeviceType } from '../../model/enums/device-type'
+import { Device } from '../../model/entities/device'
+import { DeviceConnection } from '../services/interfaces/inewsgateway-device-connection'
+import { INewsGatewayDeviceConnection } from '../services/inews-gateway-connection-implementation'
+import { DeviceConnectionFactory } from '../services/interfaces/device-connection-factory'
+
+export class DeviceConnectionFactoryImplementation implements DeviceConnectionFactory {
+  public createDeviceConnection(device: Device): DeviceConnection {
+    const deviceConnection: DeviceConnection = this.mapDeviceToDeviceConnection(device)
+    return deviceConnection
```
KvelaGorrrrnio: You can return the deviceConnection directly without the temporary variable :)

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export class DeviceConnectionServiceImplementation implements DeviceConnectionService {
+  private readonly connectedDevices: Map<string, DeviceAggregate> = new Map()
+
+  constructor(private readonly deviceConnectionFactory: DeviceConnectionFactory) {
+  }
+
+  public async createConnection(device: Device): Promise<DeviceConnectionStatus> {
+    if (this.connectedDevices.has(device.id)) throw new DeviceAlreadyConnectedError(device.id)
+    const deviceConnection: DeviceConnection = this.deviceConnectionFactory.createDeviceConnection(device)
```
KvelaGorrrrnio: As we talked about. Having the throw statement on a new line, makes the code a bit easier to read. When I first skimmed this, I assumed that line 19 was affected by the if statement.
It could be written as:
```typescript
if (this.connectedDevices.has(device.id)) {
  throw new DeviceAlreadyConnectionError(device.id)
}
```
The same applies to the other checks :)

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    await deviceConnection.connect()
+
+    this.connectedDevices.set(device.id, {deviceConnection, device})
+
+    return DeviceConnectionStatus.CONNECTED
```
KvelaGorrrrnio: Is it necessary to return the status code? Shouldn't we be able to assume that it was succesful based on the successfully resolved promise?

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public async send(_deviceId: string, _params: string[]): Promise<void> {
+    const conDevice: DeviceAggregate | undefined = this.connectedDevices.get(_deviceId)
+    if(conDevice === undefined) throw new DeviceNotFoundError(_deviceId)
+
+    await conDevice?.deviceConnection.send(_deviceId, _params)
+  }
```
KvelaGorrrrnio: The optional chaining should be removed, since you already checked if exists.

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    await deviceConnection.connect()
+
+    this.connectedDevices.set(device.id, {deviceConnection, device})
+
+    return DeviceConnectionStatus.CONNECTED
+  }
+
+  public async send(_deviceId: string, _params: string[]): Promise<void> {
+    const conDevice: DeviceAggregate | undefined = this.connectedDevices.get(_deviceId)
```
KvelaGorrrrnio: I assume that this is a very deceiving device that we should hold an open eye on 😛 
It would be nice with the full name `connectedDevice` 🙏

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public getConnectionStatusById(deviceId: string): DeviceConnectionStatus {
+    // we don't do thorough testing of both actual connection and the isConnected field since isConnected is an inheritance from CoreDevice and not needed here.
+    return this.connectedDevices.has(deviceId) ? DeviceConnectionStatus.CONNECTED : DeviceConnectionStatus.DISCONNECTED 
+  }
+
+  public async disconnectConnectionById(deviceId: string): Promise<DeviceConnectionStatus> {
+    await this.connectedDevices.get(deviceId)?.deviceConnection.disconnect()
+    return DeviceConnectionStatus.DISCONNECTED
```
KvelaGorrrrnio: Same comment as for the createConnection. Is it necessary to return the status? It should be obvious from the context that the device is disconnected if the call resolves successfully :)

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,100 @@
...
+    return this.connectedDevices.has(deviceId)
+  }
+
+  public listAllNetworkedDevices(): Device[] {
+    const networkedDevices: Device[] = []
+    this.connectedDevices.forEach((deviceAggregate, _key) => {
+      networkedDevices.push(deviceAggregate.device)
+    })
+    return networkedDevices
+  }
```
KvelaGorrrrnio: What does networked device mean?

And instead of iterating over each element and pushing to a separate array, you can use Array.map to make the syntax a bit more concise :)
```typescript
public getNetworkedDevices(): Device[] {
  return this.connectedDevices.map(connectedDevice => connectedDevice.device)
}
```

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+class DeviceAlreadyConnectedError extends Error {
+  constructor(deviceId: string) {
+    super(`Device with ID '${deviceId}' is already connected.`)
+    this.name = 'DeviceAlreadyConnectedError'
```
KvelaGorrrrnio: You do not need to set the name :)
Instead of extending directly from Error, these can extend the `Exception` class that is defined in the model layer.

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,100 @@
...
+  constructor(deviceId: string) {
+    super(`Device with ID '${deviceId}' is not in the collection. Have you forgot to create the connection?`)
+    this.name = 'DeviceAlreadyRemovedError'
+  }
+}
+
+export enum DeviceConnectionStatus {
+  DISCONNECTED = 0,
+  CONNECTED = 1
+}
```
KvelaGorrrrnio: We have a convention for internal enums that they the value is the same as the key. So in this case it would be
```typescript
export enum DeviceConnectionStatus {
  DISCONNECTED = 'DISCONNECTED',
  CONNECTED = 'CONNECTED',
}
```

Both this and the DeviceXXXErrors should be moved to separate files, since they are part of defining the contract with the abstraction and not specific for this implementation :)

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,64 @@
+import { Device } from '../../model/entities/device'
+import { DeviceConnection } from './interfaces/inewsgateway-device-connection'
+
+export class INewsGatewayDeviceConnection implements DeviceConnection {
+  private readonly device: Device
+  private isConnectingOrDisconnecting: boolean = false
+
+  constructor(_device: Device) {
+    this.device = _device
```
KvelaGorrrrnio: The linting is configured to allow unused variables if they are prefixed with `_`, so this and other variables that have intended use should have the prefix removed. It won't class with the class member :)

src/business-logic/services/interfaces/inewsgateway-device-connection.ts
```diff
@@ -0,0 +1,15 @@
+import { INewsGatewayDeviceConnection } from '../inews-gateway-connection-implementation'
+
+export interface DeviceConnection {
+  connect(): Promise<boolean>
+  disconnect(): Promise<boolean>
```
KvelaGorrrrnio: Is the returned boolean necessary?
If the promise fails, then that would be the indicator that the connect/disconnect did not happen.

src/business-logic/services/interfaces/inewsgateway-device-connection.ts
```diff
@@ -0,0 +1,15 @@
...
+  send(deviceId: string, params: string[]): Promise<void>
```
KvelaGorrrrnio: Just curious as to why the passed in parameters have the type `string[]`? :D

src/business-logic/services/interfaces/inewsgateway-device-connection.ts
KvelaGorrrrnio, line 1: The file is called inewsgateway-device-connection, but the interface is called `DeviceConnection`.

src/presentation/dtos/device-dto.ts
```diff
@@ -21,19 +21,19 @@ export abstract class DeviceDto {
...
-  public readonly type: DeviceType.INEWS
+  public readonly type: DeviceType.INEWSGATEWAY
   public readonly username: string
   public readonly password: string
 
-  constructor(device: INewsDevice) {
+  constructor(device: INewsGatewayDevice) {
     super(device)
-    this.type = DeviceType.INEWS
+    this.type = DeviceType.INEWSGATEWAY
```
KvelaGorrrrnio: It would be nice if `INEWSGATEWAY` was renamed to `INEWS_GATEWAY` to indicate that it is two words :)

src/business-logic/services/test/device-connection-service-implementation.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+      const testee: DeviceConnectionServiceImplementation = createTestee({deviceConnectionFactory: factoryMock})
+      
+      await testee.createConnection(device)
+      await testee.send(device.id, paramsList)
+
+      verify(factoryMock.createDeviceConnection(device)).called()
+      verify(deviceConnection.send(device.id, paramsList)).called()
+    })
+
+    it('should throw an error if the device is not connected', async () => {
```
KvelaGorrrrnio: Suggestion: If the `if the device is not connected` is moved out into a wrapping describe, then the `it` message only states the outcome and that the describes describe the scenario. It makes it a bit more nested, but makes the outcome more concise and allows for grouping related outcomes. It also gives a really nice test tree, where you can follow the branches for very complex and/or specialized use cases and see what the outcome is. E.g. in this case:
```typescript
describe(DeviceConnectionServiceImplementation.name, () => {
  describe(DeviceConnectionServiceImplementation.prototype.send.name, () => {
    describe('when the device is not connected', () => {
      it('throws an error', () => /* test body */)
    })
  })
})
```

src/business-logic/services/interfaces/inewsgateway-device-connection.ts
```diff
@@ -0,0 +1,15 @@
...
+  listen(deviceId: string, callback: (data: unknown) => void): Promise<void>
```
KvelaGorrrrnio: It would be nice with a specification of what we listen for. Is it data events, connection status changes or both? :)

KvelaGorrrrnio (commented):
Looks good! :rocket:

src/business-logic/services/device-connection-service-implementation.ts
aardhuis: You got a point. Due to time constraints I am not gonna refactor that at the moment since it is a matter of preference, but will note the "infer device connection status based on promise resolved/rejected" for the future.

src/business-logic/services/interfaces/inewsgateway-device-connection.ts
aardhuis: I started getting mental fatigue at the end of last week and decided to ignore what kind of paramList I needed and just use string array instead, then refactor later when I wasn't as tired. This will be addressed in an upcoming push.

src/business-logic/services/device-connection-service-implementation.ts
aardhuis: High context vs low context communication is, incidentally, the biggest challenge for international teams. Speaking about human interaction, of course.

I am a sucker for low-context code, but will adhere to the team standard. 
Just so I know, is this a strict requirement or a matter of taste?

src/business-logic/services/device-connection-service-implementation.ts
aardhuis: (Got a question on this topic further down this page)

src/business-logic/services/device-connection-service-implementation.ts
aardhuis: Wow, that was unexpected. Does they still convert to truthy and falsy values? What is the benefit of doing it this way? (value same as key)

I suppose magic numbers and readability will be mentioned, but going against the entire IT industry enum standard, the way every programmer since birth has been taught to use enums, does require some argumentation, methinks. 

Will comply with the code standard, but I am struggling to make sense of this.

src/business-logic/services/interfaces/inewsgateway-device-connection.ts
aardhuis: When I made the device connection service I did not know the nature of the conncections it would serve.

Would they be outbound only, or also inbound? 
Would they be all kinds of connections? TCP, UDP, web sockets, MQTT, CoAP, RESTful, AMQP, DDS, something else?

So I tried to make some assumptions based on similarities within these various protocols, but as you well know one size fits all seldom fit anything.

In the case of websocket comms with inewsgateway, I kind of want both, primarily want data events, but am unsure if DeviceConnection should handle everything so the service doesn't have to (in which case the callback wouldn't be needed at all).

src/business-logic/services/device-connection-service-implementation.ts
KvelaGorrrrnio: I assume you are referring the explicitness of the interface, when you mention that you prefer low-context code? I would like to hear more about that next time we are both at the office :)

I haven't spoken with others on the team about it, so I am not sure if it is just my opinion or a general consensus, but for me this seems like an extra layer added to the contract.

As a user of the interface, I would have to do two checks instead of one to see if the call was successful. I would first have to check if the promise was resolved or rejected. In the case, where it is rejected, I would most likely be provided with an error/exception and from that conclude that the connect attempt failed and infer the connection status to be `DISCONNECTED`.
In the case were it resolved, the returned connection status should be checked if it is `CONNECTED`. In the case where it is not, I know that it is disconnected, but I have no knowledge of why.

By not returning the DeviceConnectionStatus, I only need to check if the promise resolved or rejected. In the case it resolves, I know the connect attempt was succesful and in the case where it was rejected, I am provided with an error of some sort that describes what went wrong.

aardhuis: It's no murder mystery why I prefer explicit feedback: As you just said yourself you have no idea why because my info is not sufficient. The same can be said for reject and resolve. Why did it happen? 

But I can see that typescript(promises really) forces a standard here, and that for our purposes it is immaterial why it failed. It failed. Enough said. You want to know more, read the logs (I guess).

The reason I want explicit feedback is because I was brought up on C++ ActiveX coding, and we provided HResults with a numeric error code (horrible way of doing things, but that's another matter) whenever something went wrong. Being very explicit was part of my early coder DNA.

When I get rejected by the promise I just have to accept my fate. Javascript didn't want me. Sob.

src/business-logic/services/device-connection-service-implementation.ts
KvelaGorrrrnio: The primary reason for going with this is that it removes a layer of abstraction.
E.g. If I save a device configuration with the attribute `type: 22`, then the probability of the observer knowing what that means is quite low. If the attribute instead says `INEWS_GATEWAY`, then everyone – even non-programmers knows what kind of device we are talking about.
It does have the drawback of having a larger memory footprint, but that is also a consequence of writing out attribute names that we do as well.

Do they need to have boolean representations?
Let us say that a device has the attribute `connectionStatus`. In the case of the condition `device.connectionStatus`, would it check the existence of a connection status value or the state of said value? Another aspect is that a connection can be in more states than `CONNECTED` and `DISCONNECTED`. In the case where the configuration is missing, it could be `UNKNOWN`/`UNDEFINED`; or we could choose to model the transition states as well `CONNECTING`/`DISCONNECTING`. In these cases, what would a boolean representation mean?


I haven't talked with every programmer nor learned every curriculum for courses and talks about enums, and I haven't been in the industry for that long. So you might have a point 😁 But a lot of programming languages have quite good support for non-numerical enumerations and personally I was not indoctrinated with numerical enumerations at my education.
And as you mention, they can help against magic literals and with improving readability. Personally I think that having string-based enumerations help with achieving a self-explanatory data structure. E.g.
```JSON
{
  "type": "INEWS_GATEWAY",
  "connectionStatus": "DISCONNECTED",
  "name": "iNews Gateway"
}
```
does a better job at explaining the intent of the object than something like:
```JSON
{
  "type": 22,
  "connectionStatus": 0,
  "name": "iNews Gateway"
}
```
 If it is something you feel strongly about, then lets have a chat in team about it 😃

src/business-logic/services/interfaces/inewsgateway-device-connection.ts
KvelaGorrrrnio: A very good point! :) The simplest course of action is probably to sculpt this when integrating to the new iNews Gateway, and try to incorporate it as a somewhat general interface.

src/business-logic/services/device-connection-service-implementation.ts
KvelaGorrrrnio: I understand why you prefer the explicit feedback/typings, and so do I. I am a big sucker for strictly typed software. One big disadvantage of promises in TypeScript/JavaScript is that the rejected content can't be typed – it can be literally anything. So part of my suggestion is also based on this.
But one thing that can be done with promises, is that when rejecting e.g. the connect call, the rejection can carry a typed error which we can check for (to obtain the needed information).

E.g. in the following code snippet the error is of type unknown when caught, but we can check if it was an exception for a connection timeout or unknown host. These exceptions could also inherit from a base class e.g. `DisconnectedException`, if we don't care about the explicit reason, but just want to know that it failed and is disconnected.

```typescript
class ConnectionTimeoutException { /* implementation */ }
class UnknownHostException {
  public readonly host: string
  /* rest of the implementation */
}

// ... some code that gets the deviceConnection object ..

try {
  await deviceConnection.connect()
  console.log('Successfully connected.')
} catch (error: unknown) {
  if (error instanceof ConnectionTimeoutException) {
    console.log('Connection timed out when trying to connect.')
  } else if (error instanceof UnknownHostException) {
    console.log(`Tried connecting to unknown host ${error.host}.`)
  } else {
    console.log(`Failed connecting for unknown reason: ${error}`)
  }
}
```

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  public async disconnectConnectionById(deviceId: string): Promise<DeviceConnectionStatus> {
+    await this.connectedDevices.get(deviceId)?.deviceConnection.disconnect()
+    this.logger.info(`Device ${deviceId} disconnected`)
+    return DeviceConnectionStatus.DISCONNECTED
+  }
+
+  public async removeConnectionById(deviceId: string): Promise<DeviceConnectionStatus> {
+    const device: DeviceAggregate | undefined = this.connectedDevices.get(deviceId)
+    if (!device?.deviceConnection) {
```
KvelaGorrrrnio: Since the type of `device` is `DeviceAggregate | undefined` and `DeviceAggregate.deviceConnection` is a required attribute, this condition can be condensed to `!device`.

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,103 @@
...
+type DeviceAggregate = {
+  deviceConnection: DeviceConnection,
+  device: Device
+}
+
+export class DeviceConnectionServiceImplementation implements DeviceConnectionService {
+  private readonly connectedDevices: Map<string, DeviceAggregate> = new Map()
+
+  constructor(private readonly deviceConnectionFactory: DeviceConnectionFactory, private readonly deviceService: DeviceService, private readonly logger: Logger) {
+    this.deviceService = deviceService
```
KvelaGorrrrnio: This line is redundant. The device service is bound in the constructor parameter by the `private` keyword.

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.initialize(this.deviceService).then
```
KvelaGorrrrnio: Seems like something is missing after `.then` :)

src/business-logic/services/interfaces/device-connection.ts
```diff
@@ -0,0 +1,15 @@
...
+  connect(): Promise<boolean>
+  disconnect(): Promise<boolean>
+  send(deviceId: string, params: string[]): Promise<void>
+  listen(deviceId: string, callback: (data: unknown) => void): Promise<void>
+}
+
+export interface INewsGatewayParams {
+  headline?: string
+  content: string
+}
```
KvelaGorrrrnio: This seems to be unused.

src/model/entities/test/entity-test-factory.ts
```diff
@@ -76,6 +76,33 @@ export class EntityTestFactory {
...
+  public static createINewsGatewayDevice(device: Partial<INewsGatewayDevice> = {}): INewsGatewayDevice {
+    return {
+      id: 'deviceId',
+      name: 'deviceName',
+      statusCode: StatusCode.UNKNOWN,
+      statusMessage: '',
+      isConnected: false,
+      type: DeviceType.INEWS_GATEWAY,
+      username: 'testuser',
+      password: 'testpassword',
```
KvelaGorrrrnio: The way the new iNews Gateway is written, makes `username` and `password` unnecessary, since they are moved to and managed by the new iNews Gateway. The attributes needed by the iNewsGatewayDevice here in Sofie server are:
- `host`: The hostname or ip of the iNews gateway.
- `port`: The WebSocket port of the iNews gateway.
- `queues`: A list of queues to subscribe to. E.g. `NYHEDERNE-TEST.SOFIE.STAGING.GALLERY`, `MY.SPECIAL.QUEUE`, ...
- `password` (optional): If/when we add some sort of auth token.

src/business-logic/services/test/device-connection-service-implementation.spec.ts
```diff
@@ -0,0 +1,309 @@
...
+      const deviceServiceMock: DeviceService = mock<DeviceService>()
+      when(deviceConnection.connect()).thenResolve(true) 
+      when(deviceServiceMock.getDevices()).thenResolve(devices)
+      when(factoryMock.createDeviceConnection(device)).thenReturn(instance(deviceConnection))
+      const testee: DeviceConnectionServiceImplementation = createTestee({deviceConnectionFactory: factoryMock, deviceService: deviceServiceMock })
+    
+      await testee.createConnection(device)
+      const devices2: Device[] = testee.getConnectedDevices()
+
+      expect(devices2.length).toBeGreaterThan(0)
```
KvelaGorrrrnio: A convention that is used in the codebase is to call the captured test value for `result`. I am not sure if it is used for all tests, so I won' state it as a requirement. If you want to have a name that describes the content, then you could call this `connectedDevices`. It would also make the assertion more descriptive :D

src/business-logic/services/test/device-connection-service-implementation.spec.ts
```diff
@@ -0,0 +1,309 @@
...
+        const devices: Device[] = [device]
+        const factoryMock: DeviceConnectionFactory = mock<DeviceConnectionFactory>()
+        const deviceConnection: INewsGatewayDeviceConnection = mock<INewsGatewayDeviceConnection>()
+        const deviceServiceMock: DeviceService = mock<DeviceService>()
+        when(deviceConnection.connect()).thenResolve(true) 
+        when(deviceServiceMock.getDevices()).thenResolve(devices)
+        when(factoryMock.createDeviceConnection(device)).thenReturn(instance(deviceConnection))
+        const testee: DeviceConnectionServiceImplementation = createTestee({deviceConnectionFactory: factoryMock, deviceService: deviceServiceMock })
+      
+        await expect(testee.removeConnectionById(device.id)).rejects.toThrow(`Device with ID '${device.id}' is already removed.`)
```
KvelaGorrrrnio: I would watch out for basing assertions on error messages, since it makes the test brittle. E.g. to conform with the rest of the codebase, `ID` should be written as `id`.  Or if we decide to change the name of Device to Gateway, then the functionality is intact, but the test fails. If you want to test for a specific failure scenario, then we can check for a specialized error/exception classes instance. E.g. like so:
```typescript
const result: Promise<void> = testee.removeConnectionById(device.id)

await expect(result).rejects.toThrow(DeviceAlreadyRemovedException)
```
A suggestion to keep the exception count a bit down is to use `DeviceNotFoundException` instead of `DeviceNotFoundException`. I know there is a slight semantic shift, but if I call this with some gibberish ID, I would still get that it is already removed, while it has never been there :)

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,128 @@
+import { WebSocket } from 'ws'
```
KvelaGorrrrnio, line 1: If we upgrade the node version to 22, then [WebSockets are supported natively](https://nodejs.org/en/blog/announcements/v22-release-announce#websocket) :)

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  public connectionExists(deviceId: string): boolean {
+    return this.connectedDevices.has(deviceId)
+  }
+
+  public getConnectedDevices(): Device[] {
+    return Array.from(this.connectedDevices.values()).map(deviceAggregate => deviceAggregate.device)
+  }
+
+  private async initialize(deviceService: DeviceService): Promise<void> {
```
KvelaGorrrrnio: Since this is only used in the constructor, it would be nice if it was placed right under the constructor.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,128 @@
...
+      this.isConnectingOrDisconnecting = true
+
+      try {
+        if (this.client.readyState === WebSocket.OPEN) {
+          this.client.close()
+        }
+    
+        if (this.pingTimeout !== null) {
+          clearTimeout(this.pingTimeout)
+        }
```
KvelaGorrrrnio: Just to make the code a bit simpler, it can be considered to remove the check. Calling clearTimeout on `null` or `undefined` has no effect.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,128 @@
...
+    })
+
+    this.client.on('close', (code, reason) => {
+      if (this.pingTimeout !== null) {
+        clearTimeout(this.pingTimeout)
+      }
+
+      if (INewsGatewayDeviceConnection.device.isConnected) {
+        INewsGatewayDeviceConnection.device.isConnected = false
+      }
```
KvelaGorrrrnio: Both branches of the conditional statement will result in the same state, so to make the code simpler, the condition can go, such that this becomes:
```typescript
INewsGatewayDeviceConnection.device.isConnected = false
```
And we should look at another way of informing about connection changes as you have already mentioned :)

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  private terminate(): void {
+    if (this.client.readyState === WebSocket.OPEN) {
+      this.client.terminate()
+    }  
+  }
+
+  public async disconnect(): Promise<boolean> {
+    return new Promise<boolean>((_resolve, reject) => {
```
KvelaGorrrrnio: You can remove the new Promise syntax. If you throw an error inside an async function, then it is treated as a promise rejection.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,128 @@
...
+        }
+    
+        if (this.pingTimeout !== null) {
+          clearTimeout(this.pingTimeout)
+        }      
+        INewsGatewayDeviceConnection.device.isConnected = false
+        return true
+      } catch (error) {
+        INewsGatewayDeviceConnection.device.isConnected = true
+        return false
```
KvelaGorrrrnio: Kind of the same as already discussed a bit about what to return for Promise based methods, but wouldn't it be more useful for the caller to get a rejected promise with the caught error, rather than a successful call that says that it wasn't successful?

.gitignore
```diff
@@ -12,3 +12,5 @@ dist/
 # Ignore VSCode specifics
 .vscode
 
+# Custom ingores
+**/*-integration-tests.spec.ts
```
KvelaGorrrrnio: Is this needed?

src/business-logic/device-factories/device-connection-factory.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Device } from '../../model/entities/device'
+import { DeviceConnection } from '../services/interfaces/device-connection'
+import { INewsGatewayDeviceConnection } from '../services/inews-gateway-connection-implementation'
+import { DeviceConnectionFactory } from '../services/interfaces/device-connection-factory'
+import { Logger } from '@tv2media/logger/*'
+import { LoggerFacade } from '../../logger/logger-facade'
+
+export class DeviceConnectionFactoryImplementation implements DeviceConnectionFactory {
+  constructor(private readonly logger: Logger){
+  }
```
RasmusAlbrektsen: Missing newline in between this constructor and the method below.

src/business-logic/device-factories/device-connection-factory.ts
```diff
@@ -0,0 +1,25 @@
+import { DeviceType } from '../../model/enums/device-type'
+import { Device } from '../../model/entities/device'
+import { DeviceConnection } from '../services/interfaces/device-connection'
+import { INewsGatewayDeviceConnection } from '../services/inews-gateway-connection-implementation'
+import { DeviceConnectionFactory } from '../services/interfaces/device-connection-factory'
+import { Logger } from '@tv2media/logger/*'
+import { LoggerFacade } from '../../logger/logger-facade'
+
+export class DeviceConnectionFactoryImplementation implements DeviceConnectionFactory {
```
RasmusAlbrektsen, line 9: Something about this naming convention seems a bit weird to me, but i do see why it is like this. It's just the `Implementation` part that seems a bit odd, but i understand that the Factory patterns usually would just call this class `DeviceConnectionFactory` and that the fact that it uses an Interface with the same name makes that weird too. I don't know what the naming convention should be here :-D

src/business-logic/facades/factory-facade.ts
```diff
@@ -0,0 +1,9 @@
+import { LoggerFacade } from '../../logger/logger-facade'
+import { DeviceConnectionFactoryImplementation } from '../device-factories/device-connection-factory'
+import { DeviceConnectionFactory } from '../services/interfaces/device-connection-factory'
+
+export class FactoryFacade {
+  public static createDeviceConnectionFactory(): DeviceConnectionFactory{
```
RasmusAlbrektsen: Personal preference (and i know this is a draft, but just in case), i like a space after the return type `DeviceConnectionFactory` and before the `{`.

src/business-logic/services/device-connection-service-implementation.ts
aardhuis: TODO: Implement this suggestion. (note to self)

src/business-logic/services/device-connection-service-implementation.ts
aardhuis: I was imprecise and said boolean when I meant numerical. Sloppy of me, I know.

Not sure if enums are the way to go if we want to divorce ourselves from numerical representations, but that's another topic.

I will follow your suggested standard.

src/business-logic/services/device-connection-service-implementation.ts
aardhuis: Oh...

.gitignore
aardhuis: Oops

src/model/entities/test/entity-test-factory.ts
aardhuis: For the sake of brevity and completing this sprint I will ignore this for now.

RasmusAlbrektsen (dismissed):
Great work!

src/business-logic/services/inews-gateway-connection-implementation.ts
aardhuis: For the sake of brevity and completing this sprint I will ignore this for now.

RasmusAlbrektsen (approved):

src/model/entities/test/entity-test-factory.ts
aardhuis: Aaaaand it turns out I cannot ignore this, since the queues params list is needed to connect.
Okay. To paraphrase King Henry during the siege of Harfleur:

"Once more unto the breach, dear friends, once more;
Or close the wall up with our developer dead."

src/business-logic/services/inews-gateway-connection-implementation.ts
KvelaGorrrrnio: A late errata: This is more for informational value than a decree that it should be done in this PR 😁

src/business-logic/device-factories/device-connection-factory.ts
```diff
@@ -0,0 +1,26 @@
...
+  constructor(private readonly logger: Logger){
+  }
+  
+  public createDeviceConnection(device: Device): DeviceConnection | undefined {
+    return this.mapDeviceToDeviceConnection(device)
+  }
+
+  public mapDeviceToDeviceConnection(device: Device): DeviceConnection | undefined {
```
LindvedKrvang: This method is only use in this class so it should be private

src/business-logic/device-factories/device-connection-factory.ts
```diff
@@ -0,0 +1,26 @@
...
+import { INewsGatewayDeviceConnection } from '../services/inews-gateway-connection-implementation'
+import { DeviceConnectionFactory } from '../services/interfaces/device-connection-factory'
+import { Logger } from '@tv2media/logger/*'
+import { LoggerFacade } from '../../logger/logger-facade'
+
+export class DeviceConnectionFactoryImplementation implements DeviceConnectionFactory {
+  constructor(private readonly logger: Logger){
+  }
+  
+  public createDeviceConnection(device: Device): DeviceConnection | undefined {
```
LindvedKrvang: Suggestion: Maybe we should throw an error if we try to connect to an unsupported Device? Then we can get rid of the undefined :)

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+
+type DeviceAggregate = {
+  deviceConnection: DeviceConnection,
+  device: Device
+}
+
+export class DeviceConnectionServiceImplementation implements DeviceConnectionService {
+  private readonly connectedDevices: Map<string, DeviceAggregate> = new Map()
+
+  constructor(private readonly deviceConnectionFactory: DeviceConnectionFactory, private readonly deviceService: DeviceService, private readonly logger: Logger) {
```
KvelaGorrrrnio: We have the convention that when a logger is used in a context, the context tells the logger who is using the logger. This is done with `logger.tag(this.constructor.name)` which creates a new logging context.
By tagging the log entries with the class that produces them, it makes it easier to see where a log originated.
The way we usually do so is by not binding the passed in logger instance via the parameters, but instead set it in the constructor body, like so:
```typescript
export class DeviceConnectionServiceImplementation implements DeviceConnectionService {

  constructor(/* other depenencies */, logger: Logger) {
    this.logger = logger.tag(this.constructor.name)
  }
}
```
This also means that the debug statements in the constructor should use `this.logger` instead of `logger`.

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    await device.deviceConnection.disconnect()
+    this.logger.info(`Device ${deviceId} disconnected`)
+
+
+    this.connectedDevices.delete(deviceId)
+    this.logger.info(`Device ${deviceId} removed`)
+
+
+    return
```
KvelaGorrrrnio: Can't this be removed? :)

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+    await Promise.all(devices.map(device => this.createConnection(device)))
+  }
+
+  public async createConnection(device: Device): Promise<void> {
+    if (this.connectedDevices.has(device.id)) {
+      throw new DeviceAlreadyConnectedException(device.id)
+    }
+    const deviceConnection: DeviceConnection | undefined = this.deviceConnectionFactory.createDeviceConnection(device)
+
+    if(typeof deviceConnection === 'undefined'){
```
LindvedKrvang: To check if something is `undefined` you can simply do `if (!deviceConnection)`

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+        logger.error(`Initialization of DeviceConnectionService failed with error: ${error}`)
+      })
+  }
+
+  private async initialize(deviceService: DeviceService): Promise<void> {
+    const devices: Device[] = await deviceService.getDevices()
+    await Promise.all(devices.map(device => this.createConnection(device)))
+  }
+
+  public async createConnection(device: Device): Promise<void> {
```
LindvedKrvang: This method is only called from within this class and from tests.
Since it's only called from within this class it should be private.
It looks like the only reason it's public is so we can call it from the tests which is an example of modifying production code solely becase of tests, which is a bad practice.

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (this.connectedDevices.has(device.id)) {
+      throw new DeviceAlreadyConnectedException(device.id)
+    }
+    const deviceConnection: DeviceConnection | undefined = this.deviceConnectionFactory.createDeviceConnection(device)
+
+    if(typeof deviceConnection === 'undefined'){
+      this.logger.debug(`createConnection exited due to unsupported device ${device}`)
```
LindvedKrvang: This debug method references the internal workings of `DeviceConnectionFactory`. Classes shouldn't know about the internals of other classes. 
`DeviceConnectionFactory` should log this method (which it kinda does).

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+      return 
+    }
+
+    await deviceConnection.connect()
+
+    this.connectedDevices.set(device.id, {deviceConnection, device})
+    this.logger.info(`Device ${device.id} of type ${device.type} connected`)
+
+    return
```
LindvedKrvang: Since this is the end of the method and the return type is `void` (`Promise<void>` in this case which gives the same result) then you don't need to include a return statement :)

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  public async send(deviceId: string, params: string[]): Promise<void> {
+    const connectedDevice: DeviceAggregate | undefined = this.connectedDevices.get(deviceId)
+    if(connectedDevice === undefined) {
```
LindvedKrvang: This could be written as `if(!connectedDevice)` if you want :)

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+      throw new DeviceNotFoundException(deviceId)
+    }
+
+    await connectedDevice.deviceConnection.send(deviceId, params)
+  }
+
+  public async listen(deviceId: string, _callback: (data: unknown) => void): Promise<void> {
```
LindvedKrvang: I believe it was mentioned in another comment, but `_` is used for unused variables and since `_callback` is used on line 68 we can omit the underscore from the name :)

src/business-logic/services/device-connection-service-implementation.ts
KvelaGorrrrnio: `typeof` is not needed here. The condition can check against the `undefined` literal like `deviceConnection === undefined`.
Since the alternative to this being undefined is being a truthy value, and `undefined` is falsy, the condition can also be expressed as `!deviceConnection`.

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  public getConnectionStatusById(deviceId: string): DeviceConnectionStatus {
+    // we don't do thorough testing of both actual connection and the isConnected field since isConnected is an inheritance from CoreDevice and not needed here.
+    return this.connectedDevices.has(deviceId) ? DeviceConnectionStatus.CONNECTED : DeviceConnectionStatus.DISCONNECTED 
+  }
+
+  public async disconnectConnectionById(deviceId: string): Promise<void> {
+    await this.connectedDevices.get(deviceId)?.deviceConnection.disconnect()
+    this.logger.info(`Device ${deviceId} disconnected`)
+    return
```
LindvedKrvang: No need to include a return statement at the end of void methods :)

src/business-logic/services/device-connection-service-implementation.ts
LindvedKrvang: No need to include a return statement at the end of void methods :)

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,123 @@
+import { WebSocket } from 'ws'
+import { Device } from '../../model/entities/device'
+import { DeviceConnection } from './interfaces/device-connection'
+import { DeviceType } from '../../model/enums/device-type'
+import { Logger } from '@tv2media/logger/*'
+
+export class INewsGatewayDeviceConnection implements DeviceConnection {
+  private static device: Device
```
LindvedKrvang: Why is this static?

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  public async listen(_deviceId: string, _callback: (data: unknown) => void): Promise<void> {
+    // For now: Simulate an async operation, e.g., network request
+    await new Promise(resolve => setTimeout(resolve, 1000))
+  }
+
+  public async connect(): Promise<boolean> {
+    return new Promise<boolean>((resolve, reject) => {
+      if (this.isConnectingOrDisconnecting) {
+        return reject(new Error('Device is already connecting or disconnecting.'))
```
LindvedKrvang: It would be nice if you would use one of our `Exception` classes instead of `Error` :)

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,123 @@
...
+  private readonly INEWS_GATEWAY_HOST: string = process.env.INEWS_GATEWAY_HOST ?? 'ws://localhost:3008'
+
+  private client: WebSocket
+  private pingTimeout: NodeJS.Timeout
+  private isConnectingOrDisconnecting: boolean = false
+
+  constructor(device: Device, private readonly logger: Logger) {
+    if(INewsGatewayDeviceConnection.device?.type === DeviceType.INEWS_GATEWAY){
+      return 
+    }
```
KvelaGorrrrnio: Since this is an implementation for handling device connection only for iNews Gateway devices, the constructor signature can be reduced from taking a generic device to taking a INewsGatewayDevice.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+      this.isConnectingOrDisconnecting = true
+
+      try {
+        if (this.client.readyState === WebSocket.OPEN) {
+          this.client.close()
+        }
+     
+        INewsGatewayDeviceConnection.device.isConnected = false
+        resolve
```
KvelaGorrrrnio: `resolve` is missing parenthesis. Since it is never called, the disconnect method will hang if the connection is closed.

src/model/entities/test/entity-test-factory.ts
```diff
@@ -76,6 +76,33 @@ export class EntityTestFactory {
     }
   }
 
+  public static createINewsGatewayDevice(device: Partial<INewsGatewayDevice> = {}): INewsGatewayDevice {
```
LindvedKrvang: You don't need the methods `createINewsGatewayDevice` and `createTelemetricsDevice`. You already have the `createDevice` method.
In order to get an INewsDevice in your tests simply write:
```
const iNewsDevice: INewsGatewayDevice = EntityTestFactory.createDevice({
  type: DeviceType.INEWS_GATEWAY,
  username: 'someUserName',
  password: 'somePassword'
})
```

KvelaGorrrrnio (approved):

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -0,0 +1,123 @@
...
+      clearTimeout(this.pingTimeout)
+    }
+
+    this.pingTimeout = setTimeout(() => {
+      this.terminate()
+    }, 30000 + 1000)
+  }
+
+  private terminate(): void {
+    if (this.client.readyState === WebSocket.OPEN) {
```
RasmusAlbrektsen: Shouldn't this be reversed? Otherwise we will always terminate the connection with the first ping :-)

`if (this.client.readyState !== Websocket.OPEN`)

src/business-logic/services/inews-gateway-connection-implementation.ts
RasmusAlbrektsen: Changed to the suggestion!

src/business-logic/services/inews-gateway-connection-implementation.ts
RasmusAlbrektsen: Now takes the specific device type instead.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: This is now `!deviceConnection`.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: Renmoved.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: Changed to follow the procedure everywhere else.

src/model/entities/test/entity-test-factory.ts
RasmusAlbrektsen: Now uses `createDevice` and i removed the, now, unused methods you commented on.

src/business-logic/services/inews-gateway-connection-implementation.ts
RasmusAlbrektsen: Uses expection class now.

src/business-logic/services/inews-gateway-connection-implementation.ts
RasmusAlbrektsen: No longer is, didn't have to be.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: No longer there.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: Gone.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: Gone.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: It is now.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: These returns are removed everywhere possible.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: No longer there.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: Changed to private, removed from interface and instead i re-arranged the init-flow to avoid the hard coupling with the test-file.

src/business-logic/services/device-connection-service-implementation.ts
RasmusAlbrektsen: Does so now.

src/business-logic/device-factories/device-connection-factory.ts
RasmusAlbrektsen: Does so now.

src/business-logic/device-factories/device-connection-factory.ts
RasmusAlbrektsen: Private now.

src/logger/dummy-logger.ts
```diff
@@ -0,0 +1,8 @@
+import  { Logger } from './logger'
+import { LoggerBase } from '@tv2media/logger'
+
+export class DummyLogger extends LoggerBase implements Logger {
```
LindvedKrvang, line 4: Do we need a dummy logger?
I can see you use it for tests.
Can't you just make a mock of it instead?

LindvedKrvang (approved):

KvelaGorrrrnio (approved):

PR #190, opened 2024-07-30
SOF- 2126: Wipe transitions cut primary feed too early

KvelaGorrrrnio:
Sofie server now tracks if a Piece was inserted on air to check if it should have applied the piece start delay.


src/model/entities/piece.ts
```diff
@@ -46,6 +47,7 @@ export class Piece {
   private duration?: number
   private executedAt: number
   private isUnsyncedPiece: boolean = false
+  private isPieceInsertedOnAir: boolean // TODO: Add getter and setter to make it better feta
```
RasmusAlbrektsen: Should probably remove this comment.

RasmusAlbrektsen (approved):
Down right amazing work, must say.

src/model/entities/test/entity-test-factory.ts
```diff
@@ -55,12 +56,20 @@ export class EntityTestFactory {
...
       name: 'pieceName',
-      duration: 420,
       start: 0,
       pieceLifespan: PieceLifespan.WITHIN_PART,
+      isPlanned: false,
+      preRollDuration: 0,
+      postRollDuration: 0,
+      transitionType: TransitionType.NO_TRANSITION,
+      timelineObjects: [],
+
```
MadsThy: whats this space? 😄

MadsThy (approved):

src/model/entities/test/entity-test-factory.ts
KvelaGorrrrnio: It is gone!

MadsThy (approved):

src/model/entities/piece.ts
```diff
@@ -135,6 +138,14 @@ export class Piece {
     return this.duration
   }
 
+  public isInsertedOnAir(): boolean {
+    return this.isPieceInsertedOnAir
+  }
+
+  public markAsInsertedOnAir(): void {
+    this.isPieceInsertedOnAir = true
+  }
```
LindvedKrvang, line 147: Only unplanned Pieces can be inserted, so it would be nice to have guard against marking planned Pieces as inserted on Air

PR #192, opened 2024-08-07
Sof 1966 able to reconnect

aardhuis:
Working on the reconnection strategy for the inews gateway


src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -28,9 +29,28 @@ export class DeviceConnectionServiceImplementation implements DeviceConnectionSe
...
+        try {
+          await this.createConnection(device)
+        } catch(error) {
+          this.logger.error(error)
+        }
+      }
+    }
   }
 
+  // Promise.all creates a race condition. Commented out until fixed.
```
LindvedKrvang, line 44: Are there plans to fix this race condition? If not, then this commented out code will just stay here indefinately, which mean we should delete the code now.

LindvedKrvang: Same comment for the commented out code further down.

src/business-logic/services/device-service-implementation.ts
```diff
@@ -29,4 +32,8 @@ export class DeviceServiceImplementation implements DeviceService {
     await this.deviceRepository.delete(deviceId)
     this.deviceEventEmitter.emitDeviceDeletedEvent(deviceId)
   }
+
+  public reconnect(deviceId: string): void {
+    this.deviceEventEmitter.emitDeviceReconnectingEvent(deviceId)
+  }
```
LindvedKrvang, line 38: This only emit that we have reconnect to the Device, but not attempts is being made to actually reconnect?

On the same note: It doesn't look like that we emit when we initial connect to a Device?

src/business-logic/services/fixed-interval-reconnect-strategy.ts
```diff
@@ -0,0 +1,49 @@
...
+  private readonly logger: Logger
+  private delayTimer?: NodeJS.Timeout
+
+  constructor(logger: Logger) {
+    this.logger = logger.tag('FixedIntervalReconnectStrategy')
+  }
+
+  public connected(): void {
+    return
+  }
```
LindvedKrvang, line 16: What's the purpose of this method?

src/business-logic/services/fixed-interval-reconnect-strategy.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  public disconnected(connect: () => void): void {
```
LindvedKrvang, line 18: `disconnected` sounds like the boolean value that would be true when we are disconnected, not a function to be executed.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
+import { DeviceEvent, DeviceReconnectingEvent } from '../../presentation/value-objects/device-event'
+import { DeviceEventType } from '../../presentation/enums/event-type'
 
 export class INewsGatewayDeviceConnection implements DeviceConnection {
-  private static device: Device
-  private readonly INEWS_GATEWAY_HOST: string = process.env.INEWS_GATEWAY_HOST ?? 'ws://localhost:3008'
-
-  private client: WebSocket
-  private pingTimeout: NodeJS.Timeout
+  private static instance: INewsGatewayDeviceConnection | null = null
```
LindvedKrvang, line 13: Please use `undefined` instead of `null`. They do not behave the same.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
+  private readonly reconnectStrategy: FixedIntervalReconnectStrategy
+  private readonly deviceEventObserver: DeviceEventObserver = EventEmitterFacade.createDeviceEventObserver()
+  private client: WebSocket | null = null
+  private pingTimeout: NodeJS.Timeout | null = null
```
LindvedKrvang, line 17: Same comment about `undefined` vs `null`

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
-      return 
+  private constructor(private readonly device: Device, private readonly logger: Logger) {
+    if (device?.type !== DeviceType.INEWS_GATEWAY) {
+      throw new Error('Invalid device type')
     }
 
-    INewsGatewayDeviceConnection.device = device
-  }
+    if ((device as INewsGatewayDevice).queues === undefined) {
+      throw new Error('iNewsGatewayDevice not configured correctly. Remember to set up host, port and queues to listen to.')
```
LindvedKrvang, line 26: Please use one of our `Exceptions` instead of `Error`.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
+    }
 
-  public async send(_deviceId: string, _params: string[]): Promise<void> {
-    this.client.send(JSON.stringify(_params[0]))
-    // For now: Simulate an async operation, e.g., network request
-    await new Promise(resolve => setTimeout(resolve, 1000))
+    this.reconnectStrategy = new FixedIntervalReconnectStrategy(this.logger)
```
LindvedKrvang, line 29: Classes shouldn't maintain their own dependencies. Instead dependencies should be injected into them.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
 
-  public async listen(_deviceId: string, _callback: (data: unknown) => void): Promise<void> {
-    // For now: Simulate an async operation, e.g., network request
-    await new Promise(resolve => setTimeout(resolve, 1000))
+  public static getInstance(device: Device, logger: Logger): INewsGatewayDeviceConnection {
+    if (!INewsGatewayDeviceConnection.instance) {
+      INewsGatewayDeviceConnection.instance = new INewsGatewayDeviceConnection(device, logger)
+    }
+    return INewsGatewayDeviceConnection.instance
   }
```
LindvedKrvang, line 38: Would be nice if all static things were places above all non-static things. It makes it easier to get an overview of all the static things in the class.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
 
   public async connect(): Promise<boolean> {
+    if (this.isConnectingOrDisconnecting) {
+      return Promise.reject(new Error('Already in a connecting or disconnecting state'))
+    }
+
+    if (this.client && this.client.readyState === WebSocket.OPEN) {
+      return Promise.reject(new Error('Already connected to iNews Gateway'))
```
LindvedKrvang, line 46: Please use one of our `Exception` classes instead of `Error`.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
+    const queuesString: string = iDevice.queues.join(',')
+    const encodedQueues: string = encodeURIComponent(queuesString)
+    return `ws://${iDevice.host}:${iDevice.port}?queues=${encodedQueues}`
+  }
+
+  private setupClientEventHandlers(): void {
+    this.client?.on('error', (error) => {
       this.logger.error('WebSocket error:', error)
-      INewsGatewayDeviceConnection.device.isConnected = false
+      INewsGatewayDeviceConnection.instance!.device.isConnected = false
```
LindvedKrvang, line 100: We are in the `INewsGatewayDeviceConnnection` which mean we _are_ the `instance`. Why are we trying to access ourself?
`this.device.isConnected = false` should do the exact same.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -88,36 +160,52 @@ export class INewsGatewayDeviceConnection implements DeviceConnection {
 
     this.pingTimeout = setTimeout(() => {
       this.terminate()
-    }, 30000 + 1000)
+    }, 300_000)
```
LindvedKrvang, line 163: This is a magic number. It would be nice if it was extracted into a constant value

aardhuis: I want to have different devices store their timeout / reconnect values with the device. Does that make sense? i.e. Device.Timeout = 300_000, or, to follow your logic: 

const INEWS_TIMEOUT_VALUE: number = 300_000
iNewsDevice.Timeout = device.timeout ?? INEWS_TIMEOUT_VALUE

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
+  private initializeClientEvents(iDevice: INewsGatewayDevice): void {
+    if (!this.validateDeviceParameters(iDevice)) {
+      throw new Error('Invalid device parameters')
+    }
+
+    const url: string = this.buildWebSocketUrl(iDevice)
+
+    this.client = new WebSocket(url)
+
+    if (this.client) {
```
KvelaGorrrrnio, line 81: The check is redundant. The client has just been set, so the true-branch will always be triggered.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
       }
     })
   }
 
-  private initializeClientEvents(): boolean {
-    this.client.on('error', (error) => {
+  private initializeClientEvents(iDevice: INewsGatewayDevice): void {
+    if (!this.validateDeviceParameters(iDevice)) {
+      throw new Error('Invalid device parameters')
+    }
```
KvelaGorrrrnio, line 75: It would be nice if the `iDevice` was spelled out :)
The `validateDeviceParameters` checks for the existence of the attributes, but from the type of `iDevice` these attributes are required, so the validation is unnecessary here.

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -88,36 +160,52 @@ export class INewsGatewayDeviceConnection implements DeviceConnection {
...
-      } finally {
-        this.isConnectingOrDisconnecting = false
+        this.client.close()
       }
-    })
+
+      INewsGatewayDeviceConnection.instance!.device.isConnected = false
+      return true
+    } catch (error) {
+      INewsGatewayDeviceConnection.instance!.device.isConnected = true
```
KvelaGorrrrnio, line 191: Are we sure that the we are still connected in this case?
Instead of making the assumption, the readyState could be checked either here in the catch clause or in the finally clause.

src/presentation/dtos/device-dto.ts
```diff
@@ -22,20 +22,26 @@ export abstract class DeviceDto {
...
+  host: string
+  port: number
+  queues: string[]
 }
 
 export class INewsDeviceDto extends DeviceDto implements INewsDeviceDtoInterface {
   public readonly type: DeviceType.INEWS_GATEWAY
-  public readonly username: string
-  public readonly password: string
+  public readonly password: string | undefined
```
KvelaGorrrrnio, line 33: We can remove the `password` for now, since it is not used. When we introduce authentication to the iNews Gateway it should re-introduced. Until then it is just dead code :)

src/business-logic/services/inews-gateway-connection-implementation.ts
LindvedKrvang: With different devices do you mean that an iNews and Atem device have different timeouts or that you could configure two iNews devices that have different timeouts?

If it's the first, then the timeout is already device specific because you are in `INewsGatewayConnectionImplementation` and then I would do:
```
const TIMEOUT_DURATION: number = 300_000

export class INewsGatewayConnectionImplementation {
  ...

  private heartbeat(): void {
    this.pingTimeout = setTimeout(() => {...}, TIMEOUT_DURATION)
  }
}
```

If it's the latter, then I ask: Is that something we need?

src/business-logic/services/device-connection-service-implementation.ts
```diff
@@ -28,9 +29,28 @@ export class DeviceConnectionServiceImplementation implements DeviceConnectionSe
 
   private async initialize(deviceService: DeviceService): Promise<void> {
     const devices: Device[] = await deviceService.getDevices()
-    await Promise.all(devices.map(device => this.createConnection(device)))
+    
+    for (const device of devices) {
+      if (!this.connectionExists(device.type)) {
```
KvelaGorrrrnio, line 34: Shouldn't the connection existence be checked against the device id?

src/business-logic/services/fixed-interval-reconnect-strategy.ts
KvelaGorrrrnio: To conform to the ReconnectStrategy interface. Using a fixed interval does not keep state about previous connect attempts because the behaviour is the same each time.

src/business-logic/services/fixed-interval-reconnect-strategy.ts
KvelaGorrrrnio: By our boolean naming convention, that would be `isDisconnected`. The method tells the strategy about a disconnect event. I don't have any names that don't become kinda awkward (e.g. `informAboutDisconnection`).

src/business-logic/services/inews-gateway-connection-implementation.ts
KvelaGorrrrnio, line 1: There is a mismatch between the filename and the class name :)

src/business-logic/services/inews-gateway-connection-implementation.ts
KvelaGorrrrnio, line 13: Why is this made into a singleton?

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -1,84 +1,156 @@
...
+  private readonly reconnectStrategy: FixedIntervalReconnectStrategy
+  private readonly deviceEventObserver: DeviceEventObserver = EventEmitterFacade.createDeviceEventObserver()
+  private client: WebSocket | null = null
+  private pingTimeout: NodeJS.Timeout | null = null
   private isConnectingOrDisconnecting: boolean = false
 
-  constructor(device: Device, private readonly logger: Logger) {
-    if(INewsGatewayDeviceConnection.device?.type === DeviceType.INEWS_GATEWAY){
-      return 
+  private constructor(private readonly device: Device, private readonly logger: Logger) {
```
KvelaGorrrrnio, line 20: Since this a DeviceConnection for the `INewsGatewayDevice`, it would make sense that it expects an INewsGatewayDevice instead of a Device. It will also get rid of the assertions. The assertions on the data structure might need to be performed elsewhere though (e.g. when this class is constructed).

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -88,36 +160,52 @@ export class INewsGatewayDeviceConnection implements DeviceConnection {
...
 
   public async disconnect(): Promise<boolean> {
-    return new Promise<boolean>((resolve, reject) => { 
-      if (this.isConnectingOrDisconnecting) {
-        return reject(new Error('Device is already connecting or disconnecting.'))
-      }
-      this.isConnectingOrDisconnecting = true
+    if (this.isConnectingOrDisconnecting) {
+      return Promise.reject(new Error('Device is already connecting or disconnecting.'))
+    }
```
KvelaGorrrrnio, line 176: Since this is in an async method, this can be reduced to:
```typescript
    if (this.isConnectingOrDisconnecting) {
      throw new Error('Device is already connecting or disconnecting.')
    }
```

src/business-logic/services/inews-gateway-connection-implementation.ts
```diff
@@ -88,36 +160,52 @@ export class INewsGatewayDeviceConnection implements DeviceConnection {
...
-      try {
-        if (this.client.readyState === WebSocket.OPEN) {
-          this.client.close()
+    this.isConnectingOrDisconnecting = true
+
+    try {
+      if (this.client?.readyState === WebSocket.OPEN) {
+        if (this.pingTimeout !== null) {
+          clearTimeout(this.pingTimeout)
         }
```
KvelaGorrrrnio, line 184: Since calling `clearTimeout` on a null or undefined value has no effect, the condition can safely be removed.

src/model/entities/test/entity-test-factory.ts
```diff
@@ -84,8 +84,10 @@ export class EntityTestFactory {
       statusMessage: '',
       isConnected: false,
       type: DeviceType.INEWS_GATEWAY,
-      username: 'testuser',
       password: 'testpassword',
+      host: 'ws://localhost',
```
KvelaGorrrrnio, line 88: Wouldn't the host just be `localhost`?

src/presentation/controllers/device-controller.ts
```diff
@@ -101,4 +101,17 @@ export class DeviceController extends BaseController {
...
   }
+
+  @PostRequest('/:deviceId/reconnect')
+  public reconnectDevice(request: Request, response: Response): void {
+    try {
+      const deviceId: string = request.params.deviceId
+
+      this.deviceService.reconnect(deviceId)
+      
+      response.send(this.httpResponseFormatter.formatSuccessResponse({ message: 'Device reconnected successfully.' }))
```
KvelaGorrrrnio, line 112: I guess this is somewhat a asynchronous vs synchronous/ monolithic vs distributed mindset, but at this moment, when we respond to the client, can guarantee that the device has reconnected successfully?

src/business-logic/device-factories/device-connection-factory.ts
```diff
@@ -17,7 +17,7 @@ export class DeviceConnectionFactoryImplementation implements DeviceConnectionFa
   public mapDeviceToDeviceConnection(device: Device): DeviceConnection | undefined {
     switch (device.type) {
       case DeviceType.INEWS_GATEWAY:
-        return new INewsGatewayDeviceConnection(device, LoggerFacade.createLogger())
+        return  INewsGatewayDeviceConnection.getInstance(device, LoggerFacade.createLogger())
```
MadsThy, line 20: I suspect a linter hasn't caught this "double space" 😄

src/business-logic/services/device-connection-service-implementation.ts
MadsThy: Without knowing the area, would it be an idea to use Promise.allSettled to avoid the race condition? :)

PR #194, opened 2024-08-20
SOF-2119 Introduce Atem Preview ME. 

LindvedKrvang:
Introduce Atem Preview ME.

In order for the library `atem-state` to update the Preview in Atem it needs to have the ME object `{ programInput: number, previewInput?: number }`. In order for `atem-state` to make a transition it needs the object: `{ input: number, transition: AtemTransition, transitionSettings: AtemTransitionSettings }`. Due to how `atem-state` is implemented (and how VideoMixers works), the Preview won't be updated if we provide: `{ input: number, previewInput: number }`.

We now differentiate between the two above objects depending on if we need a transition or not. This also means that when trying to find the program source, it could either be `me.programInput` or `me.input` instead of just `me.input` as it was before.

Note: We don't set `previewInput` here. `previewInput` is being set by TSR.


RasmusAlbrektsen (approved):

src/blueprints/tv2/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
```diff
@@ -285,12 +290,17 @@ export class Tv2AtemVideoMixerTimelineObjectFactory implements Tv2VideoMixerTime
...
     }
 
-    const atemMeTimelineObject: AtemMeTimelineObject = blueprintTimelineObject as AtemMeTimelineObject
-    return  atemMeTimelineObject.content.me.input
+    const atemMeTimelineObject: AtemMixEffectTimelineObject = blueprintTimelineObject as AtemMixEffectTimelineObject
+    if (atemMeTimelineObject.content.me.type === AtemMixEffectType.TRANSITION) {
+      return atemMeTimelineObject.content.me.input
+    } else {
+      return atemMeTimelineObject.content.me.programInput
+    }
```
MadsThy: Just a pet peeve of mine 😄 You can do without the else, since you have a return in the if.

```suggestion
    }
    return atemMeTimelineObject.content.me.programInput
```

MadsThy (approved):
Looks great :)

src/blueprints/tv2/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
LindvedKrvang: It has been updated

PR #195, opened 2024-08-22
SOF-2156: Audio bed actions

KvelaGorrrrnio:
- **Creates audio bed actions from action manifests**
The primary outcome of the PR is that audio bed actions are created from action manifests. `Tv2ActionManifestData` is updated with an audio bed variant, that captures the necessary attributes. The audio bed configurations from Sofie core has been exposed as `AudioBedConfiguration`.

- **Adds Yarn build-watch script**
The `yarn build-watch` script watches the source code and build on changes with no code emission. It is added for convenience to check for code compilation.

- **Renames Tv2AudioTimelineObjectFactory to Tv2AudioMixerTimelineObjectFactory**
The responsibility of the factory is to generate timeline objects for controlling an audio mixer – not for e.g. playing out an audio file.

- **Creation of Fade audio bed timeline objects moved**
`createFadeAudioBedTimelineObject` method is moved from `Tv2VideoClipTimelineObjectFactory` to `Tv2AudioBedTimelineObjectFactory`, since it has nothing to do with video clips.

- **TestEntityFactory provides full default entity values**
To get rid of `as <EntityInterface>`, default values have been specified for all required fields.

- **Misc renaming**
E.g. `ActionFactory.ts` is renamed to `action-factory.ts` for consistency.


src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
```diff
@@ -308,6 +317,52 @@ export class Tv2CasparCgTimelineObjectFactory implements Tv2GraphicsElementTimel
...
+          },
+          outTransition: {
+            type: CasparCgTransitionType.MIX,
+            easing: CasparCgTransitionEase.LINEAR,
+            direction: CasparCgTransitionDirection.LEFT,
+            duration: this.convertFramesToMilliseconds(audioBedConfiguration.fadeOutDurationInFrames),
+          },
+        }
+      },
+      classes: ['lyd_on_air'],
```
LindvedKrvang: A thing I have been trying to do while moving stuff from Blueprints is to see how many of these `classes` we can remove.
The sole purpose of `classes` on TimelineObjects is to help with figuring out when to put the TimelineObject on Air (just like `priority` and `enable`).
It could for example look like this:
```
cons timelineObjectA = {
  id: 'a',
  enable: {
    start: 0,
    end: 1000
  },
  classes: ['some_class'],
  content: ...
}

const timelineObjectB = {
  id: 'b',
  enable: {
    start: '.some_class.end'
  },
  content: ...
}
```
The above snippet makes `timelineObjectB` start when `timelineObjectA` ends.

Looking through old Blueprints it looks like that `lyd_on_air` is only to stop the baseline audio object while we are playing an audio bed. 

Would be nice if we could test if this `lyd_on_air` is needed :) 

Note: It's not that the concept of `classes` is bad. Looking through old Blueprints it just haven't been utilized in a good way. (We do use `classes` for some things in the new Blueprints)

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
```diff
@@ -308,6 +317,52 @@ export class Tv2CasparCgTimelineObjectFactory implements Tv2GraphicsElementTimel
...
+    }
+  }
+
+  private convertFramesToMilliseconds(frames: number): number {
+    return 1000 * frames / FRAME_RATE
+  }
```
LindvedKrvang, line 364: We should come up with a way so we don't have to have this helper function in all our TimelineObjectFactories

src/model/entities/test/entity-test-factory.ts
```diff
@@ -91,4 +130,12 @@ export class EntityTestFactory {
       ...statusMessage
     }
   }
+
+  public static createActionManifest<ActionManifestData>(actionManifest: Partial<ActionManifest> & { data: ActionManifestData }): ActionManifest<ActionManifestData> {
+    return {
+      actionId: `action-manifest-${process.hrtime.bigint()}`,
```
LindvedKrvang, line 136: Is this to give it a random id?
If yes, I think there could be a point in streamlining how we do that since we do it in different ways elsewhere.

LindvedKrvang (approved):
Solid work

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
```diff
@@ -23,6 +24,7 @@ interface CoreShowStyleBlueprintConfiguration {
   BreakerConfig: CoreBreaker[]
   Transitions: Transition[]
   ShowstyleTransition: string
+  LYDConfig: CoreAudioBedConfiguration[]
```
MadsThy, line 27: is this some danish sneaking in? 😄 or is it an actual abbreviation?

src/model/entities/test/entity-test-factory.ts
MadsThy: I think, if we can agree that this is a good way to generate unique ID's, then let's make a task for replacing other methods with this one, for later :)

MadsThy (approved):

RasmusAlbrektsen (approved):

src/blueprints/tv2/helpers/tv2-show-style-blueprint-configuration-mapper.ts
KvelaGorrrrnio: It is danish sneaking in from the old code base. So we will have to live with it until we control ingest.

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
KvelaGorrrrnio: A very good point. I also had the thought of whether it is necessary or not. I have created a TODO on that line to check if it is necessary.

src/model/entities/test/entity-test-factory.ts
KvelaGorrrrnio: It is :) And a very good point indeed. I have created a task in Jira for doing this.

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
KvelaGorrrrnio: I have created a separate PR for it, to make the changes easier to review: https://github.com/tv2/sofie-server/pull/197

RasmusAlbrektsen (approved):

PR #196, opened 2024-08-23
SOF-2165 No longer blindly put all of the old Parts into the new Segment if the Segment is on Air

LindvedKrvang:
Until now, we have just blindly put all the Parts of the old Segment onto the new Segment, if the Segment was on Air while it was being updated. This made it so all updated Pieces and TimelineObjects of the new Parts were overridden and we wouldn't get Parts etc of the onAir Segment updated at all.

We no longer put the old Parts onto the new Segment.


RasmusAlbrektsen (approved):

src/model/entities/rundown.ts
```diff
@@ -590,9 +590,14 @@ export class Rundown extends BasicRundown {
...
     const oldSegment: Segment = this.segments[segmentIndex]
     if (oldSegment.isOnAir()) {
-      segment.setParts(oldSegment.getParts())
+      const newOnAirPart: Part | undefined = segment.getParts().find(part => part.isOnAir())
+      if (newOnAirPart) {
+        this.activeCursor = this.createCursor(this.activeCursor, { part: newOnAirPart })
+      }
+
       segment.putOnAir()
       this.activeCursor = this.createCursor(this.activeCursor, { segment })
```
MadsThy, line 599: Im not sure i understand how it can be ok to set activeCursor twice in the same function, but I trust that you know that this is ok :)

MadsThy (approved):

src/model/entities/rundown.ts
KvelaGorrrrnio: The introduced activeCursor assignment only changes the active part, while the next one on line 599, one only updates the segment.
One could refactor it, such that the introduced code gets the on air part to use, such that there is only one assignment.

src/model/entities/rundown.ts
```diff
@@ -590,9 +590,14 @@ export class Rundown extends BasicRundown {
...
+      this.updateInfinitePieces()
```
KvelaGorrrrnio, line 600: I assume that it has already been thought through, but won't this create logic branches where the on air content can get updated by ingest changes?

KvelaGorrrrnio (approved):
Looks good 🚀

src/model/entities/rundown.ts
LindvedKrvang: Potentially yes, but that is what Jacob wants for KLAR_ON_AIR

PR #197, opened 2024-08-26
refactor: Centralizes frame time conversion in the FrameTimeConverter class.

KvelaGorrrrnio:
_**Note:** Is based on #195._

Centralizes conversion from frames to milliseconds into the `FrameTimeConverter` class.


RasmusAlbrektsen (dismissed):

RasmusAlbrektsen (approved):

MadsThy (approved):

src/blueprints/tv2/helpers/frame-time-converter.ts
```diff
@@ -0,0 +1,7 @@
+export class FrameTimeConverter {
```
LindvedKrvang, line 1: To follow our pattern of dependency conversion, this should have an interface

PR #199, opened 2024-08-27
Sof 2164/planned transitions are not overridden by transition actions

LindvedKrvang:
If we tried to insert a TransitionEffectAction into a Part that already had a planned Transition, if the planned Transition is longer than the unplanned, then the planned Transition would still execute after the unplanned Transition. This is no longer the case.

We know insert another TimelineObject, alongside the Transition TimelineObject, that 'overrules' the original TimelineObject with the planned Transition. This is necessary since Atem changes source and executes Transitions as one. So we need something after our unplanned Transition ended to tell Atem not to do planned Transition.

Also included in this PR:
When inserting a Piece into a Part. If the Part is OnAir, then we stop previous Pieces on the same layer as the Inserted Piece. If the Part is not OnAir, then we replace previous Pieces on the same layer on the Part.
The reason being, if the Part is OnAir, then we need to save, that the Piece has already executed. If the Part is not OnAir, then the Piece has not been executed and never will be, because the Inserted Piece overrides it.


RasmusAlbrektsen (approved):

src/model/entities/part.ts
```diff
@@ -203,10 +203,38 @@ export class Part {
...
   }
 
+  private stopPiecesOnLayer(layer: string): void {
+    const existingPiecesOnLayer: Piece[] = this.pieces.filter(piece => piece.layer === layer)
+    existingPiecesOnLayer.forEach(piece => {
+      if (piece.getDuration()) {
+        return
+      }
+      piece.stop()
+    })
```
KvelaGorrrrnio: Suggestion The `piece.getDuration()` check can be put into the filter and the temporary variable can be removed. E.g.:
```typescript
this.pieces
  .filter(piece => piece.layer === layer && piece.getDuration())
  .forEach(stoppablePieceOnLayer => stoppablePieceOnLayer.stop())

File: src/model/entities/part.ts [hunk 2]
```diff
@@ -203,10 +203,38 @@ export class Part {
...
+      return
+    }
+    const removedPieces: Piece[] = this.pieces.splice(indexOfExistingPieceOnLayer, 1)
+    removedPieces.forEach(piece => {
+      if (!piece.isPlanned) {
+        return
+      }
+      this.replacedPlannedPieces.push(piece)
+    })
+  }
```
KvelaGorrrrnio, line 231: I was a bit confused about the types and names in this. The method name states that a single piece is removed, but the `removedPieces` and its type says that potentially multiple are removed.
This got me thinking about whether we only want to remove a single piece, all pieces on a layer or why we want to do this. The use case for this PR is replacing all overlapping transitions in the part, but one could imagine a scenario where we want to keep some future pieces on that layer, e.g.  a bundt or audio effect. Under these assumptions the pieces that needs to be removed are the ones that overlap with the new piece.
So something like the snippet below should give the desired effect.
```typescript
class Part {
  private removeOverlappingPiecesOnSameLayer(referencePiece: Piece): void {
    const piecesToRemove: Piece[] = this.pieces.filter(piece => piece.layer === referencePiece.layer && piece.doesOverlapWith(referencePiece))
    this.replacedPlannedPieces.push(...piecesToRemove)

    const pieceIdsToRemove: string[] = piecesToRemove.map(piece => piece.id)
    this.setPieces(this.pieces.filter(piece => !pieceIdsToRemove.includes(piece.id)))
  }
}
```

KvelaGorrrrnio (approved):
Looks good :rocket:
As we talked about, we are currently having side effects by updating the timeline objects on the piece.
For future reference, when we move blueprints into Alba, then it should be considered if the transition timeline objects should be on the transition piece. It will remove the need for the side effects :)

MadsThy (approved):

src/model/entities/part.ts
LindvedKrvang: I'm inclined to say YAGNI for now. We haven't run into such a scenario yet (to my knowledge).

Sidenote: In your suggested code you are putting _all_ removed Pieces into the `replacedPlannedPieces` array. Only planned Pieces are allowed to be put in there.

src/model/entities/part.ts
LindvedKrvang: Updated

PR #203, opened 2024-08-29
SOF 2160 - Adding to an Empty Active Rundown Doesnt Give a Next Cursor

RasmusAlbrektsen:
This PR fixes an issue that was encountered in the scenario where we have an active rundown and want to switch out the content entirely. Once a rundown was active and had no content, if we then added something back, we would not set a next cursor. We would also have a stale reference to the old next cursor that pointed to the now deleted part from earlier. 

The solution ensures that we unmark and clear the next cursor before setting it again. Also we now set the earliest available part in the Rundown as next, so long as nothing is on air currently in the active rundown.


src/model/entities/rundown.ts
```diff
@@ -150,8 +153,15 @@ export class Rundown extends BasicRundown {
...
-
     if (!this.activeCursor) {
+      try {
+        this.unmarkNextSegment()
+        this.setFirstSegmentAndPartNextCursor()
+      } catch (exception) {
+        if (!(exception instanceof NotFoundException)) {
+          throw exception
+        }
+      }
```
LindvedKrvang: This seems like some good logic to have tests for ;) 
I'm available if you want a hand in writing the tests.

MadsThy (approved):

src/model/entities/rundown.ts
```diff
@@ -1,30 +1,30 @@
...
+import {AlreadyExistException} from '../exceptions/already-exist-exception'
+import {LastSegmentInRundownException} from '../exceptions/last-segment-in-rundown-exception'
+import {NoPartInHistoryException} from '../exceptions/no-part-in-history-exception'
+import {OnAirException} from '../exceptions/on-air-exception'
+import {RundownTiming} from '../value-objects/rundown-timing'
+import {InTransition} from '../value-objects/in-transition'
+import {RundownMode} from '../enums/rundown-mode'
+import {AlreadyRehearsalException} from '../exceptions/already-rehearsal-exception'
+import {InvalidSegmentException} from '../exceptions/invalid-segment-exception'
+import {InvalidPartException} from '../exceptions/invalid-part-exception'
```
KvelaGorrrrnio: This seems to use the wrong spacing :D

KvelaGorrrrnio (approved):

src/model/entities/rundown.ts
RasmusAlbrektsen: Well i'll be damned if you are not correct sir. I added tests for removing the last segment and thus setting the next cursor to undefined since we have nothing left.
There is also now a test for when we add back a single segment to an empty active rundown, in which case it should be the one set as next.

src/model/entities/rundown.ts
RasmusAlbrektsen: I spaced it up good.

KvelaGorrrrnio (approved):
The spacing for imports in the test file is a bit weird, but otherwise looks good :+1:

PR #212, opened 2024-09-09
SOF-2204: Expose planned replaced pieces on part and when inserting pieces into part marked as next.

KvelaGorrrrnio:
Since inserting a piece into the part marked as next can remove another piece, a single piece inserted event is not enough information for clients to assemble the correct state. Therefore replaced planned pieces are now exposed on the PartDto and when a piece is inserted into the next part, a part updated event is emitted.
An alternative solution would be to see if a piece was removed and emit a replaced piece event instead. The part updated event was chosen due to its simplicity.

The PR also adds a Promise.catch to the auto next take, to avoid uncaught promise rejections, and logs the error.

The PR also removes the AUTO_NEXT_STARTED event, since no clients are using it.


MadsThy (approved):

LindvedKrvang (approved):
Not sure how I feel about the removal of the AUTO_NEXT_STARTED event, but I guess it's YAGNI sooo... delete away ;)

RasmusAlbrektsen (approved):

PR #214, opened 2024-09-10
SOF-2212 IngestUpdateEvents creates entities if missing

LindvedKrvang:
There is an extremely low chance that an IngestCreateEvent is missed. If it is missed, then following IngestUpdateEvents won't have an entity to update and we run into a "stalemate" untill a new IngestCreateEvent is received (either delete Rundown and reingest or float/unfloat the Segment).

This PR introduces that if we receive an IngestUpdateEvent for a Part or Segment that doesn't exist, we then create the Part or Segment instead.


src/business-logic/services/ingest-data-changed-service.ts
```diff
@@ -29,6 +29,20 @@ import { ActionEventEmitter } from './interfaces/action-event-emitter'
 
 const BULK_EXECUTION_TIMESPAN_IN_MS: number = 500
 
+const enum IngestEventPriority {
```
KvelaGorrrrnio: Is the const intentional? :)

src/business-logic/services/ingest-data-changed-service.ts
```diff
@@ -433,6 +453,14 @@ export class IngestDataChangedService implements DataChangeService {
     await this.persistRundown(rundown)
   }
 
+  private async fetchSegmentIfExist(segmentId: string): Promise<Segment | undefined> {
+    try {
+      return this.segmentRepository.getSegment(segmentId)
+    } catch (error) {
+      return Promise.resolve(undefined)
+    }
```
KvelaGorrrrnio: Suggestion: Use the promise syntax to make it a bit cleaner:
```typescript
    return this.segmentRepository.getSegment(segmentId).catch(() => undefined)
```

src/business-logic/services/ingest-data-changed-service.ts
```diff
@@ -433,6 +453,14 @@ export class IngestDataChangedService implements DataChangeService {
     await this.persistRundown(rundown)
   }
 
+  private async fetchSegmentIfExist(segmentId: string): Promise<Segment | undefined> {
```
KvelaGorrrrnio: Does fetch have a different meaning than get? If not, I suggest that we use `getSegmentIfExists`.

KvelaGorrrrnio (approved):
Looks good. A few suggestions :)

src/business-logic/services/ingest-data-changed-service.ts
LindvedKrvang: Force of habit. It has been removed

src/business-logic/services/ingest-data-changed-service.ts
LindvedKrvang: I like it. Updated to use it and removed the other methods.

src/business-logic/services/ingest-data-changed-service.ts
LindvedKrvang: Not really. But the methods has been removed in favour of the promise syntax

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

PR #216, opened 2024-09-11
Sof 2213/disallow source change during transition

LindvedKrvang:
Introduce `blockTakeDuration` on `InTransition`. This attribute already exists in Core/Blueprints and is used to determine whether or not we are currently doing a Take. (This also means no Blueprints updates :) )

If `Date.now() - part.executedAt` is smaller than ` part.inTransition.blockTakeDuration` that means that we are currently doing a Transition. So if we try to do a Take or insert a Part on air in this window, we throw an exception.

Our TransitionActionFactory has been updated to set the `blokTakeDuration` so the Transitions from our TransitionActions are blocked also.


src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -421,8 +423,18 @@ export class Tv2TransitionEffectActionFactory extends ActionFactory {
     ]
   }
 
+  private createPartInTransitionForEffect(durationInFrames: number): InTransition {
+    const durationInMilliSeconds: number = this.frameTimeConverter.convertFramesToMilliseconds(durationInFrames)
+    return {
+      blockTakeDuration: durationInMilliSeconds,
```
KvelaGorrrrnio: Minor detail: `Milliseconds` is one word 😁

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -386,6 +389,194 @@ describe(RundownTimelineService.name, () => {
 
       verify(ingestService.reloadIngestData(rundown.id)).never()
     })
+
+    describe('no Part has been Taken yet', () => {
+      it('does a Take', async () => {
+        // const rundownMock: Rundown = EntityMockFactory.createRundownMock()
```
KvelaGorrrrnio: I guess this can be removed?

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -386,6 +389,194 @@ describe(RundownTimelineService.name, () => {
...
+        const rundown: Rundown = instance(rundownMock)
+
+        const rundownRepository: RundownRepository = mock<RundownRepository>()
+        when(rundownRepository.getRundown(rundown.id)).thenReturn(Promise.resolve(rundown))
+
+        const testee: RundownTimelineService = createTestee({rundownRepository})
+
+        try {
+          await testee.takeNext(rundown.id)
+        } catch (error) {
```
KvelaGorrrrnio: Since we don't use the caught error, then we can omit it like so:
```typescript
try {
  await testee.takeNext(rundown.id)
} catch {
   // Comment
}
 ```

File: src/model/value-objects/in-transition.ts [hunk 4]
```diff
@@ -1,4 +1,5 @@
 export interface InTransition {
+  blockTakeDuration: number
```
KvelaGorrrrnio, line 2: Just a thought: Should we have units on these?

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -182,6 +186,21 @@ export class RundownTimelineService implements RundownService {
...
+    let onAirPart: Part
+
+    try {
+      onAirPart = rundown.getActivePart()
+    } catch (error) {
+      // If 'getActivePart()' throws it means that we don't have any active Part yet which means the Take is not blocked - hence we can simply return.
+      return
+    }
+    const now: number = Date.now()
+    if (now - onAirPart.getExecutedAt() < onAirPart.getInTransition().blockTakeDuration) {
```
KvelaGorrrrnio: Suggestion: Since `now` is just used once and the expression is fairly simple, I think we can dissolve it.
A suggestion for making the condition a bit easier to reason about is to make the condition about absolute time references rather than durations. E.g.:
```typescript
if (Date.now() < onAirPart.getExecutedAt() + onAirPart.getInTransition().blockTakeDuration)
```

KvelaGorrrrnio (approved):

src/model/value-objects/in-transition.ts
LindvedKrvang: I think that would be a good idea. I am also very open for a new name. I just took the same as was used by Core

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
LindvedKrvang: Updated.

src/business-logic/services/test/rundown-timeline-service.spec.ts
LindvedKrvang: Nice catch. It has been removed.

src/business-logic/services/test/rundown-timeline-service.spec.ts
LindvedKrvang: Updated the three methods doing this

src/business-logic/services/rundown-timeline-service.ts
LindvedKrvang: Fine for me. It's been updated

RasmusAlbrektsen (approved):

MadsThy (approved):
I'm not entirely sure i understand why the tests have try/catch in them where we don't use the catch for anything, but I don't want to make this PR wait just because of that :) Would love to hear why you are doing it at some point though

PR #219, opened 2024-09-12
SOF-2162  Delete status messages of deleted Devices

LindvedKrvang:
Delete StatusMessages for Devices that has been deleted.
On startup: delete StatusMessages for Devices that no longer exists.
Added the timestamp `lastUpdatedTimestamp` to StatusMessages.

Note: The notification that is going to be send when a Device is deleted is, in lack of better words, not good.
This is because when we receive a delete event from Mongo, we only receive the id of the deleted entity in Mongo. That means that we have no idea of what the Device was called.


MadsThy (approved):

src/business-logic/services/device-changed-service.ts
```diff
@@ -74,10 +82,22 @@ export class DeviceChangedService implements DataChangeService {
...
       title: device.name,
       message: device.statusMessage
     }
   }
+
+  private async onDeviceDeleted(deviceId: string): Promise<void> {
+    const deletedDevice: Device = {
+      id: deviceId,
+      name: '',
+      statusMessage: 'Device was deleted',
```
KvelaGorrrrnio, line 96: Since this is a sentence, it could end with a period :)

src/model/entities/status-message.ts
```diff
@@ -5,4 +5,5 @@ export interface StatusMessage {
   statusCode: StatusCode
   title: string
   message: string
+  lastUpdatedTimestamp?: number
```
KvelaGorrrrnio, line 8: It would be nice if this was a required attribute, so clients can make use of it reliably. I have made a suggestion for it, that uses `PeripheralDevice.lastSeen` for device status messages in the commit: 1d2e080f21f1494a1db51f93eafbafc6d398f2cb

KvelaGorrrrnio (approved):
Nice work :)

src/model/entities/status-message.ts
LindvedKrvang: The reason for the optional was so that we could just let the `StatusMessageRepository` handle the timestamp. This way, we wouldn't need to set the same timestamp over and over in the code.

I need some more context as to how `lastSeen` make sense in a status message?

KvelaGorrrrnio: By having the status message repository set the timestamp for when it was last updated, the attribute expresses when it was last persisted, not when it was last updated.
Also by doing it this way, we expect that a status message always have the attribute, but we have to check for the existence before using it.

Regarding `lastSeen`: In the case that a device status is updated to good, then the timestamp will express when it was last seen, which will coincide with the last connected timestamp. When a device is disconnected, it will provide a good estimation of when the disconnect happened. For internal errors in devices it might not be precise, but should not give any issues.

PR #220, opened 2024-09-12
SOF-2139 Readable error message on reingest fail.

LindvedKrvang:
Returns a readable error message when trying to reingest while iNewsGateway is not connected


src/business-logic/services/tv2-inews-ingest-service.ts
```diff
@@ -12,7 +13,15 @@ export class Tv2INewsIngestService implements IngestService {
   public async reloadIngestData(rundownId: string): Promise<void> {
     const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
     const url: string = this.getReingestUrl(rundown.name)
-    await this.httpService.post(url, null)
+    try {
+      await this.httpService.post(url, null)
+    } catch (error) {
+      const errorResponse: HttpErrorResponse = error as HttpErrorResponse
+      if (errorResponse.code === HttpErrorResponseCode.CONNECTION_REFUSED) {
+        throw new ServiceUnavailableException('Unable to reingest data from iNews. Check your iNews connection...')
```
KvelaGorrrrnio, line 21: Note: `ECONNREFUSED` indicates that we can't reach the iNews Gateway. So this probably has nothing to do with the connection to iNews :)
And `connection...` should just be `connection.`.

KvelaGorrrrnio (approved):

src/business-logic/services/tv2-inews-ingest-service.ts
LindvedKrvang: Would you want it to say `iNews-gateway` instead?

RasmusAlbrektsen (approved):

PR #221, opened 2024-09-12
SOF-2207: Fix non unique action ids

KvelaGorrrrnio:
Action manifests might share name and content, but only references the name for determining uniqueness.
This is now changed to use a hash of the content in the action id to differentiate between them. In some cases, where both name and content is the same, the duplicates are filtered out and the action with the lowest rank is kept.

Actions for overlay action manifests are treated as different entities if the rank differs.

`TvActionManifestSplitScreenSourceType.LIVE` is renamed to `TvActionManifestSplitScreenSourceType.REMOTE` to match the terminology used other places in the codebase.


src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
```diff
@@ -173,10 +173,12 @@ export class Tv2AudioActionFactory extends ActionFactory {
...
   private createAudioBedActionsFromActionManifests(blueprintConfiguration: Tv2BlueprintConfiguration, actionManifests: Tv2ActionManifest[]): Tv2AudioAction[] {
-    return actionManifests
-      .filter(this.isAudioBedActionManifest.bind(this))
-      .filter(audioBedActionManifest => this.isAudioBedConfigured(audioBedActionManifest.data.name, blueprintConfiguration))
-      .map(audioBedActionManifest => this.createAudioBedActionsFromActionManifest(blueprintConfiguration, audioBedActionManifest))
+    return this.removeDuplicateActions(
+      actionManifests
+        .filter(this.isAudioBedActionManifest.bind(this))
+        .filter(audioBedActionManifest => this.isAudioBedConfigured(audioBedActionManifest.data.name, blueprintConfiguration))
+        .map(audioBedActionManifest => this.createAudioBedActionFromActionManifest(blueprintConfiguration, audioBedActionManifest))
```
LindvedKrvang: This is quite a lot to put inside a parameter. Consider extrating what used to be just the return statement into a local variable.

src/blueprints/tv2/action-factories/tv2-graphics-action-factory.ts
```diff
@@ -163,7 +163,7 @@ export class Tv2GraphicsActionFactory extends ActionFactory {
     })
   }
 
-  public createGraphicsActions(blueprintConfiguration: Tv2BlueprintConfiguration, actionManifests: Tv2ActionManifest[]): Action[] {
+  public createGraphicsActions(blueprintConfiguration: Tv2BlueprintConfiguration, actionManifests: Tv2ActionManifest[]): Tv2Action[] {
```
LindvedKrvang, line 166: We should start looking into merging `Action` and `Tv2Action` :)

src/blueprints/tv2/action-factories/tv2-graphics-action-factory.ts
```diff
@@ -298,25 +298,27 @@ export class Tv2GraphicsActionFactory extends ActionFactory {
...
+            partInterface: partInterface,
+            pieceInterfaces: [pieceInterface]
+          },
+          metadata: {
+            contentType: Tv2ActionContentType.GRAPHICS
+          }
         }
-      }
-    })
+      })
```
LindvedKrvang: Same comment about this being quite a lot to put inside a parameter.

src/blueprints/tv2/helpers/tv2-action-manifest-mapper.ts
```diff
@@ -26,7 +26,7 @@ const OVERLAY_GRAPHICS_ACTION_MANIFEST_IDS: string[] = ['studio0_graphicsLower',
 
 export class Tv2ActionManifestMapper {
 
-  public mapToSplitScreenManifestData(blueprintConfiguration: Tv2BlueprintConfiguration, actionManifests: ActionManifest[]): Tv2SplitScreenManifestData[] {
+  public getSplitScreenManifestData(blueprintConfiguration: Tv2BlueprintConfiguration, actionManifests: ActionManifest[]): Tv2SplitScreenManifestData[] {
```
LindvedKrvang: Disagree with this name change. This method is not getting anything. It's mapping the manifest that you _already have_ into SplitScreenManifests.

src/blueprints/tv2/value-objects/tv2-action-manifest.ts
```diff
@@ -1,4 +1,4 @@
 import { ActionManifest } from '../../../model/entities/action'
 import { Tv2ActionManifestData } from './tv2-action-manifest-data'
 
-export type Tv2ActionManifest = ActionManifest<Tv2ActionManifestData>
+export type Tv2ActionManifest<Tv2ActionManifestDataType extends Tv2ActionManifestData = Tv2ActionManifestData> = ActionManifest<Tv2ActionManifestDataType>
```
LindvedKrvang, line 4: This just got a lot more difficult to read

LindvedKrvang (approved):

src/blueprints/tv2/action-factories/tv2-audio-action-factory.ts
KvelaGorrrrnio: In this scenario, where it is solely composition I don't have an issue with it, but the other one you commented on I agree with. I have moved this to a local variable as suggested :)

src/blueprints/tv2/action-factories/tv2-graphics-action-factory.ts
KvelaGorrrrnio: Good point!

src/blueprints/tv2/action-factories/tv2-graphics-action-factory.ts
KvelaGorrrrnio: I have moved the logic to the new method `createFullscreenGraphicsAction`.

src/blueprints/tv2/helpers/tv2-action-manifest-mapper.ts
KvelaGorrrrnio: `map` indicates that each element in the given collection is transformed, which is not the case for this method. It filters for split screen action manifests and then transform/map the remaining entries.
I have updated the name to `filterAndMapToSplitScreenManifestData`.

src/blueprints/tv2/value-objects/tv2-action-manifest.ts
KvelaGorrrrnio: And got a lot more usable :) When we remove TV2 prefixes and types, then this will go as well.

RasmusAlbrektsen (approved):

src/blueprints/tv2/helpers/tv2-action-manifest-mapper.ts
LindvedKrvang: Obligatory `and` in method name comment ;)

LindvedKrvang (approved):

PR #226, opened 2024-09-18
SOF-2259 Unplanned Transitions no longer uses the Effect source

LindvedKrvang:
Note: Dependent on https://github.com/tv2/sofie-server/pull/225
Update: All dependent branches has been merged.

When overriding a planned Effect Transition with an unplanned Transition, we would sometimes use the source of the planned Effect instead of the source of the Part we were going to (the source of the Camera or Remote).

This happend because, in order to find the "program" Piece we just checked if it had a TimelineObject for the VideoMixerMeProgram layer (for Atem it is `atem_me_program`).

When having a planned Effect Transition we would get a Part that looked like this:
```
// Part
{
  pieces: [
    {
      name: 'EffectPiece',
      timelineObjects: [
        { layer: 'atem_me_program', source: 29 }
      ]
    },
    {
      name: 'CameraOnePiece',
      timelineObjects: [
        { layer: 'atem_me_program', source: 1 }
      ]
    }
  ]
}
```
With the previous implementation, both Pieces above were eligable for selection of the Transition source, but only one had the right source.

We now also go and look at the `metadata.outputLayer` on Piece, because only the Pieces that we want to "show" in program has the value `Tv2PieceOutputLayer.PROGRAM`.


src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -137,12 +137,19 @@ export class Tv2TransitionEffectActionFactory extends ActionFactory {
...
       type: MutateActionType.PIECE,
       updateActionWithPiece: (action: Action, piece: Piece) => this.updateTimelineObjectsWithTransitionEffect(action as Tv2TransitionEffectAction, piece),
-      piecePredicate: (piece: Piece) => piece.getTimelineObjects().some(timelineObject => timelineObject.layer === this.videoMixerTimelineObjectFactory.getProgramLayer()),
+      piecePredicate: (piece: Piece) => this.isPieceProgramPiece(piece),
     }
     mutateActionMethods.push(updateTransitionMutateAction)
     return mutateActionMethods
   }
 
+  private isPieceProgramPiece(piece: Piece): boolean {
```
RasmusAlbrektsen: I guess the name `isProgramPiece` would suffice here, especially since we also have `piece` parameter already.

RasmusAlbrektsen (dismissed):

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
LindvedKrvang: No objections here. Its been updated.

RasmusAlbrektsen (approved):

MadsThy (approved):

PR #228, opened 2024-09-19
SOF-2206/Split Screen Action Locators Look Weird Due to Wrong Opacity Value

RasmusAlbrektsen:
The opacity value which used to be 100 was passed all the way through to the CasparCG. The theory is that this used to be handled elsewhere before the TSR was updated. Changing it to 1 which is the correct CasparCG value (interval between 0 and 1) fixed the issue we experienced.


src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
```diff
@@ -63,7 +63,7 @@ export class Tv2CasparCgTimelineObjectFactory implements Tv2GraphicsElementTimel
         data: this.createFullscreenGraphicsTemplateData(blueprintConfiguration, fileName),
         useStopCommand: false,
         mixer: {
-          opacity: 100
+          opacity: 1
```
LindvedKrvang, line 66: Maybe we should put a comment on the `Mixer.opacity` interface about the value range been 0 to 1?

LindvedKrvang (approved):

src/blueprints/tv2/timeline-object-factories/tv2-caspar-cg-timeline-object-factory.ts
RasmusAlbrektsen: Good point! I have added it :-)

MadsThy (approved):

LindvedKrvang (approved):

PR #230, opened 2024-09-23
SOF-2276 Send Ingest events in bulks

LindvedKrvang:
Since we used to send an event for _every_ ingest change we received, we could potentially spam the event system. For a large Rundown, it is not uncommon to have hundreds of Ingest changes (when ingesting it from scratch). Sending hundreds of events had a severe impact on the clients.

This PR bulks the Ingest event before sending them.
When an IngestEvent is received we "store" the event and start a Timer for 500ms. If another IngestEvent is received within the 500ms, we stop the timer, add the new event to the "storage" and sets a new timer. 
This continues untill we either: Don't receive a new Event, or: We haven't sent a Bulk event within the last 5000ms. The latter is so we don't have to wait for however many seconds it takes to process the Rundown (for a large Rundown it could easily be 20+ seconds), before seeing anything happen in the GUI.

 Whenever we send the Bulk Event, we take all the IngestEvent in "storage" put them in a collection and send them in the event.

Also in this PR:
Updated when we send the SetNextEvent. Whenever we build the Timeline during Ingest, we would also send a SetNext Event in case the Next Part had changed. This happened twice a second, which also ended up being a lot of events emitted for what is, for the most part, the same Segment and Part set as next. 

Now we only send a SetNext Event during Ingest, if the Segment or Part has changed.


src/presentation/services/rundown-event-service.ts
```diff
@@ -96,6 +144,14 @@ export class RundownEventService implements RundownEventEmitter, RundownEventObs
 
   public emitSetNextEvent(rundown: Rundown): void {
     const event: PartSetAsNextEvent = this.rundownEventBuilder.buildSetNextEvent(rundown)
+    if (this.timeoutIdentifier) { // We are currently ingesting
```
MadsThy: How about renaming this timeoutIdentifier to "currentlyIngesting" then? 😄

MadsThy (approved):

src/presentation/services/rundown-event-service.ts
LindvedKrvang: I would like to keep the `timeoutIdentifier` part since that is what it is. `currentlyIngesting` sounds like, to me, as a boolean. 
But something like `ingestTimeoutIdentifier` could be nice.

LindvedKrvang: I have updated the name to `ingestTimeoutIdentifier`

RasmusAlbrektsen (approved):

src/presentation/value-objects/rundown-event.ts
```diff
@@ -121,3 +121,8 @@ export interface PartUnsyncedEvent extends RundownEvent {
   type: IngestEventType.PART_UNSYNCED
   part: PartDto
 }
+
+export interface BulkIngestEvent extends RundownEvent {
+  type: IngestEventType.BULK_INGEST_UPDATES,
+  ingestEvents: RundownEvent[]
```
KvelaGorrrrnio: It would be nice to limit these to ingest events only, with an interface like
```typescript
export interface IngestEvent extends RundownEvent {
  type: IngestEventType
}
```
Then the ingest events can extend from this, including the `BulkIngestEvent`. One thing to keep in mind is that we should not support nested BulkIngestEvents. This can be achieved by using type arguments for the interface, e.g. something like:

```typescript
export interface IngestEvent<Type extends IngestEventType extends RundownEvent {
  type: Type
}

export interface BulkIngestEvent extends IngestEvent<IngestEventType.BULK_INGEST_UPDATES> {
  ingestEvents: IngestEvent<Omit<IngestEventType, IngestEventType.BULK_INGEST_UPDATES>>[]
}

File: src/presentation/services/rundown-event-service.ts [hunk 3]
```diff
@@ -43,12 +47,56 @@ export class RundownEventService implements RundownEventEmitter, RundownEventObs
...
+    }
+    this.ingestEventQueue.get(rundownEvent.rundownId)?.push(rundownEvent)
+
+    this.startBulkEventCallback()
+  }
+
+  private startBulkEventCallback(): void {
+    const isTimeoutStarted: boolean = !!this.ingestTimeoutIdentifier
+    const isBulkEmitEventOverdue: boolean = Date.now() > this.lastBulkEmittedTimestamp + MAX_TIME_BEFORE_SENDING_BULK_EVENT_IN_MS
+    const isTooLateToDelayBulkEvent: boolean = Date.now() > this.callbackStartedTimestamp + BULK_EVENT_THRESHOLD_IN_MS
```
KvelaGorrrrnio: This might just be me, but I find the names hard to understand. The `isTooLateToDelayBulkEvent` I would assume, seen in isolation, was checking against the max duration allowed before sending a bulk event.

Suggestion:
```typescript
const isPastDebouncePeriod = Date.now() > this.lastQueuedIngestEventTimestamp + INGEST_EVENT_DEBOUNCE_DURATION_IN_MS
const hasReachedMaxDebounceDuration = Date.now() > this.firstQueuedIngestEventTimestamp + MAX_INGEST_EVENT_DEBOUNCE_DURATION_IN_MS
```

KvelaGorrrrnio (approved):

src/presentation/value-objects/rundown-event.ts
LindvedKrvang: Good suggestion. I have implemented it. Your `Omit` suggestion, however, did not compile so a `BulkIngestEvent` is possible to have a nested `BulkIngestEvent`. Though in the code as of this writing, it will never happend.

src/presentation/services/rundown-event-service.ts
LindvedKrvang: Those are better names. They have been updated :)

KvelaGorrrrnio (approved):

PR #231, opened 2024-09-23
SOF-2275 Duplicate TimelineObjects in Lookahead get a random Postfix to its id

LindvedKrvang:
When resolving the Timeline. If there are two TimelineObjects that have the same Id, the Timeline is not able to resolve.
Our TimelineObjects useally have a "static" id i.e. the TimelineObject that decides the program source in Atem has the id `atem_program`. 

This is useally not a problem since when we convert the Rundown to a Timeline, we make `TimelineObjectGroups` which gets their own Id and all TimelineObjects within that group gets the groups Id as a prefix. So `atem_program` actually becomes something like `active_part_group_atem_program`.  This combined with that we make a single group the for Active Part, another group for the Next Part and a third group for the Previous Part, makes a clash of Ids highly unlikely.

However, it's possible for the Lookahead group to find two or more of the same TimelineObject (belonging to different Parts) and put all of those TimelineObjects within the Lookahead group, which means the ids will clash.

To prevent this, we check to see if there are any duplicates in the Lookahead group. If yes, then we assign a random postfix to the id to prevent a clash.

Note: We don't actually update the Id. We make a copy of the TimelineObject, which has the updated id, and use that for the rest of the Timeline resolving.


RasmusAlbrektsen (approved):

src/business-logic/services/superfly-timeline-builder.ts
```diff
@@ -339,10 +339,25 @@ export class SuperflyTimelineBuilder implements TimelineBuilder {
...
+        timelineObjects.push({
+          ...timelineObject,
+          id: `${timelineObject.id}_${Math.round(Math.random() * 1000)}`
+        })
+      } else {
+        timelineObjects.push(timelineObject)
+      }
+      return timelineObjects
+    }, [] as LookaheadTimelineObject[])
+  }
```
MadsThy: I found this really hard to read 😄 Could just be me though, but i have a few suggestions that could maybe be more readable, but i'm not sure if it is actually solving the thing, since i don't have a test to hold it up to 😄 

If you like any of these suggestions, could you check if the code works still? Like i said, i am not sure at all if my suggestions work, but i found them more readable :) Maybe you do too.

```
private postFixRandomIdToDuplicateObjects(timelineObjects: LookaheadTimelineObject[]): LookaheadTimelineObject[] {
    const existingIds = new Set<string>()
    return timelineObjects.map(timelineObject => {
      if (existingIds.has(timelineObject.id)) {
        return { ...timelineObject, id: `${timelineObject.id}_${Math.round(Math.random() * 1000)}` }
      }
      existingIds.add(timelineObject.id)
      return timelineObject
    })
  }
```

```
private postFixRandomIdToDuplicateObjects(timelineObjects: LookaheadTimelineObject[]): LookaheadTimelineObject[] {
    const uniqueObjects: LookaheadTimelineObject[] = []
    const idSet: Set<string> = new Set()

    for (const timelineObject of timelineObjects) {
      let newId = timelineObject.id
      while (idSet.has(newId)) { newId = `${timelineObject.id}_${Math.round(Math.random() * 1000)}` }
      idSet.add(newId)
      uniqueObjects.push({ ...timelineObject, id: newId })
    }

    return uniqueObjects
  }
```

MadsThy (approved):

src/business-logic/services/superfly-timeline-builder.ts
```diff
@@ -339,10 +339,25 @@ export class SuperflyTimelineBuilder implements TimelineBuilder {
...
     })
   }
 
+  private postFixRandomIdToDuplicateObjects(timelineObjects: LookaheadTimelineObject[]): LookaheadTimelineObject[] {
+    return timelineObjects.reduce((timelineObjects: LookaheadTimelineObject[], timelineObject: LookaheadTimelineObject) => {
+      const timelineObjectIdAlreadyExist: boolean = timelineObjects.some(to => to.id === timelineObject.id)
+      if (timelineObjectIdAlreadyExist) {
+        timelineObjects.push({
+          ...timelineObject,
+          id: `${timelineObject.id}_${Math.round(Math.random() * 1000)}`
```
KvelaGorrrrnio: Note: We should have a better way of generating uniqueness. `Math.random() * 1000` is not guaranteed to give uniqueness. `process.hrtime.bigint()` would give a more unique id in this use case :)

src/business-logic/services/superfly-timeline-builder.ts
KvelaGorrrrnio: The map solution @MadsThy presents seems like a good alternative :)

KvelaGorrrrnio (approved):

src/business-logic/services/superfly-timeline-builder.ts
LindvedKrvang: I agree. This is not my proudest work...
I went with your first option. I'm not a fan of the side-effect (updating the `Set` outside of the `map`), BUT in this case the readability wins :)

LindvedKrvang: It's been updated

src/business-logic/services/superfly-timeline-builder.ts
LindvedKrvang: Updated

KvelaGorrrrnio (approved):

PR #237, opened 2024-09-26
SOF-2290 Gracefully handle misconfiguration in Transitions table

LindvedKrvang:
When creating TransitionActions, if there is an effect configured in the `Transitions` table that doesn't exist in the `Breakers` table, we now log the exception as a warning instead of stopping the entire flow.


src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -82,10 +82,18 @@ export class Tv2TransitionEffectActionFactory extends ActionFactory {
...
+        try {
+          return [
+            this.createBreakerTransitionEffectAction(PieceActionType.INSERT_PIECE_AS_NEXT, transitionEffect, blueprintConfiguration),
+            this.createBreakerTransitionEffectAction(PieceActionType.INSERT_PIECE_AS_NEXT_AND_TAKE, transitionEffect, blueprintConfiguration)
+          ]
+        } catch (exception) {
+          if (exception instanceof Tv2MisconfigurationException) {
+            this.logger.data(exception).warn(exception.message)
+            return []
+          }
```
KvelaGorrrrnio, line 94: Is the if statement necessary?
If it fails in another way, won't we still want to generate rest of the actions?

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -82,10 +82,18 @@ export class Tv2TransitionEffectActionFactory extends ActionFactory {
...
-          this.createBreakerTransitionEffectAction(PieceActionType.INSERT_PIECE_AS_NEXT_AND_TAKE, transitionEffect, blueprintConfiguration)
-        ]
+        try {
+          return [
+            this.createBreakerTransitionEffectAction(PieceActionType.INSERT_PIECE_AS_NEXT, transitionEffect, blueprintConfiguration),
+            this.createBreakerTransitionEffectAction(PieceActionType.INSERT_PIECE_AS_NEXT_AND_TAKE, transitionEffect, blueprintConfiguration)
+          ]
+        } catch (exception) {
+          if (exception instanceof Tv2MisconfigurationException) {
+            this.logger.data(exception).warn(exception.message)
```
KvelaGorrrrnio, line 92: In this case, the data will be the message with a stack trace. To avoid the duplication and get a bit more context, I suggest that the message is changed to something like:
```typescript
this.logger.data(exception).warn(`Failed generating transition effect actions for the transition effect '${transitionEffect.name}'.`)
```

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
KvelaGorrrrnio: It might also be that other types of exceptions should be logged as error?

LindvedKrvang: I think that highly depends on the Exception whether or not we want to not throw it.
Right now, our approach is to throw errors when they happen. If we find another error that we still want to generate Actions for, then we can add it. Untill then it's YAGNI.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
LindvedKrvang: I think I would be more inclined to simply only provide the stacktrace to the data then.

KvelaGorrrrnio: For what reason? :)

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
KvelaGorrrrnio: Do we have a use case where we don't want to generate the rest of the actions?
If not, then I think YAGNI would be that we don't need the check. By allowing other errors to fall through here, we allow that static actions, like the ones created by `createEmptyMixTransitionEffectAction`, are not available to the user because e.g. some action manifest data is malformed.

LindvedKrvang: We actually don't, but that's because of somewhere else in the system.
The `ingest-data-changed-service` makes atleast two calls to `Blueprints.generateActions()`. One with `actionManifests` for each Rundown and one call without ActionManifest for the "System Actions".

This makes it so that we generate "Rundown specific" Actions through the Manifests, but also ensures that we have all the "non-Rundown specific" Actions available, like the `createEmptyMixTransitionEffectAction`.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
LindvedKrvang: So we limit the amount of magic strings in the codebase. Or, so we don't need to write the (almost) same error message more than once.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
KvelaGorrrrnio: Generating actions from action manifests and generating system actions are called in the same flow, so if this throws, then the method for generating system actions is not called. If the static actions are not available prior to the generate actions call, then they are not there after.
In SOF-2296, I am looking into making this more robust.

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
KvelaGorrrrnio: I will argue that having a better understanding of the context when debugging with logs is more important than avoiding having some "magic" strings in error messages :)
Magic strings are only dangerous when they pertain to code functionality. This is a string literal that never will and never should be referenced. If we type this error message multiple times, then we have most likely spread out the same logic to too many places, and the creation of the error or the scenario can be extracted, to avoid the maintenance cost of the duplication.

KvelaGorrrrnio (approved):

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
LindvedKrvang: Since you are working on this in SOF-2296 I suggest to not do more about it in this PR?

RasmusAlbrektsen (approved):

PR #238, opened 2024-09-27
SOF-2296: Misconfiguration should not crash server on startup

KvelaGorrrrnio:
- Ensures that rejected calls for generating actions are caught, so it can't crash the server. Same goes for synchronizing ingest entities on startup.
- To make the system a bit more robust and usable in terms of action generation, `Tv2ActionService.generateActions` skips failed action category factory calls. E.g. if the creation of camera actions fail, then the subsequent action categories are still generated.
- Adds support for action manifests for split screens with replay sources.


RasmusAlbrektsen (dismissed):

src/blueprints/tv2/tv2-action-service.ts
```diff
@@ -80,18 +72,27 @@ export class Tv2ActionService implements BlueprintGenerateActions {
...
-      ...this.remoteActionFactory.createRemoteActions(blueprintConfiguration),
-      ...this.audioActionFactory.createAudioActions(blueprintConfiguration, actionManifests),
-      ...this.transitionEffectActionFactory.createTransitionEffectActions(blueprintConfiguration),
-      ...this.graphicsActionFactory.createGraphicsActions(blueprintConfiguration, actionManifests),
-      ...this.videoClipActionFactory.createVideoClipActions(blueprintConfiguration, actionManifests),
-      ...this.videoMixerActionFactory.createVideoMixerActions(blueprintConfiguration),
-      ...this.splitScreenActionFactory.createSplitScreenActions(blueprintConfiguration, actionManifests),
-      ...this.replayActionFactory.createReplayActions(blueprintConfiguration),
-      ...this.robotActionFactory.createRobotActions()
+    const actionGenerators: [string, () => Action[]][] = [
```
LindvedKrvang: This is an interesting type. I think it could benefit from being extracted to an actual type. Don't have a name for it thouhg...

RasmusAlbrektsen (approved):

src/business-logic/exhaustive-case-checker.ts
```diff
@@ -7,8 +7,8 @@ export class ExhaustiveCaseChecker {
 	 * It's allowed to be static since it verifies at compile time that we are not missing e.g. a switch case for an enum.
 	 * Since this happens at compile time we would never need to substitute this method, not even in tests, which is what allows this to be static.
 	 */
-  public static assertAllCases(unexpectedCase: never): never {
+  public static assertAllCases(unexpectedCase: never, scenario: string): never {
```
LindvedKrvang, line 10: Personally, I wouldn't mind if we deleted this method.

LindvedKrvang (approved):

src/blueprints/tv2/tv2-action-service.ts
```diff
@@ -80,18 +72,27 @@ export class Tv2ActionService implements BlueprintGenerateActions {
...
+      ['camera', (): Action[] => this.cameraActionFactory.createCameraActions(blueprintConfiguration)],
```
LindvedKrvang: Not a fan of having to write all these magic strings.

I suggest either having an enum for it or let `ActionFactory` implement the abstract method `getActionType(): string` so you can get the name from each ActionFactory implementation

src/blueprints/tv2/tv2-action-service.ts
KvelaGorrrrnio: It is a list of tuples :)

src/business-logic/exhaustive-case-checker.ts
KvelaGorrrrnio: I think it provides good value with the assertion :)

src/blueprints/tv2/tv2-action-service.ts
KvelaGorrrrnio: I have extracted it out into an enum.

LindvedKrvang (approved):

src/blueprints/tv2/tv2-action-service.ts
LindvedKrvang: This still looks "interesting": `[]][]` ;)

PR #241, opened 2024-10-01
SOF-2296: Ignore erroneous action creation from action manifests

KvelaGorrrrnio:
- Split screen action creation method calls are caught, so the following split screen action manifests can be processed.
- Adds blueprint validation check for ensuring all defined transitions have a breaker configuration.
- The methods on Tv2ActionManifestMappers are made more robust, by catching failed conversions, so following action manifests can be processed.
- Errors thrown from the `HttpService` uses the body content for error message when available. This provides better error messages, at the moment for the the reingest endpoint, that calls the iNews Gateway.

It is only split screen that was updated here, since it is one of the action categories where action generation relies on blueprint configuration per individual action manifest. For example, for graphics, the dependencies to the configuration is not tied to a specific graphics template/action. It is tied to the overall graphics category.

The current solution logs the failed attempt, but ideally it should return succesful and failed attempts, so somewhere in the call-hierarchy, someone can convert them to status messages and notify the user.


src/business-logic/services/interfaces/http-service.ts
```diff
@@ -1,11 +1,3 @@
...
-  code: HttpErrorResponseCode
-}
-
-export enum HttpErrorResponseCode {
-  CONNECTION_REFUSED = 'ECONNREFUSED'
-}
-
 export interface HttpService {
-  post(url: string, body: unknown): unknown
+  post(url: string, body?: string): unknown
```
MadsThy, line 2: are you sure you want to make body optional for posts? 😄 I see why you want to do it, when all you were sending before was null, but... maybe it should be made into a GET instead then? :)

src/business-logic/services/tv2-inews-ingest-service.ts
```diff
@@ -14,11 +16,15 @@ export class Tv2INewsIngestService implements IngestService {
...
+      await this.httpService.post(url)
     } catch (error) {
-      const errorResponse: HttpErrorResponse = error as HttpErrorResponse
-      if (errorResponse.code === HttpErrorResponseCode.CONNECTION_REFUSED) {
-        throw new ServiceUnavailableException('Unable to reingest data from iNews. Check your iNews connection...')
+      if (error instanceof HttpError) {
+        if (error.code === HttpErrorCode.CONNECTION_REFUSED) {
+          throw new ServiceUnavailableException('Unable to reingest data from iNews. Check your iNews connection.')
+        }
+        if (/does not exist in playlist/i.test(error.message)) {
```
MadsThy, line 25: Just had a little chat with CoPilot, cause I couldn't remember what / / notation does 😄 Turns out I forgot that that is how you can do "contains".

Just a suggestion, but could it be more readable by doing this? :)

```suggestion
        if (error.message.toLowerCase().includes('does not exist in playlist')) {
```

MadsThy (approved):
Looks great :)

src/business-logic/services/interfaces/http-service.ts
LindvedKrvang: That's a good discussion to have. Which HTTP request should you use when you want to "activate" a flow in the system?

We shouldn't have the discussion here, but it would be a good thing to revisit. From our previous talks about it, we landed on POST, but maybe you have some other insights Mads? :)

src/business-logic/services/tv2-inews-ingest-service.ts
LindvedKrvang: Isn't the // just a regex?

src/business-logic/services/interfaces/http-service.ts
LindvedKrvang: Hmm, I think my comment took perspective in a different context than the one we are in here :p

LindvedKrvang (approved):

src/business-logic/services/interfaces/http-service.ts
KvelaGorrrrnio: It is a valid point. Alternatively the body can default to an empty string?
I don't think replacing it with GET will be a good solution for the use cases we have at the moment. As Lindved mentions, it is a discussion that we should have.

src/business-logic/services/tv2-inews-ingest-service.ts
KvelaGorrrrnio: `/<content>/` is just a regex. The `i` flag indicates insensitive case-matching :)
Personally I find it more verbose, without adding to the semantics – but that is probably just me :D

src/blueprints/tv2/helpers/test/tv2-action-manifest-mapper.spec.ts
```diff
@@ -0,0 +1,256 @@
...
+import { Tv2PieceType } from '../../enums/tv2-piece-type'
+import { Tv2BlueprintConfiguration } from '../../value-objects/tv2-blueprint-configuration'
+import { Tv2BlueprintConfigurationTestFactory } from '../../test/tv2-blueprint-configuration-test-factory'
+import { Tv2Logger } from '../../tv2-logger'
+import { anyString, anything, instance, mock, when } from '@typestrong/ts-mockito'
+import { Tv2SourceLayer } from '../../value-objects/tv2-layers'
+
+describe(Tv2ActionManifestMapper.name, () => {
+  describe(Tv2ActionManifestMapper.prototype.filterAndMapToSplitScreenManifestData.name, () => {
+    describe('when an action manifest has a invalid input source is given', () => {
```
RasmusAlbrektsen: I think the `is` should be removed for this to make sense right?
`when an action manifest has an invalid input source given`

RasmusAlbrektsen (approved):

src/blueprints/tv2/helpers/test/tv2-action-manifest-mapper.spec.ts
KvelaGorrrrnio: I have updated the description so it makes sense :D

src/business-logic/services/tv2-inews-ingest-service.ts
LindvedKrvang: Personally, I lean more towards the `.includes()` suggestion, but that's because that is closer to the code simply being an english sentence.

LindvedKrvang (approved):

PR #243, opened 2024-10-02
SOF-2299 Introduce InsertPieceAndTake method on RundownService

LindvedKrvang:
The Action InsertPieceAndTake would first call the `InsertPieceAsNext` method in `RundownService` and thereafter call the `TakeNext` method.

This meant that before we could Take the next Part, we had to wait for `InsertPieceAsNext` to save to the database.

This PR introduces the `insertPieceAsNextAndTake` on `RundownService` so we can insert the Piece and take the next Part as "one transaction", so we don't have to wait for writing to the database.

P.S. naming suggestions for the new private methods in the `RundownService` are highly appreciated :)


MadsThy (approved):

KvelaGorrrrnio (approved):
🚀

RasmusAlbrektsen (approved):

PR #244, opened 2024-10-03
SOF-2298: Fix ingest state synchronization (Part 1)

KvelaGorrrrnio:
Fixing the synchronization will be a series of PRs to make it somewhat manageable to review, since it will consist of multiple parts.

This PR swaps RUNDOW_UPDATE and PART_UPDATE priority, since the logic for UPDATE events can be executed as a create event.

Added more contexts to logs, since it was very hard to see what happened when debugging.

When deleting a rundown, the actions are deleted and the rundown is unmarked for generating actions.


MadsThy (approved):

src/business-logic/services/ingest-data-changed-service.ts
```diff
@@ -132,21 +139,21 @@ export class IngestDataChangedService implements DataChangeService {
   }
 
   private listenForRundownChanges(rundownChangeListener: DataChangedListener<IngestedRundown>): void {
-    rundownChangeListener.onCreated(rundown => this.enqueueEvent(IngestEventPriority.RUNDOWN_CREATE, () => this.createRundown(rundown)))
-    rundownChangeListener.onUpdated(rundown => this.enqueueEvent(IngestEventPriority.RUNDOWN_UPDATE, () => this.updateRundown(rundown)))
-    rundownChangeListener.onDeleted(rundownId => this.enqueueEvent(IngestEventPriority.RUNDOWN_DELETE, () => this.deleteRundown(rundownId)))
+    rundownChangeListener.onCreated(rundown => this.enqueueEvent(IngestEventPriority.RUNDOWN_CREATE, 'create', 'rundown', rundown.id, () => this.createRundown(rundown)))
```
LindvedKrvang, line 142: I know these "techinically" aren't magic strings, but it would still be nice if they weren't strings.
Since this PR is a first of many, just keep this mind when doing the rest :)

LindvedKrvang (approved):

src/business-logic/services/ingest-data-changed-service.ts
KvelaGorrrrnio: We will in the next part most likely remove this all together. So bear with it for now 😁

PR #247, opened 2024-10-08
SOF-2298: Synchronize-only ingest flow

KvelaGorrrrnio:
_**Note:** For the web client to behave correctly, the PR https://github.com/tv2/sofie-web-client/pull/272 is needed._

The PR introduces a new approach for handling ingest events observed through MongoDB that Sofie Core generates.

The approach used is to analyze the entire state of a rundown, when a sequence of observed ingest events for the rundown (or more rundowns) ends. By doing it this way, the code that runs at startup to ensure correct state and the code that runs when ingest events are received is the same.

The code analyzes the ingest data for the rundown against the existing rundown and classifies any changes to the rundown metadata, along with create, update and delete for segments and parts.

If a segment is affected (created, updated or deleted), then part changes under that segment is ignored, since the segment includes the new state of the parts. So the part changes are redundant and will most likely throw errors, when applied to the rundown - e.g. adding a part that is already there from the segment update.

One of the trickiest parts regarding handling ingest events from Sofie Core is to properly manage changes for on air segments and parts. The way changes to an on air segment is handled is by allowing all changes, since they are metadata changes.
Deleting an on air part results in an unsynced on air part, and an update results in an unsynced on air part, followed by a synced version of the part that will receive future updates.

To determine whether a part is updated or not, `EntityChangeDetector` is used, where all relevant attributes are compared. Complex objects are stringified and key-value pairs with the key `id` is removed. The removal of `id` attributes is because random ids were generated for sisyfos layers for each ingest, so a fixed amount of parts would always be detected as changed.


src/model/entities/test/entity-test-factory.ts
```diff
@@ -110,7 +142,7 @@ export class EntityTestFactory {
       name: 'pieceName',
       start: 0,
       pieceLifespan: PieceLifespan.WITHIN_PART,
-      isPlanned: false,
+      isPlanned: true,
```
LindvedKrvang, line 145: Remember to dobbeltcheck that this doesn't break any existing tests

KvelaGorrrrnio: What should I test besides running the tests? :)

src/business-logic/services/ingest-data-change-service.ts
```diff
@@ -0,0 +1,297 @@
...
+        originalSegmentId,
+      }
+    })
+    rundownSynchronizeResult.createdParts.forEach(part => rundown.addPart(part))
+    rundownSynchronizeResult.updatedParts.forEach(part => rundown.updatePart(part))
+
+    return { deletedPartsInfo, deletedSegmentsInfo }
+  }
+
+  private emitEventsFromRundownSynchronizeResult(rundown: Rundown, rundownSynchronizeResult: RundownSynchronizeResult, deletedSegmentsInfo: readonly DeletedSegmentInfo[], deletedPartsInfo: readonly DeletedPartInfo[]): void {
```
RasmusAlbrektsen: These parameter are not nice to read :-D 
Maybe we should consider a different way of showing them and add it to linting?

RasmusAlbrektsen: Generally the line length we allow is a bit much.

src/model/entities/part.ts
```diff
@@ -204,30 +204,28 @@ export class Part {
...
-      return
-    }
-    const removedPieces: Piece[] = this.pieces.splice(indexOfExistingPieceOnLayer, 1)
-    removedPieces.forEach(piece => {
-      if (!piece.isPlanned) {
-        return
-      }
-      this.replacedPlannedPieces.push(piece)
-    })
+  private doesPiecesOverlap(pieceA: Piece, pieceB: Piece): boolean {
```
RasmusAlbrektsen: Personal opinion: 
I'm not a fan of how `does` is making it in when it isn't grammatically correct, in this case it would be `do`. I understand that we are used to doing certain things when naming boolean return functions.

RasmusAlbrektsen (approved):

src/model/entities/test/entity-test-factory.ts
LindvedKrvang: image

src/business-logic/services/ingest-data-change-service.ts
```diff
@@ -221,34 +233,51 @@ export class IngestDataChangeService implements DataChangeService {
...
   }
 
-  private applyRundownSynchronizeResult(rundown: Rundown, rundownSynchronizeResult: RundownSynchronizeResult): DeletedSegmentInfo[] {
-    rundownSynchronizeResult.createdSegments.forEach(segment => rundown.addSegment(segment))
-    rundownSynchronizeResult.updatedSegments.forEach(segment => rundown.updateSegment(segment))
-    rundownSynchronizeResult.createdParts.forEach(part => rundown.addPart(part))
-    rundownSynchronizeResult.updatedParts.forEach(part => rundown.updatePart(part))
-    rundownSynchronizeResult.deletedParts.forEach(part => rundown.removePartFromSegment(part.id))
-    return rundownSynchronizeResult.deletedSegments.map(segment => {
+  private applyRundownSynchronizeResult(rundown: Rundown, rundownSynchronizeResult: RundownSynchronizeResult): { deletedPartsInfo: DeletedPartInfo[], deletedSegmentsInfo: DeletedSegmentInfo[] } {
```
LindvedKrvang: You have a nice `DeletedInfo` type that you can use here :)

src/business-logic/services/ingest-data-change-service.ts
LindvedKrvang: The last two parameters can be combine in the type `DeletedInfo`.

src/business-logic/services/ingest-rundown-synchronizer.ts
```diff
@@ -23,23 +24,52 @@ export class IngestRundownSynchronizer {
...
       .map(ingestedSegment => this.ingestedEntityToEntityMapper.convertIngestedSegmentToSegment(ingestedSegment))
+  }
 
-    const createdSegmentIds: ReadonlySet<string> = new Set(createdSegments.map(segment => segment.id))
-    const updatedSegments: readonly Segment[] = ingestedRundown.ingestedSegments.reduce<readonly Segment[]>((updatedSegments, ingestedSegment) => {
-      if (createdSegmentIds.has(ingestedSegment.id)) {
-        return updatedSegments
-      }
-      const segment: Segment | undefined = rundown.getSegments().find(segment => segment.id === ingestedSegment.id)
+  public getUpdatedSegments(segments: readonly Segment[], ingestedSegments: readonly IngestedSegment[]): readonly Segment[] {
```
LindvedKrvang: Should be private

src/business-logic/services/ingest-rundown-synchronizer.ts
```diff
@@ -51,48 +81,65 @@ export class IngestRundownSynchronizer {
...
+  }
 
-    const parts: readonly Part[] = rundown.getSegments().flatMap(segment => segment.getParts())
-
-    const ingestedParts: readonly IngestedPart[] = ingestedRundown.ingestedSegments.flatMap(ingestedSegment => ingestedSegment.ingestedParts)
-    const ingestedPartIds: ReadonlySet<string> = new Set(ingestedParts.map(ingestedPart => ingestedPart.id))
-
-    const deletedSegmentIds: ReadonlySet<string> = new Set(deletedSegments.map(segment => segment.id))
-    const deletedParts: readonly Part[] = parts.filter(part => part.isPlanned && !part.isUnsynced() && !ingestedPartIds.has(part.id) && !deletedSegmentIds.has(part.getSegmentId()))
+  public getDeletedSegments(segments: readonly Segment[], ingestedSegments: readonly IngestedSegment[]): readonly Segment[] {
```
LindvedKrvang: Should be private

src/model/entities/part.ts
```diff
@@ -204,30 +204,28 @@ export class Part {
       unPlannedPiece.setStart(timeSincePutOnAir)
       unPlannedPiece.markAsInsertedOnAir()
 
-      this.stopPiecesOnLayer(unPlannedPiece.layer)
+      this.stopOverlappingPiecesOnSameLayer(unPlannedPiece)
     } else {
-      this.removePieceOnLayer(unPlannedPiece.layer)
+      this.removeOverlappingPiecesOnSameLayer(unPlannedPiece)
```
LindvedKrvang, line 209: I'm quite sure this is going to have unintended side-effects i.e. wrong Piece behaviour. What does this have to do with the synchronize ingest flow?

src/model/entities/rundown.ts
```diff
@@ -693,6 +698,9 @@ export class Rundown extends BasicRundown {
       throw new NotFoundException(`Unable to find segment for part with id '${partId}' in rundown ${this.id}.`)
     }
     const removedPart: Part | undefined = segment.removePart(partId)
+    if (removedPart?.isOnAir()) {
+      this.activeCursor = this.createCursor(this.activeCursor, { part: removedPart })
+    }
```
LindvedKrvang, line 703: Is this to make the Unsynced copi of the OnAir Part be the one the Cursor is referencing? I.e. the update you did to `Segment.removePart()`?

LindvedKrvang (changes requested):
The request changes is only for making sure this PR isn't merged until we are certain that your `removeOverlappingPiecesOnSameLayer()` doesn't create unintended sideeffects

LindvedKrvang (approved):

src/business-logic/services/ingest-data-change-service.ts
KvelaGorrrrnio: I have used `DeletedInfo` instead.

src/model/entities/part.ts
KvelaGorrrrnio: I agree. I have updated it to `doPiecesOverlap`.

src/business-logic/services/ingest-data-change-service.ts
KvelaGorrrrnio: Updated :)

src/business-logic/services/ingest-rundown-synchronizer.ts
KvelaGorrrrnio: Updated :)

src/business-logic/services/ingest-rundown-synchronizer.ts
KvelaGorrrrnio: Updated :)

src/model/entities/part.ts
KvelaGorrrrnio: I don't see any current use cases where this will have any problems. But it is worth having in mind going forward :+1:

src/model/entities/rundown.ts
KvelaGorrrrnio: It is :)

LindvedKrvang (approved):

RasmusAlbrektsen (approved):
LGTM

PR #248, opened 2024-10-09
Introduce new SetNext endpoint

LindvedKrvang:
Introduce a new endpoint for setting Parts as Next.
This endpoint does not take the id of the Part that needs to be set as Next. Instead it takes a "direction". The "direction" could be `SEGMENT_BEFORE_NEXT_SEGMENT` or `PART_AFTER_NEXT_PART` etc.

The purpose of this endpoint is to allow Clients to be able to set a new Part as next, without being forced to listen to all Rundown and Ingest events, in order to find the correct Part for the given time. A Client interested in this could be a MixerPanel, where we simply want to set the next Part as next. Now this responsibility is in the backend.


src/model/entities/rundown.ts
```diff
@@ -543,6 +546,151 @@ export class Rundown extends BasicRundown {
...
+    this.assertNotUndefined(this.nextCursor, 'next Cursor')
+    const currentNextSegment: Segment = this.nextCursor?.segment
+
+    const currentNextSegmentIndex: number = this.segments.findIndex(segment => segment.id === currentNextSegment.id)
+    this.setFirstPartInSegmentAfterSegmentIndexAsNext(currentNextSegmentIndex, owner)
+  }
+
+  private setFirstPartInSegmentAfterSegmentIndexAsNext(segmentIndex: number, owner?: Owner): void {
+    if (segmentIndex === this.segments.length - 1) {
+      throw new LastSegmentInRundownException('')
```
MadsThy: empty string? :D

You have a few of them. Don't know if its intended, just wanted to point at them :)

src/model/entities/rundown.ts
```diff
@@ -543,6 +546,151 @@ export class Rundown extends BasicRundown {
...
+      }
+      case SetNextDirection.SEGMENT_BEFORE_NEXT_SEGMENT: {
+        this.setFirstPartInSegmentBeforeNextSegmentAsNext(owner)
+        break
+      }
+      case SetNextDirection.SEGMENT_AFTER_NEXT_SEGMENT: {
+        this.setFirstPartInSegmentAfterNextSegmentAsNext(owner)
+        break
+      }
+    }
```
MadsThy:
```suggestion
    switch (direction) {
      case SetNextDirection.PART_AFTER_NEXT_PART: {
        return this.setPartAfterCurrentNextPartAsNext(owner)
      }
      case SetNextDirection.PART_BEFORE_NEXT_PART: {
        return this.setPartBeforeCurrentNextPartAsNext(owner)
      }
      case SetNextDirection.SEGMENT_BEFORE_NEXT_SEGMENT: {
        return this.setFirstPartInSegmentBeforeNextSegmentAsNext(owner)
      }
      case SetNextDirection.SEGMENT_AFTER_NEXT_SEGMENT: {
        return this.setFirstPartInSegmentAfterNextSegmentAsNext(owner)
      }
    }
```

Don't ask me why I prefer switch cases with returns, but I just do :D Just as a consideration for you of course :)

RasmusAlbrektsen (approved):

MadsThy (approved):

src/model/entities/rundown.ts
KvelaGorrrrnio: Having `return` on same line as the method calls, might give the impression that the methods returns some sort of result - even though these don't :)

LindvedKrvang: They look nicer ;)

`return` in front of a method indicates that something is returned. Since all these methods return void I find it misleading to put a return statement in front of them.

src/model/entities/rundown.ts
LindvedKrvang: Good catch.
I think I didn't write a message partly because the branch was still POC at the time and partly because I didn't knew what to write that I felt wasn't already conveyed in the Exception name.

I have added messages to the Exceptions :)

src/model/entities/rundown.ts
```diff
@@ -543,6 +546,151 @@ export class Rundown extends BasicRundown {
...
+  }
+
+  private setPartAfterCurrentNextPartAsNext(owner?: Owner): void {
+    this.assertNotUndefined(this.nextCursor, 'next Cursor')
+    const currentNextSegment: Segment = this.nextCursor?.segment
+    const currentNextPart: Part = this.nextCursor?.part
```
KvelaGorrrrnio: Since the next cursor is not undefined, is the `?`s necessary?

src/model/entities/rundown.ts
```diff
@@ -510,8 +513,8 @@ export class Rundown extends BasicRundown {
     })
   }
 
-  public setNext(segmentId: string, partId: string, owner?: Owner): void {
-    this.assertActive(this.setNext.name)
+  public setNextFromIds(segmentId: string, partId: string, owner?: Owner): void {
```
KvelaGorrrrnio, line 512: Is this name change necessary?

src/model/entities/segment.ts
```diff
@@ -127,18 +147,36 @@ export class Segment {
...
     }
-    const nextPart: Part | undefined = this.parts.slice(fromPartIndex + 1).find(part => !part.invalidity)
+    const nextPart: Part | undefined = this.parts.slice(fromPartIndex + 1).find(part => !part.invalidity && !part.isOnAir())
     if (!nextPart) {
       throw new LastPartInSegmentException(`The part "${fromPart.name}" with id "${fromPart.id}" is the last part in the segment "${this.name}" with id "${this.id}".`)
     }
     return nextPart
   }
 
+  public findPreviousPartNotOnAir(fromPart: Part): Part {
```
KvelaGorrrrnio: Just a note: It is the previous off air part that is valid :)

src/model/entities/rundown.ts
```diff
@@ -543,6 +546,151 @@ export class Rundown extends BasicRundown {
...
+        break
+      }
+      case SetNextDirection.SEGMENT_BEFORE_NEXT_SEGMENT: {
+        this.setFirstPartInSegmentBeforeNextSegmentAsNext(owner)
+        break
+      }
+      case SetNextDirection.SEGMENT_AFTER_NEXT_SEGMENT: {
+        this.setFirstPartInSegmentAfterNextSegmentAsNext(owner)
+        break
+      }
```
KvelaGorrrrnio: Separating the analytical phase from the execution, would make the functionality more flexible. So this method could be something like:
```typescript
  public setNextFromDirection(direction: SetNextDirection, owner?: Owner): void {
    this.assertActive(this.setNextFromDirection.name)
    const nextCursor = this.findNextCursorFromDirection(direction)
    this.setNext(nextCursor.segment.id, nextCursor.part.id, owner)
  }

**Review by KvelaGorrrrnio — COMMENTED** [review]
Overall looks good. I would still prefer if this was a analytical-only feature, since delay in network or in the server, can have the user press 2 times, putting the user in an unintended position.

File: src/model/entities/rundown.ts (hunk 4 above)
Comment (LindvedKrvang) [reply src/model/entities/rundown.ts:512]: Depends on your philosophy.
There are two different ways to set the next. Which one will `setNext` utilize?
Ideally, both would just be called `setNext` and we would use method overloading, but TypeScript doesn't support that (in the extend that is needed here)

File: src/model/entities/segment.ts (hunk 5 above)
Comment (LindvedKrvang) [reply src/model/entities/segment.ts]: That's what it says

File: src/model/entities/rundown.ts (hunk 6 above)
Comment (LindvedKrvang) [reply src/model/entities/rundown.ts]: Updated

File: src/model/entities/rundown.ts (hunk 3 above)
Comment (LindvedKrvang) [reply src/model/entities/rundown.ts]: Not really. They have been removed.

File: src/model/entities/rundown.ts [hunk 7]
```diff
@@ -543,6 +546,159 @@ export class Rundown extends BasicRundown {
...
+  }
+
+  private findCursorWithPartAfterCurrentNextPart(): RundownCursor {
+    this.assertNotUndefined(this.nextCursor, 'next Cursor')
+    const currentNextSegment: Segment = this.nextCursor.segment
+    const currentNextPart: Part = this.nextCursor.part
+    try {
+      const nextPart: Part = currentNextSegment.findNextPartNotOnAir(currentNextPart)
+      return this.createCursorFromSegmentAndPart(currentNextSegment, nextPart)
+    }  catch (exception) {
```
MadsThy:
```suggestion
    } catch (exception) {
```

There's something with the formatting not having been run on this file :D It's not a problem, just wanted to say that there are formatter things that could be applied :)

MadsThy (approved):

src/model/entities/rundown.ts
LindvedKrvang: I can confidently say that formatting has been run. I run `yarn test` and `yarn lint-fix` before every commit ;) 
As to why the formatting didn't catch this, I don't know...
(Just did a `yarn lint-fix` again on this branch to very it didn't catch it)

src/model/entities/segment.ts
KvelaGorrrrnio: My point is that it does not say anything about validity :)

src/model/entities/rundown.ts
KvelaGorrrrnio: I think my philosophy for this is that we have one setNext method that sets the next cursor, which is plentiful for almost any use case we have. The introduction of finding the next/previous part/segment from a direction, is separate from that of setting the next cursor. As I see it, `setNextFromDirection` is a method for convenience for a use case that lies beyond the responsibilities of the rundown entity - e.g. in a service or directly in the controller (if no appropriate service lies between).

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -247,9 +248,20 @@ export class RundownTimelineService implements RundownService {
...
+    rundown.setNextFromIds(segmentId, partId, owner)
+
+    await this.buildAndPersistTimeline(rundown)
+
+    this.rundownEventEmitter.emitSetNextEvent(rundown)
+
+    await this.saveRundown(rundown)
+  }
+
+  public async setNext(rundownId: string, direction: SetNextDirection, owner?: Owner): Promise<void> {
```
KvelaGorrrrnio: It is confusing that the `setNext` name is given another purpose than what it had. An alternative is to have this renamed to `setNextFromDirection` instead.

src/model/entities/rundown.ts
```diff
@@ -543,6 +546,159 @@ export class Rundown extends BasicRundown {
...
+        throw error
+      }
+      return this.findCursorWithFirstPartInSegmentAfterSegmentIndex(segmentIndex + 1)
+    }
+  }
+
+  private findFirstValidSegmentAfterIndex(searchIndex: number): Segment {
+    for (let i: number = searchIndex + 1; i < this.segments.length; i++) {
+      const segment: Segment = this.segments[i]
+      if (!segment.invalidity && segment.getParts().some(part => !part.invalidity)) {
```
KvelaGorrrrnio: Checking the validity of a segment is done a couple of times. It could be an idea to extract it to a method like `isValidSegment`.

src/model/entities/rundown.ts
LindvedKrvang: But both sets the Next cursor. Only difference is how they do it.

You can flip it around and say that we have one SetNext method that sets the Next from a direction, since the responsibility of finding the Next lies within the Rundown. It's a convenience that we have a method that simply takes two ids and set the Next from them.

I would argue, that since the `setNext(segmentId, partId)` requires that the client knows about the current Rundown state and is aware of all ingest changes that the `setNext(segmentId, partId)` should not be the primary way of setting the Next cursor. Instead `setNext(direction)` should be the primary way, where `setNext(segmentId, partId)` is a convenience when we _already_ know which Segment/Part to set as next (for example right-clicking a Part in the GUI and selecting it as Next).

Just because `setNext(segmentId, partId)` was created before `setNext(direction)` does not mean that it would be the default/primary way of doing a thing.

But the ideal solution here would be if TypeScript supported proper method overloading :p

src/business-logic/services/rundown-timeline-service.ts
LindvedKrvang: It's still the same purpose: To set the Next. Is just how it does it that has changed.

src/model/entities/rundown.ts
LindvedKrvang: I have updated Segment with a `isValid()` method and replaced the checks for validity in `rundonw.ts` with it.

src/model/entities/segment.ts
LindvedKrvang: Ahh missed that. Updated to `findPreviousValidPartNotOnAir`

src/model/entities/rundown.ts
LindvedKrvang: Nevertheless, the space is now gone :)

src/model/entities/rundown.ts
KvelaGorrrrnio: I disagree with the point that `setNextFromIds` is or should be the convenience method. `setNextFromDirection` uses `setNextFromIds` internally - so it builds upon it. Setting the next cursor given the segment and part is the core functionality of setting the next cursor.

I will argue that how a user places the next cursor is not the rundowns responsibility. The responsibility of the set next functionality for the rundown is to set the next cursor (if valid) - not finding out what to set as next, with the exceptions being when the next cursor needs to be updated when doing takes and updating from ingest changes. But the exceptions are for keeping the state consistent, not for user navigational needs.

At the moment we have two ways of setting the next cursor from a client perspective. One is setting next cursor based on a specific part (set next with mouse) and the other is setting the next based on a direction (set next with arrow keys). These are both client needs and not something that the server needs to know about. In the future, e.g. when we support multiple show style variants in a rundown, a navigational pattern could be to jump between `KLAR ON AIR` segments. Then this is something that the rundown should handle too. The rundown file is quite big already, and will become more bloated the more responsibilities we put on it.

The order in which they were created does not tell which are the primary/secondary core/convenience - but the logic and the responsibilities they carry does :)

Method overloading can be done, but as you mention it is a bit cumbersome 😁 For me it is not method separation, but rather what the core logic is for setting the next cursor and which responsibilities the rundown should have.

src/business-logic/services/rundown-timeline-service.ts
KvelaGorrrrnio: I will still like that this is more expressive about what it does. Setting next from a direction is not the core functionality of setting next on a rundown. So if the core functionality needs a more descriptive name, so does this :)

LindvedKrvang: Setting Next from a Direction should be the primary way of setting Next.
This way, we don't force clients a _lot_ of extra work just to be able to set the Next.

LindvedKrvang: I have renamed it to `SetNextFromDirection`

KvelaGorrrrnio (approved):
I will leave a link to https://github.com/tv2/sofie-server/pull/248#discussion_r1802724374 here. I am not a fan of it living in the Rundown. But that is something that we should discuss outside of this PR as it also ties into the thoughts of where this should live in general.

PR #249, opened 2024-10-10
SOF-2298: Make repository transactions conform to ACID

KvelaGorrrrnio:
_**Note:** Since this is a major change to how the system operates, I encourage reviewers to have extra focus on the details in the PR - especially in the repository files._

This PR does 2 major things.

One is to make the Rundown, Segment, Part and Piece repositories to an aggregate structure, where everything goes through the Rundown repository. This is done to increase consistency between in-memory and database state, e.g. by allowing the cache decorator of the rundown aggregate repository to update the cache, even if a method call was on e.g, segment level. It also enables the Rundown aggregate repository to do optimizations like bulk writes for larger operations - which is also done in this PR.

The second major change is to make compound commands to the database for the Rundown hierarchy transactional. So if a portion of the command fails, then the entire thing is rolled back and reported as an error.
In conjunction with this, bulk writes is used to optimize the execution speed. In doing so, most methods in the segment, part, and piece repositories build mongo queries, instead of executing them. They can then be bundled and executed with the method `executeQueries`.

Additional changes made in the PR are:
- The cache decorators for segment and part repositories are removed, due to them being unused and the new rundown aggregate repository structure.
- The model type for a repository is moved from the `getCollection` to the class declaration, to fixate it for the entire class. This ensures that whenever `getCollection` is used, the correct model type is used instead of the reduced default document type `MongId`.
- Removes the `InMemoryMongo` database integration tests from the unit test suite. If they are needed, then we should implement them as integration-tests, e.g. with [Testcontainers](https://testcontainers.com/).
- Saving a rundown also removes orphaned content from that rundown in the database.
- Unplanned pieces are given unique ids.
- Rundown id is set on unplanned parts.


RasmusAlbrektsen (commented):
Gotta say, i like it when you do ACID.

src/data-access/repositories/interfaces/piece-repository.ts
```diff
@@ -1,12 +1 @@
...
-  getPieces(partId: string): Promise<Piece[]>
-  getPiecesFromIds(pieceIds: string[]): Promise<Piece[]>
-  savePiece(piece: Piece): Promise<void>
-  deletePiecesForPart(partId: string): Promise<void>
-  deletePieces(pieceIds: string[]): Promise<void>
-  deleteUnsyncedInfinitePiecesNotOnAnyRundown(): Promise<void>
-  deleteAllUnsyncedPieces(): Promise<void>
-  deleteAllUnplannedPieces(): Promise<void>
-}
+export interface PieceRepository {}
```
MadsThy: You cleaned house? :P

src/data-access/repositories/interfaces/rundown-aggregate-repository.ts
```diff
@@ -0,0 +1,6 @@
+import { RundownRepository } from './rundown-repository'
+import { SegmentRepository } from './segment-repository'
+import { PartRepository } from './part-repository'
+import { PieceRepository } from './piece-repository'
+
+export interface RundownAggregateRepository extends RundownRepository, SegmentRepository, PartRepository, PieceRepository {}
```
MadsThy: Why are you making so many empty types? xD

src/data-access/repositories/mongo/mongo-piece-repository.ts
```diff
@@ -19,48 +20,70 @@ export class MongoPieceRepository extends BaseMongoRepository implements PieceRe
...
+  public getPiecesFromIds(pieceIds: string[] = []): Promise<Piece[]> {
     this.assertDatabaseConnection(this.getPiecesFromIds.name)
-    const mongoPieces: MongoPiece[] = await this.getCollection().find<MongoPiece>(
-      {
-        _id: { $in: pieceIds }
-      }
-    ).toArray()
-    return this.mongoEntityConverter.convertToPieces(mongoPieces)
+    return this.getCollection()
+      .find<MongoPiece>({_id: { $in: pieceIds } })
```
MadsThy:
```suggestion
      .find<MongoPiece>({ _id: { $in: pieceIds } })
```

Formatting :D

MadsThy (approved):

src/business-logic/services/execute-action-service.ts
```diff
@@ -150,17 +150,19 @@ export class ExecuteActionService implements ActionService {
   }
 
   private async insertPartAsOnAir(partAction: PartAction, rundownId: string): Promise<void> {
-    const part: Part = this.createPartFromAction(partAction)
+    const part: Part = this.createPartFromAction(partAction, rundownId)
     await this.rundownService.insertPartAsOnAir(rundownId, part)
   }
 
-  private createPartFromAction(partAction: PartAction): Part {
+  private createPartFromAction(partAction: PartAction, rundownId: string): Part {
```
LindvedKrvang, line 157: FYI: `PartAction` does have a `rundownId` already. However, only "rundown-specific" Actions have it, so it might not be useable for this usecase (assuming that the usecase here is to provide the RundownId of the Rundown the Part is being inserted into).

src/data-access/facades/repository-facade.ts
```diff
@@ -82,13 +81,18 @@ export class RepositoryFacade {
   }
 
   public static createRundownRepository(): RundownRepository {
-    const mongoRundownRepository: RundownRepository = new MongoRundownRepository(
+    return this.createRundownAggregateRepository()
+  }
+
+  public static createRundownAggregateRepository(): RundownAggregateRepository {
```
LindvedKrvang: Should be private

src/data-access/repositories/cache/cached-rundown-aggregate-repository.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.cachedRundowns.delete(rundownId)
+  }
+
+  public getSegment(segmentId: string): Promise<Segment> {
+    return this.rundownAggregateRepository.getSegment(segmentId)
+  }
+
+  public getPart(partId: string): Promise<Part> {
+    return this.rundownAggregateRepository.getPart(partId)
+  }
```
LindvedKrvang, line 54: We are not using the cache when fetching Segments and Parts?

src/data-access/repositories/interfaces/piece-repository.ts
LindvedKrvang: If it's empty, why do we still have it?

src/data-access/repositories/mongo/mongo-action-trigger-repository.ts
```diff
@@ -4,10 +4,11 @@ import { ActionTrigger } from '../../../model/entities/action-trigger'
 import { MongoDatabase } from './mongo-database'
 import { UuidGenerator } from '../interfaces/uuid-generator'
 import { NotFoundException } from '../../../model/exceptions/not-found-exception'
+import { MongoId } from './mongo-entity-converter'
 
 const ACTION_TRIGGER_COLLECTION: string = 'actionTriggers'
 
-export class MongoActionTriggerRepository extends BaseMongoRepository implements ActionTriggerRepository {
+export class MongoActionTriggerRepository extends BaseMongoRepository<ActionTrigger & MongoId> implements ActionTriggerRepository {
```
LindvedKrvang, line 11: One of the reasons why the generic was on `getCollection` and not on the class of `BaseMongoRepository` was exactly for this reason that we didn't need to add `MongoId` on things where we don't need/use the MongoId syntax.

src/data-access/repositories/mongo/mongo-ad-lib-actions-repository.ts
```diff
@@ -6,6 +6,7 @@ import { BaseMongoRepository } from './base-mongo-repository'
 const AD_LIB_ACTIONS_COLLECTION: string = 'adLibActions'
 
 interface MongoAdLibAction {
+  _id: string
```
LindvedKrvang, line 9: I'm quite sad to see that this is added when we don't use it.

src/data-access/repositories/mongo/mongo-ad-lib-piece-repository.ts
```diff
@@ -8,6 +8,7 @@ import { MongoId } from './mongo-entity-converter'
 const AD_LIB_PIECES_COLLECTION: string = 'adLibPieces'
 
 interface MongoAdLibPiece {
+  _id: string
```
LindvedKrvang, line 11: Same comment as for `MongoAdLibAction`

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -134,6 +136,7 @@ export interface MongoTimeline extends MongoId {
 }
 
 export interface MongoStudio {
+  _id: string
```
LindvedKrvang, line 139: Same comment about adding this id when it is not used. Goes for all the places where this happened in this PR

src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -177,6 +181,18 @@ export interface MongoSystemInformation extends MongoId {
   name: string
 }
 
+export interface MongoAction extends MongoId {
```
LindvedKrvang, line 184: A big reason for the `MongoX` interfaces is because Core saves entities in certains ways that we don't want our entities to have.

`Action` is not something that comes from Core so we don't need a `MongoAction` interface. We can simply use our `Action` interfacej.

Also, our Actions doesn't have the `_id` so it shouldn't extend `MongoId`.

src/data-access/repositories/mongo/mongo-part-repository.ts
```diff
@@ -1,21 +1,22 @@
...
 import { NotFoundException } from '../../../model/exceptions/not-found-exception'
 import { Piece } from '../../../model/entities/piece'
 import { MongoEntityConverter, MongoPart } from './mongo-entity-converter'
+import { MongoPieceRepository } from './mongo-piece-repository'
 
-export const PART_COLLECTION_NAME: string = 'executedParts' // TODO: Once we control ingest rename to "parts".
+const PART_COLLECTION_NAME: string = 'executedParts' // TODO: Once we control ingest rename to "parts".
 
-export class MongoPartRepository extends BaseMongoRepository implements PartRepository {
+export class MongoPartRepository extends BaseMongoRepository<MongoPart> {
```
LindvedKrvang, line 15: Why does this no longer extend `PartRepository`?
What's the point of having the abstraction if the implementation don't use it?

src/data-access/repositories/mongo/mongo-rundown-aggregate-repository.ts
```diff
@@ -0,0 +1,110 @@
...
+const RUNDOWN_COLLECTION_NAME: string = 'executedRundowns' // TODO: Once we control ingest renamed this to "rundowns".
+
+export class MongoRundownAggregateRepository extends BaseMongoRepository<MongoRundown> implements RundownAggregateRepository {
+
+  constructor(
+    mongoDatabase: MongoDatabase,
+    private readonly mongoSegmentRepository: MongoSegmentRepository,
+    private readonly mongoPartRepository: MongoPartRepository,
+    private readonly mongoPieceRepository: MongoPieceRepository,
+    private readonly mongoEntityConverter: MongoEntityConverter,
```
LindvedKrvang, line 25: We should depend on abstractions, not implementations.

src/data-access/repositories/mongo/mongo-part-repository.ts
```diff
@@ -1,21 +1,22 @@
...
   constructor(
     mongoDatabase: MongoDatabase,
+    private readonly mongoPieceRepository: MongoPieceRepository,
```
LindvedKrvang, line 18: You should depend on the abstraction, not the implementation.

src/data-access/repositories/mongo/mongo-part-repository.ts
```diff
@@ -46,74 +47,103 @@ export class MongoPartRepository extends BaseMongoRepository implements PartRepo
...
-    await Promise.all(part.getPieces().map(piece => this.pieceRepository.savePiece(piece)))
+  public getPartIdsForRundown(rundownId: string): Promise<readonly string[]> {
+    return this.getCollection().find({ rundownId }).map(document => document._id).toArray()
   }
 
-  public async delete(partId: string): Promise<void> {
-    this.assertDatabaseConnection(this.delete.name)
-    await this.pieceRepository.deletePiecesForPart(partId)
-    await this.getCollection().deleteMany({ _id: partId })
+  public getPartIdsForSegment(segmentId: string, mongoPart: Partial<MongoPart> = {}): Promise<readonly string[]> {
```
LindvedKrvang: This method isn't being used.

src/data-access/repositories/mongo/mongo-part-repository.ts
```diff
@@ -46,74 +47,103 @@ export class MongoPartRepository extends BaseMongoRepository implements PartRepo
...
+      return
     }
+    await this.getCollection().bulkWrite([...queries], { session, ignoreUndefined: true })
   }
 
-  public async deleteUnsyncedPartsForSegment(segmentId: string): Promise<void> {
-    this.assertDatabaseConnection(this.deleteUnsyncedPartsForSegment.name)
-    const unsyncedFilter: Partial<MongoPart> = { isUnsynced: true }
-    const unsyncedParts: Part[] = await this.getParts(segmentId, unsyncedFilter)
+  public buildDeletePartsForRundownQuery(rundownId: string): AnyBulkWriteOperation<MongoPart> {
```
LindvedKrvang, line 90: This method isn't used

src/data-access/repositories/mongo/mongo-part-repository.ts
```diff
@@ -46,74 +47,103 @@ export class MongoPartRepository extends BaseMongoRepository implements PartRepo
...
+    return {
+      deleteMany: {
+        filter: { rundownId },
+      },
+    }
+  }
 
-    await Promise.all(unsyncedParts.map(async (part) => this.pieceRepository.deletePiecesForPart(part.id)))
+  public buildDeletePartsForSegmentQuery(segmentId: string, mongoPart: Partial<MongoPart> = {}): AnyBulkWriteOperation<MongoPart> {
```
LindvedKrvang: This method isn't used

src/data-access/repositories/mongo/mongo-part-repository.ts
```diff
@@ -46,74 +47,103 @@ export class MongoPartRepository extends BaseMongoRepository implements PartRepo
...
+    return {
+      deleteMany: {
+        filter: { ...mongoPart, segmentId },
+      },
+    }
+  }
 
-    const partsDeletedResult: DeleteResult = await this.getCollection().deleteMany({ ...unsyncedFilter, segmentId: segmentId })
+  public buildDeletePartQuery(partId: string): AnyBulkWriteOperation<MongoPart> {
```
LindvedKrvang: This method isn't used

src/data-access/repositories/mongo/mongo-part-repository.ts
```diff
@@ -46,74 +47,103 @@ export class MongoPartRepository extends BaseMongoRepository implements PartRepo
...
+    return {
+      deleteOne: {
+        filter: { _id: partId }
+      }
+    }
+  }
 
-    if (!partsDeletedResult.acknowledged) {
-      throw new DeletionFailedException(`Deletion of parts was not acknowledged, for segmentId: ${segmentId}`)
+  public buildDeleteUnsyncedPartsForSegmentQuery(segmentId: string): AnyBulkWriteOperation<MongoPart> {
```
LindvedKrvang: This method isn't used

src/data-access/repositories/mongo/mongo-part-repository.ts
```diff
@@ -46,74 +47,103 @@ export class MongoPartRepository extends BaseMongoRepository implements PartRepo
...
-    await this.getCollection().deleteMany({
-      isPlanned: false
-    })
+  public buildDeleteAllUnplannedPartsQuery(): AnyBulkWriteOperation<MongoPart> {
+    return {
+      deleteMany: {
+        filter: { isPlanned: false },
+      }
+    }
   }
```
LindvedKrvang: These methods are not used

src/data-access/repositories/mongo/mongo-piece-repository.ts
```diff
@@ -1,15 +1,16 @@
...
+  ClientSession,
+} from 'mongodb'
 import { MongoEntityConverter, MongoId, MongoPiece } from './mongo-entity-converter'
 import { PieceLifespan } from '../../../model/enums/piece-lifespan'
 
-export const PIECE_COLLECTION_NAME: string = 'executedPieces' // TODO: Once we control ingest rename to "pieces".
+const PIECE_COLLECTION_NAME: string = 'executedPieces' // TODO: Once we control ingest rename to "pieces".
 
-export class MongoPieceRepository extends BaseMongoRepository implements PieceRepository {
+export class MongoPieceRepository extends BaseMongoRepository<MongoPiece> {
```
LindvedKrvang, line 12: What's the reason for this not implementing `PieceRepository`?

src/data-access/repositories/mongo/mongo-piece-repository.ts
```diff
@@ -19,48 +20,70 @@ export class MongoPieceRepository extends BaseMongoRepository implements PieceRe
...
-    await this.getCollection().deleteMany({ _id: { $in: pieceIds }})
+  public buildDeletePiecesForRundownQuery(partIds: readonly string[]): AnyBulkWriteOperation<MongoPiece> {
+    return {
+      deleteMany: {
+        filter: { partId: { $in: partIds } },
+      },
+    }
+  }
+
+  public buildDeletePiecesForPartQuery(partId: string): AnyBulkWriteOperation<MongoPiece> {
```
LindvedKrvang: This method isn't used

src/data-access/repositories/mongo/mongo-piece-repository.ts
```diff
@@ -19,48 +20,70 @@ export class MongoPieceRepository extends BaseMongoRepository implements PieceRe
...
+    return {
+      deleteOne: {
+        filter: { partId }
+      }
+    }
   }
 
   public async deleteUnsyncedInfinitePiecesNotOnAnyRundown(): Promise<void> {
+    this.assertDatabaseConnection(this.deleteUnsyncedInfinitePiecesNotOnAnyRundown.name)
```
LindvedKrvang: Can't comment directly on the method declaration since it didn't change :p
This method isn't used anymore

src/data-access/repositories/mongo/mongo-piece-repository.ts
```diff
@@ -80,27 +103,29 @@ export class MongoPieceRepository extends BaseMongoRepository implements PieceRe
...
-    await this.getCollection().deleteMany({
-      isPlanned: false
-    })
+  public buildDeleteAllUnplannedPiecesQuery(): AnyBulkWriteOperation<MongoPiece> {
+    return {
+      deleteMany: {
+        filter: { isPlanned: false }
+      }
+    }
   }
```
LindvedKrvang: These methods aren't used anymore

src/data-access/repositories/mongo/mongo-segment-repository.ts
```diff
@@ -1,22 +1,22 @@
...
 import { NotFoundException } from '../../../model/exceptions/not-found-exception'
 import { Part } from '../../../model/entities/part'
 import { MongoEntityConverter, MongoSegment } from './mongo-entity-converter'
+import { MongoPartRepository } from './mongo-part-repository'
 
-export const SEGMENT_COLLECTION_NAME: string = 'executedSegments' // TODO: Once we control ingest rename to "segments".
+const SEGMENT_COLLECTION_NAME: string = 'executedSegments' // TODO: Once we control ingest rename to "segments".
 
-export class MongoSegmentRepository extends BaseMongoRepository implements SegmentRepository {
+export class MongoSegmentRepository extends BaseMongoRepository<MongoSegment> {
```
LindvedKrvang, line 15: Same question as for the other repositories: Why does it not implement `SegmentRepository` and why are you depending on the concrete implementation in the constructor?

src/data-access/repositories/mongo/mongo-segment-repository.ts
```diff
@@ -34,75 +34,81 @@ export class MongoSegmentRepository extends BaseMongoRepository implements Segme
...
+      deleteMany: {
+        filter: { rundownId },
+      },
+    }
+  }
 
-    await Promise.all(segments.map(async (segment) => this.partRepository.deletePartsForSegment(segment.id)))
 
-    const segmentDeleteResult: DeleteResult = await this.getCollection().deleteMany({ ...unsyncedFilter, rundownId: rundownId })
+  public buildDeleteUnsyncedSegmentsForRundownQuery(rundownId: string): AnyBulkWriteOperation<MongoSegment> {
```
LindvedKrvang: This method isn't used

src/data-access/repositories/mongo/mongo-segment-repository.ts
```diff
@@ -34,75 +34,81 @@ export class MongoSegmentRepository extends BaseMongoRepository implements Segme
...
   /*
   * NOTE: This will delete ALL unsynced Segments in the database. Should only be used on deactivate or activate Rundown.
   * NOTE: This will NOT delete the associated Parts.
   */
-  public async deleteAllUnsyncedSegments(): Promise<void> {
-    this.assertDatabaseConnection(this.deleteAllUnsyncedSegments.name)
-    await this.getCollection().deleteMany({
-      isUnsynced: true
-    })
+  public buildDeleteAllUnsyncedSegmentsQuery(): AnyBulkWriteOperation<MongoSegment> {
```
LindvedKrvang: This method isn't used

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -180,8 +162,8 @@ export class RundownTimelineService implements RundownService {
     this.rundownEventEmitter.emitSetNextEvent(rundown)
     this.startAutoNext(timeline, rundown.id)
 
-    await this.deleteUnsyncedPreviousPart(rundown)
-    await this.deleteUnsyncedSegments(rundown)
+    this.emitDeleteUnsyncedPreviousPart(rundown)
+    this.deleteUnsyncedSegments(rundown)
     await this.saveRundown(rundown)
```
LindvedKrvang, line 167: A thing to note: 
Since `emitDeleteUnsyncedPreviousPart` is emitted before we try to save the Rundown; it's possible for us to to emit the delete event, fail the save of the Rundown, and then not actually delete the Part that we just emitted was deleted.


Another thing to note - the actual deletion unsynced Parts:
In these changes (the entire PR) we no longer make an "active effort" to delete unsynced Parts. They are still deleted, but only because the unsynced Part is removed from the Parts array of the Segment when we do `Rundown.takeNext()` (happens inside the private method `Rundown.takeNextCursor()`). Note, that we only ever try to remove unsynced Parts from a Segment when the Segment is in the _active_ cursor of the Rundown when starting the Take.

This may, or may not, lead to an egdecase where we have a Part that becomes unsynced, we do a Take away from it, the save Rundown flow _fails_ and we now have a rouge unsynced Part.

The above egdecase is just a thought scenario. I haven't checked if it's an actual issue. But I wanted to mention it :)

LindvedKrvang (commented):
Good work. A much needed improvement :D 

Before I can approve, I need to hear your reasoning for some of the choices you made. I have made comments where applicable.

Note: Not a fan of moving the placement of the generic in `BaseMongoRepositoy`. It was chosen on the method so we weren't force to put `MongoId` on a lot of our entities that simply doesn't use it (as you have added a lot of places in this PR).
Not saying we should move it back, but I would prefer that we avoid adding an attribute that we don't use.

src/data-access/facades/repository-facade.ts
KvelaGorrrrnio: It should and now is :)

src/data-access/repositories/mongo/mongo-segment-repository.ts
KvelaGorrrrnio: And it now no longer exists 💀

src/data-access/repositories/mongo/mongo-segment-repository.ts
KvelaGorrrrnio: And it now no longer exists 💀

src/data-access/repositories/mongo/mongo-piece-repository.ts
KvelaGorrrrnio: I really thought I had gone through the repositories and removed unused methods. But they are gone now :D

src/data-access/repositories/mongo/mongo-piece-repository.ts
KvelaGorrrrnio: Is now removed.

src/data-access/repositories/mongo/mongo-piece-repository.ts
KvelaGorrrrnio: Is now removed.

src/data-access/repositories/mongo/mongo-part-repository.ts
KvelaGorrrrnio: And they are gone!

src/data-access/repositories/mongo/mongo-part-repository.ts
KvelaGorrrrnio: It is no longer with us 👼

src/data-access/repositories/mongo/mongo-part-repository.ts
KvelaGorrrrnio: And it is time for it to go ⌛

src/data-access/repositories/mongo/mongo-part-repository.ts
KvelaGorrrrnio: image

src/data-access/repositories/mongo/mongo-part-repository.ts
KvelaGorrrrnio: It is now gone.

src/data-access/repositories/mongo/mongo-part-repository.ts
KvelaGorrrrnio: It is now gone.

src/business-logic/services/execute-action-service.ts
KvelaGorrrrnio: I know, but as you state, we want to have the rundown id on all parts that are inserted into a rundown :)

src/data-access/repositories/cache/cached-rundown-aggregate-repository.ts
KvelaGorrrrnio: We are not :) Currently, they are only used when we receive a delete X event from Sofie core via MongoDB to get the rundown to which they belong. We could search through all cached rundowns to try and find the segment or part before asking the database?

src/data-access/repositories/interfaces/piece-repository.ts
KvelaGorrrrnio: Good question! It is now removed :)

KvelaGorrrrnio: I did :) This interface is now removed :)

src/data-access/repositories/interfaces/rundown-aggregate-repository.ts
KvelaGorrrrnio: This is not an empty type, since it combines multiple interfaces. I have though changed it to a type declaration instead :)

src/data-access/repositories/mongo/mongo-piece-repository.ts
KvelaGorrrrnio: Formatting applied ✏️

RasmusAlbrektsen (approved):
Except for the unused methods i think it looks very good! But Lindved seems to have caught every one i saw.
I must agree with Lindved that i'm confused about us not depending on abstractions for the mongo repositories, i'd say we shouldn't change the approach unless we decide it together, just to avoid having multiple approaches in the same repository.

src/data-access/repositories/mongo/mongo-entity-converter.ts
KvelaGorrrrnio: The biggest reason for having the MongoX interfaces is being able to map from a specific database format (currently MongoDB) to the data format used by the application. All documents we fetch from the database has the `_id` attribute, so for actions (and other models) the `_id` attribute is implicitly passed along. I have updated the MongoActionRepository to make use of the entity converter.

And the action repository uses the `_id` for grabbing actions.

src/data-access/repositories/mongo/mongo-ad-lib-actions-repository.ts
KvelaGorrrrnio: That doesn't contribute towards the domains OKR 😢 Personally I don't see the issue with having the identifier on a model which has unique instances. It is the contract for working with MongoDB.

src/data-access/repositories/mongo/mongo-part-repository.ts
KvelaGorrrrnio: It never has extended the interface 😉 But the reason being that it no longer implements the interface is that it doesn't conform to the interface. It is the RundownAggregateRepository that conforms to the repository interfaces.

src/data-access/repositories/mongo/mongo-rundown-aggregate-repository.ts
KvelaGorrrrnio: In most cases yes, but the Mongo repositories for segments, parts and pieces are under the rundown aggregate repository and should not be used outside of the aggregate. The functionality that they provide are exposed via the repository interfaces through the RundownAggregateRepository.

src/data-access/repositories/mongo/mongo-part-repository.ts
KvelaGorrrrnio: See https://github.com/tv2/sofie-server/pull/249#discussion_r1802459321.

src/data-access/repositories/mongo/mongo-piece-repository.ts
KvelaGorrrrnio: See https://github.com/tv2/sofie-server/pull/249#discussion_r1802456012.

src/data-access/repositories/mongo/mongo-segment-repository.ts
KvelaGorrrrnio: See https://github.com/tv2/sofie-server/pull/249#discussion_r1802456012.

src/business-logic/services/rundown-timeline-service.ts
KvelaGorrrrnio: Nice observations! I agree with them. It is the same approach we have for other operations on the rundown, where we emit events, build and persist the timeline before we store the rundown. So I think if this should be called after saving the rundown, so should all other event emitting. But that is balancing perceived performance vs data consistency :)

src/data-access/repositories/mongo/mongo-action-trigger-repository.ts
KvelaGorrrrnio: By having it on the `getCollection` method, defining the type for what is saved to the database becomes optional. E.g. unless I as developer explicitly set up the type argument for the `getCollection` call, there is nothing stopping me from saving a Rundown to the ActionTrigger collection.
By having it on the class, no extra steps are needed from the developer and the type system helps with only saving valid data to the database.

src/data-access/repositories/mongo/mongo-ad-lib-piece-repository.ts
KvelaGorrrrnio: https://github.com/tv2/sofie-server/pull/249#discussion_r1802452972

src/data-access/repositories/mongo/mongo-entity-converter.ts
KvelaGorrrrnio: https://github.com/tv2/sofie-server/pull/249#discussion_r1802452972

src/data-access/repositories/mongo/mongo-ad-lib-actions-repository.ts
LindvedKrvang: Don't really know what OKR has to do with anything here..?

src/data-access/repositories/mongo/mongo-entity-converter.ts
LindvedKrvang: As we talk about I don't see the need for it, but I guess it's what we are going with for now ;)

src/data-access/repositories/mongo/mongo-piece-repository.ts
LindvedKrvang: Are you using WebStorm for that? 
Atleast for me, WebStorm changes the color of unused methods so they stand out.

src/business-logic/services/rundown-timeline-service.ts
LindvedKrvang: Agree. Leaving the comment unresolved for other people to also see it :)

src/data-access/repositories/interfaces/rundown-aggregate-repository.ts
```diff
@@ -1,6 +1,5 @@
 import { RundownRepository } from './rundown-repository'
 import { SegmentRepository } from './segment-repository'
 import { PartRepository } from './part-repository'
-import { PieceRepository } from './piece-repository'
 
-export interface RundownAggregateRepository extends RundownRepository, SegmentRepository, PartRepository, PieceRepository {}
+export type RundownAggregateRepository = RundownRepository & SegmentRepository & PartRepository
```
LindvedKrvang, line 5: Is it necessary to have `RundownAggregateRepository` be an actual type?
We couldn't just have?:
```
export class MongoRundownAggreaterepository implements RundownRepository, SegmentRepository, PartRepository {
  ...
}
```

LindvedKrvang (approved):

src/data-access/repositories/interfaces/rundown-aggregate-repository.ts
KvelaGorrrrnio: I introduced it to have one place to update it, e.g. if the piece repository interface needs to be added again. It is also used in the repository facade. If it needs to go, I suggest that it is done as boy-scouting :)

PR #250, opened 2024-10-10
fix: Makes device status messages more descriptive for frontend notifications

RasmusAlbrektsen:


src/business-logic/services/device-changed-service.ts
```diff
@@ -85,10 +85,20 @@ export class DeviceChangedService implements DataChangeService {
...
+  private getDeviceMessage(device: Device): string {
+    if (device.statusMessage){
+      return device.statusMessage
+    }
+    if (device.statusCode === StatusCode.GOOD) {
+      return 'The device is in a good state.'
+    }
+    return ''
+  }
+
```
MadsThy: Seems like the formatter hasn't ran on this :D There is inconsistency in where the space is after declaring the check :)

MadsThy (approved):

KvelaGorrrrnio (approved):
Looks good except for the import statement linting :D

src/business-logic/services/device-changed-service.ts
RasmusAlbrektsen: It isn't part of the linting rules so that's why formatting doesn't catch it. I keep forgetting that isn't enforced on this repository. Good catch i changed it.

LindvedKrvang (approved):

PR #253, opened 2024-10-14
SOF-2298: Lock access to RundownAggregateRepository.

KvelaGorrrrnio:
Adds a lock to the RundownAggregateRepository, to ensure that no concurrent control flows are competing about a rundown ressource.
This is done to avoid user requests going through controllers and ingest updates competing with each other about what state is correct.

Using a lock on database access has some trade-offs that should be weighed before merging this PR.
Locking the repository might cause a delay subsequent requests to the repository, so e.g. cutting to a camera might take longer than usual. By not having the lock, we might get in a wrong state, of which the likelihood grows linearly with the size of the rundown. We can get around the downside of having the lock to a degree by reducing the time that operations uses the lock - e.g. optimizing database queries.


LindvedKrvang (commented):
I must admit, this is not how I expected the lock to work. I thought the lock would be on the database only, i.e. repository level.

As you point out yourself, this implementation blocks execution of the _entire_ method in the lock.

The way I had imagined it was something similar to:
```
export class RepositoryFacade {
  public static createRundownRepository(): RundownRepository {
    return new CachedRundownRepository(
      new LockedRundownRepository(
        new MongoRundownRepository()
      )
    )
  }
}

export class MongoRundownRepository implements RundownRepository {
  public saveRundown(...): void {...}
}

export class CachedRundownRepository implements RundownRepository {
  constructor(rundownRepository: RundownRepository) {...}

  public saveRundown(...): void {
    // add to cache so that when others tries to read the Rundown, they retrieve the latest updated value
    this.rundownRepository.save(...)
  }
  ...
}

export class LockedRundownRepository implements RundownRepository {
  constructor(rundownRepository: RundownRepository) {...}

  public saveRundown(...): void {
     // lock logic before calling this.rundownRepository.save()
  }
}
```
Disclaimer: The above is just what was at the top of my head.

In the above example, we don't block any "execution logic" of the Rundown, i.e. we don't block Takes, setNexts etc. Since our "execution logic" ends with saving to the database (it emits events before saving) then we don't need to wait for database actions to finish before we are ready start a new "execution logic" loop.

Also, since we simply retrieve data from the cache when fetching data from the repository we don't need to wait for the locked database to unlock, since we don't actually call the database for data, we get it from the cache. (But we need to be aware that the cache has been removed from Segments and Parts in the previoues PR...)

A downside to the above approach is there will potential be spans of time where we are operating on data only in the cache that is still waiting to be saved to the database.

I'm not saying the above example is how we must implement locks on the database. It is just what I had imagined.
I think we need to have a talk about what we expect the lock and I don't think this PR is the place to do it :)

LindvedKrvang:
A thing I don't like about this approach: All callers to the database that want to use the lock _has_ to make an _active choice_ of using it. It's extremely easy to bypass the lock (or simply forget to use it).

I had hoped that the lock logic was just something that was automatically in place whenever I called `RundownRepository.save()` etc

RasmusAlbrektsen (approved):

src/business-logic/facades/service-facade.ts
```diff
@@ -33,23 +33,28 @@ import { IngestRundownSynchronizer } from '../services/ingest-rundown-synchroniz
...
-      RepositoryFacade.createRundownRepository(),
-      RepositoryFacade.createTimelineRepository(),
-      ServiceFacade.createTimelineBuilder(),
-      ServiceFacade.createIngestService(),
-      ServiceFacade.createPlayoutService(),
-      TimeoutCallbackScheduler.getInstance(LoggerFacade.createLogger()),
-      BlueprintsFacade.createBlueprint(),
-      LoggerFacade.createLogger(),
-    )
+  private static rundownService?: RundownService
```
LindvedKrvang: Singletons should be made within the class that wants to be a singleton. Not in the Facade

src/business-logic/services/ingest-data-change-service.ts
```diff
@@ -20,6 +20,7 @@ import { TimelineRepository } from '../../data-access/repositories/interfaces/ti
 import { ActionGenerationService } from './action-generation-service'
 import { Part } from '../../model/entities/part'
 import { Timeline } from '../../model/entities/timeline'
+import { AsyncLock } from '../../data-access/async-lock'
```
LindvedKrvang: You are importing a concrete class from the data-access layer in the business-logic layer.

Layers are only allowed to know about abstractions from other layers.

src/business-logic/services/synchronized-rundown-service.ts
```diff
@@ -0,0 +1,63 @@
+import { Part } from '../../model/entities/part'
+import { Piece } from '../../model/entities/piece'
+import { Owner } from '../../model/enums/owner'
+import { InTransition } from '../../model/value-objects/in-transition'
+import { RundownService } from './interfaces/rundown-service'
+import { AsyncLock } from '../../data-access/async-lock'
```
LindvedKrvang: Violation of what layers are allowed to know from other layers.

src/data-access/async-lock.ts
```diff
@@ -0,0 +1,54 @@
+import { Logger } from '../logger/logger'
+
+interface EnqueuedOperation {
+  operation: () => Promise<void>
+  enqueuedAtTimestampInMs: number
+}
+
+export class AsyncLock {
```
LindvedKrvang: There is nothing in this class that is data related. It's all pure logic. Why is this placed in the data-access layer instead of the logic layer?

src/data-access/facades/repository-facade.ts
```diff
@@ -72,9 +72,13 @@ import {
   MongoShowStyleVariantConfigurationListener
 } from '../repositories/mongo/mongo-show-style-variant-configuration-listener'
 import { RundownAggregateRepository } from '../repositories/interfaces/rundown-aggregate-repository'
+import { AsyncLock } from '../async-lock'
 
 export class RepositoryFacade {
 
+  private static rundownAggregateRepository?: RundownAggregateRepository
+  public static rundownLock: AsyncLock = new AsyncLock(LoggerFacade.createLogger())
```
LindvedKrvang: "Singletons" should be made within the class that needs to be a singleton. Not in the facade.

src/presentation/controllers/logger-controller.ts
```diff
@@ -0,0 +1,40 @@
+import { BaseController, PutRequest, RestController } from './base-controller'
+import { Logger, LogLevel } from '../../logger/logger'
+import { Request, Response } from 'express'
+import { ExhaustiveCaseChecker } from '../../business-logic/exhaustive-case-checker'
+import { Exception } from '../../model/exceptions/exception'
+import { HttpResponseFormatter } from '../interfaces/http-response-formatter'
+import { HttpErrorHandler } from '../interfaces/http-error-handler'
+
+@RestController('/loggers')
+export class LoggerController extends BaseController {
```
LindvedKrvang, line 10: This looks like changes from another PR. You should target this PR towards that PR, so we don't re-review changes already reviewed.

LindvedKrvang (changes requested):

src/presentation/controllers/logger-controller.ts
KvelaGorrrrnio: I mistakenly merged the other PR which was based on this.

src/business-logic/facades/service-facade.ts
KvelaGorrrrnio: Reverted.

src/business-logic/services/ingest-data-change-service.ts
KvelaGorrrrnio: AsyncLock is moved to business layer.

src/business-logic/services/synchronized-rundown-service.ts
KvelaGorrrrnio: AsyncLock is moved to business layer.

src/data-access/async-lock.ts
KvelaGorrrrnio: AsyncLock is moved to business layer.

src/data-access/facades/repository-facade.ts
KvelaGorrrrnio: This is moved to the ServiceFacade, but is kept as this, since it is a private ressource.

LindvedKrvang (approved):

PR #254, opened 2024-10-15
Log level API endpoint

KvelaGorrrrnio:
Note that this does not apply the the Blueprints logger.

By having this endpoint we can easily turn on/off logs in staging and production environments, so we have more information when we need to do log-based debugging.

Postman collection is updated with this.


src/presentation/controllers/logger-controller.ts
```diff
@@ -0,0 +1,40 @@
+import { BaseController, PutRequest, RestController } from './base-controller'
+import { Logger, LogLevel } from '../../logger/logger'
+import { Request, Response } from 'express'
+import { ExhaustiveCaseChecker } from '../../business-logic/exhaustive-case-checker'
+import { Exception } from '../../model/exceptions/exception'
+import { HttpResponseFormatter } from '../interfaces/http-response-formatter'
+import { HttpErrorHandler } from '../interfaces/http-error-handler'
+
+@RestController('/logger')
```
LindvedKrvang: To follow REST conventions we should write our endpoints using plural, so this would be `loggers`. That also means that the PUT endpoint should be `levels`. (The last does seem a bit weird though...)

LindvedKrvang (approved):

src/presentation/controllers/logger-controller.ts
KvelaGorrrrnio: Good point :) I have changed the base path to `/loggers`, but I have kept `level` as it represents an attribute on the ressource type.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #255, opened 2024-10-16
SOF-2302: Unplanned Parts no longer stay after moving next cursor

RasmusAlbrektsen:
We observed an issue where the unplayed unplanned parts did not get removed whenever we moved the next cursor away from them. This PR removes them by only looking at the currently active segment and checking whether we have any parts that fulfil this condition.


src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -227,9 +227,22 @@ export class RundownTimelineService implements RundownService {
 
     this.rundownEventEmitter.emitSetNextEvent(rundown)
 
+    this.deleteUnplayedUnplannedPartsFromActiveSegment(rundown)
     await this.saveRundown(rundown)
   }
 
+  private deleteUnplayedUnplannedPartsFromActiveSegment(rundown: Rundown): void {
```
LindvedKrvang, line 236: This looks like good logic to have tests for ;)

LindvedKrvang (approved):

src/business-logic/services/rundown-timeline-service.ts
RasmusAlbrektsen: @LindvedKrvang I added a test for the scenario under `setNext`, i'll be the first to admit that i could use more experience with test writing, so opinions/suggestions are most welcome :-)

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -227,9 +227,22 @@ export class RundownTimelineService implements RundownService {
...
+    if (!rundown.isActivePartSet()) {
+      return
+    }
+    rundown.getActiveSegment().getParts().forEach(part => {
+      if (!part.isPlanned && !part.isNext() && part.getExecutedAt() === 0) {
+        rundown.removePartFromSegment(part.id)
+        this.rundownEventEmitter.emitPartDeleted(rundown, part.getSegmentId(), part.id)
+      }
+    })
```
KvelaGorrrrnio, line 245: A bit of nesting can be reduced by using a guard or a filter instead. You don't have to change it - it is just an alternate way :) 
```typescript
    rundown.getActiveSegment().getParts()
      .filter(part => !part.isPlanned && !part.isNext() && part.getExecutedAt() === 0)
      .forEach(part => {
        rundown.removePartFromSegment(part.id)
        this.rundownEventEmitter.emitPartDeleted(rundown, part.getSegmentId(), part.id)
    })
```

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -265,6 +265,50 @@ describe(RundownTimelineService.name, () => {
...
+  describe(`${RundownTimelineService.prototype.setNext.name}`, () => {
+    const activePiece: Piece = EntityTestFactory.createPiece({ id: 'activePiece' })
+    const activePart: Part = EntityTestFactory.createPart({ id: 'activePart', pieces: [activePiece] })
+    const nextPart: Part = EntityTestFactory.createPart({ id: 'nextPart', pieces: [activePiece] })
+    const nextSegment: Segment = EntityTestFactory.createSegment({ parts: [nextPart] })
+
+    const rundownRepository: RundownRepository = mock<RundownRepository>()
+    const rundownEventEmitter: RundownEventEmitter = mock<RundownEventEmitter>()
+
+    describe('has active segment with unplanned part as next', () => {
```
KvelaGorrrrnio, line 269: Should the describe state `unplayed` as well?

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -265,6 +265,50 @@ describe(RundownTimelineService.name, () => {
...
   })
 
+  describe(`${RundownTimelineService.prototype.setNext.name}`, () => {
+    const activePiece: Piece = EntityTestFactory.createPiece({ id: 'activePiece' })
+    const activePart: Part = EntityTestFactory.createPart({ id: 'activePart', pieces: [activePiece] })
+    const nextPart: Part = EntityTestFactory.createPart({ id: 'nextPart', pieces: [activePiece] })
+    const nextSegment: Segment = EntityTestFactory.createSegment({ parts: [nextPart] })
+
+    const rundownRepository: RundownRepository = mock<RundownRepository>()
+    const rundownEventEmitter: RundownEventEmitter = mock<RundownEventEmitter>()
```
KvelaGorrrrnio: Aren't these only used in one test case? It can be considered to move them into the test case.

src/business-logic/services/test/rundown-timeline-service.spec.ts
```diff
@@ -265,6 +265,50 @@ describe(RundownTimelineService.name, () => {
...
+        })
+
+        const testee: RundownTimelineService = createTestee({
+          rundownEventEmitter,
+          rundownRepository,
+        })
+        when(rundownRepository.getRundown(rundown.id)).thenResolve(rundown)
+
+        await testee.setNext(rundown.id, nextSegment.id, nextPart.id)
+        expect(activeSegment.getParts()).not.toContain(unplayedUnplannedPart)
```
KvelaGorrrrnio: Note that this test is based on the assumption that object references do not change. If the object reference is broken for the segment or the part in either the RundownTimelineService or in one of its dependencies, then this test will break even though the functionality might still work. To create a more resilient test, one could find the active segment on the rundown and make an expectation on the shape of the part instead :)

KvelaGorrrrnio (approved):
🚀

src/business-logic/services/test/rundown-timeline-service.spec.ts
RasmusAlbrektsen: Absolutely right, they have been moved and i removed the superfluous setup as well :-)

src/business-logic/services/test/rundown-timeline-service.spec.ts
RasmusAlbrektsen: I chose not to, simply because i thought it made more sense that we set the initial scenario as just having the unplanned part. The way you suggest just sounded like very specific nesting of the describe. But i don't mind changing it if you want me to?

src/business-logic/services/test/rundown-timeline-service.spec.ts
RasmusAlbrektsen: I have changed it to use the active segment on the rundown instead, good point :-)

LindvedKrvang (approved):

PR #259, opened 2024-10-18
SOF-2304: Adding a part right after on air part might break next cursor, if the next cursor was set by the system.

KvelaGorrrrnio:
When adding a part right after the on air part, while the next cursor points to an off air segment and the next cursor was set by the system, the segment previously marked as next would not be unmarked, and the on air segment would not be marked as next.


RasmusAlbrektsen (approved):

src/model/entities/test/rundown.spec.ts
```diff
@@ -1317,42 +1324,48 @@ describe(Rundown.name, () => {
...
+          const firstSegment: Segment = EntityTestFactory.createSegment({
+            id: firstSegmentId,
             parts: [firstPart],
           })
 
-          const mockMiddlePiece: Piece = EntityMockFactory.createPieceMock({
-            id: 'middlePiece',
+          const middlePieceId: string = 'middle-piece-id'
+          const middlePartId: string = 'middle-part-id'
+          const middlePiece: Piece = EntityTestFactory.createPiece({
```
LindvedKrvang, line 1347: There is no need to create all these extra local variables. You already have that information in the entities.

In this specific scenario your tests looks like this:
```
const middlePartId: string = 'middle-part-id'
const middlePieceId: string = 'middle-piece-id'
const middlePiece: Piece = EntityTestFactory.createPiece({
  id: midlePieceId,
  partId: middlePartId,
   // other attributes
})

const middlePart: Part = TestEntityFactory.createPart({
  id: middlePartId
  pieces: [middlePiece]
})
...
const result: Piece | undefined = testee.getPart(middlePartId)?.getPieces().find(piece => piece.id === middlePieceId)
```
This could be written without the local variables:
```
const middlePiece: Piece = EntityTestFactory.createPiece({
  id: 'middle-piece-id',
   // other attributes
})
const middlePart: Part = EntityTestFactory.createPart({
  id: 'middle-part-id',
  pieces: [middlePice]
})
...
const result: Piece | undefined = testee.getPart(middlePart.id)?.getPieces().find(piece => piece.id === middlePiece.id)
```

The only reason for extracting one of the ids to a local variable is if `middlePiece` needs to have the id of `middlePart` which I'm quite sure is not the case in majority of the tests in this file.

This might not seem like a lot, but in a file like this it adds to be a quite substantial number of unneeded lines of code.

LindvedKrvang: This comment applies to the entire file

src/model/entities/test/rundown.spec.ts
```diff
@@ -3516,6 +3548,158 @@ describe(Rundown.name, () => {
...
       // It should also update the NextCursor, but that is being tested by "addSegment()".
     })
+
+    describe('when a part is added right after the on air part in same segment', () => {
+      describe ('when the segment marked as next is not the on air segment', () => {
+        describe('when the next cursor is set by the system', () => {
+          it('marks the on air segment as next', () => {
+            const onAirSegmentId: string =  'on-air-segment-id'
+            const onAirPart: Part = EntityTestFactory.createPart({ id: 'on-air-part-id', segmentId: onAirSegmentId, isOnAir: true })
+            const onAirSegment: Segment = EntityTestFactory.createSegment({ id: onAirSegmentId, parts: [onAirPart] })
```
LindvedKrvang, line 3558: The `onAirPart` does not need to the have the SegmentId so this could be written as 
```
const onAirPart: Part = EntityTestFActory.createpart({
  id: 'on-air-part-id,
  isOnAir: true
})
const onAirSegment: Segment = EntityTestFactory.createSegment({
  id: 'on-air-segment-id',
  parts: [onAirPart]
})
```

src/model/entities/rundown.ts
```diff
@@ -168,10 +168,12 @@ export class Rundown extends BasicRundown {
       return
     }
 
+    this.unmarkNextSegment()
```
LindvedKrvang, line 171: If I outcomment this line all test still pass, but I assume this is the essential part of the bugfix?

LindvedKrvang:
I'm a fan of you changing the usages of `EntityMockFactory` to `EntityTestFactory`.

I'm not a fan of the introduction of a lot of unneeded local variables containing ids.
Essential, all the places you added a Part- or SegmentId as a local variable it is not needed.

src/model/entities/test/rundown.spec.ts
KvelaGorrrrnio: The reason I started setting this up, is because the test failed without the relations. E.g. we search for segments based on the `Part.getSegmentId()` and so forth. I have not set it up for all tests, but for those where it mattered (it might be that I have done it in a few where it wasn't necessary).
I don't think that having extra lines should be a winning argument over working on incorrect data as we currently do. I would strongly prefer if we created valid data for our tests, since it will make the test more robust against changes.
Before settling on this, I tried updating the ids when composing the entities with the EntityTestFactory, but the entities are not set up for that kind of behavior. I suggest that this is kept and that we create a "tech-debt" task for updating the flow for creating actually valid rundowns e.g. by updating how the EntityTestFactory is used/behaves.

src/model/entities/test/rundown.spec.ts
KvelaGorrrrnio: See https://github.com/tv2/sofie-server/pull/259#discussion_r1808209509.

src/model/entities/rundown.ts
KvelaGorrrrnio: Nicely spotted! `isNext: true` was missing on the segment that was intended as marked as next in the test case that was written for checking this.

LindvedKrvang (approved):

src/model/entities/rundown.ts
LindvedKrvang: And that's why we make sure to see the test fail before we are done writing it :D

MadsThy (approved):

PR #260, opened 2024-10-18
SOF-2251 Expose endpoint to create StatusMessages

LindvedKrvang:
Expose endpoint to create/update a StatusMessage so that "clients" can send messages through the Alba system.


KvelaGorrrrnio (commented):
Why is this needed?

LindvedKrvang:
> Why is this needed?

For https://tv2cms.atlassian.net/browse/SOF-2246

So that we can send messages from the `tv2-server` to all clients in the system without having to reimplement it.

It's currently going to be used to indicate the connection status of the Panels. Are we connecting, did we lose connection, did we reconnect etc

src/presentation/controllers/system-information-controller.ts
```diff
@@ -42,4 +44,23 @@ export class SystemInformationController extends BaseController {
       this.httpErrorHandler.handleError(response, error as Exception)
     }
   }
+
+  @PutRequest('/statusMessages')
+  public async updateStatusMessage(request: Request, response: Response): Promise<void> {
+    try {
+      const statusMessageDto: StatusMessageDto = request.body as StatusMessageDto
```
MadsThy, line 51: I don't know if it is the case anywhere else, but I think a validation check here would be wonderful in terms of knowing if the incoming data is correct :)

MadsThy (approved):

LindvedKrvang:
Closing this PR since the functionality is no longer needed.

PR #261, opened 2024-10-22
SOF-2310: Listen for piece ingest changes and other ingest fixes

KvelaGorrrrnio:
Sometimes piece ingest changes would be lost during synchronization. The cause was two-fold, where the first was due to piece changes sometimes being the last that Sofie Core writes. And since we didn't listen for piece changes, the synchronization could happen before the piece changes were written to the database from Sofie Core. To combat this, we are now listening for piece ingest changes from Sofie Core.
The 3-split listening for _create_, _update_ and _delete_ is kept to give greater control to the `IngestDataChangeService` - e.g. in handling failed retrievals of deleted entities.

The other cause was that the set that holds the affected rundown ids, was deleted after the synchronization would happen, so all observed changes during the synchronization would be ignored until next synchronization.

The PR also:
- introduces `rundownId` on pieces (was already present on ingest pieces) for easier handling of e.g. deletion and ingest change management (determining which rundown had a piece deleted).
- Increases ingest synchronize debounce delay from 200ms to 600ms. The changes made in core for an update would often create a rundown replace and a entity CUD event (e.g. segment replace), where there is a delay of 300-400ms between the two events, which causes two synchronizations to trigger.
- Reverts `RundownAggregateRepository` to an interface, such that WebStorm can actually find out which classes and methods implements the type/interface.
- The piece deletion logic on `saveRundown`/`deleteRundown` is updated to use the rundownId attribute on pieces.
- The `MongoIngestedXChangedListener` classes uses the `fullDocument` of the change instead of fetching it from the database.
- `Segment.expectedDurationInMs` is stored to and restored from the database.
- `IngestedRundown.timing` is parsed correctly without null values.


src/data-access/repositories/mongo/mongo-ingested-entity-converter.ts
```diff
@@ -33,20 +33,20 @@ type MongoIngestedRundownTiming = MongoForwardRundownTiming | MongoBackwardRundo
 interface MongoForwardRundownTiming { // PlaylistTimingForwardTime from blueprints-integration
   type: MongoRundownTimingType.FORWARD
   expectedStart: number
-  expectedDuration?: number
-  expectedEnd?: number
+  expectedDuration?: number | null
+  expectedEnd?: number | null
```
LindvedKrvang, line 37: Optional _and_ null? 
So right now the type is `number | undefined | null`? 
If that's true it would be nice to remove either `undefined` or `null`. 
Comment goes for the other places in here aswell.

LindvedKrvang (approved):
Nice work :)

src/data-access/repositories/mongo/mongo-ingested-entity-converter.ts
KvelaGorrrrnio: The type of the attribute is `number | null`, but it is also a possibility that it is not present, hence the optionality. Mongo uses `null` for empty values where we use `undefined` in our TypeScript code. So when we try to access the attribute, we can get either a number, `undefined` or `null`.
So by having both we model the data correct and TypeScript can help us ensure that we map it correctly to only being undefined/optional or a number.

src/business-logic/services/ingest-data-change-service.ts
```diff
@@ -64,28 +70,45 @@ export class IngestDataChangeService implements DataChangeService {
...
+    this.segmentChangedListener.onCreated(segment => this.registerChangeForRundown(segment.rundownId))
+    this.segmentChangedListener.onUpdated(segment => this.registerChangeForRundown(segment.rundownId))
     this.segmentChangedListener.onDeleted(segmentId => {
       this.segmentRepository.getSegment(segmentId)
-        .then(segment => this.registerReceivedDataChangeEvent(segment.rundownId))
-        .catch(error => this.logger.data(error).error(`Failed getting segment with id '${segmentId}' for delete segment event.`))
+        .then(segment => this.registerChangeForRundown(segment.rundownId))
+        .catch(error => {
+          this.logger.data(error).error(`Failed getting segment with id '${segmentId}' for delete segment event.`)
+          this.markAllRundownsAsAffected().catch(error => this.logger.data(error).error(`Failed marking all rundowns as affected when deletion of the segment with id '${segmentId}' failed.`))
```
RasmusAlbrektsen, line 84: I'm just curios, but what is the reasoning for marking all rundowns as affected when a singular entity fails? Is it just to ensure that we register this for all rundowns in the case they make use of it?

RasmusAlbrektsen (approved):

src/business-logic/services/ingest-data-change-service.ts
KvelaGorrrrnio: The idea is that since we can't find the entity to delete, it indicates that we are either in-between states or that we are in a bad state. Since we are unable to get the entity and find the rundown, the only real action we can take to get into a good state is to synchronize everything.

PR #264, opened 2024-10-23
SOF-2313: Bases ingest change detection for parts on ingested parts

KvelaGorrrrnio:
Prior to this change, the detection of ingest changes for a part was based on the values from an `IngestedPart` and the `Part`. This caused the issue that if a part was modified, e.g. updated inTransition from a _mix_-action, then the part would be updated (and unsynced if on air) on re-ingest or server restart. Now the comparison for whether a part is changed by an ingest update is based on the observed `IngestedPart` and the ingested part of the part (`Part.ingestedPart`).

The PR changes the lower limit of transition piece durations from 1000 milliseconds to 0 milliseconds, so shorter transitions (e.g. `Mix 8F`) are displayed correctly in the web client. It also makes it easier to debug, when there are issues with effects that a transition should have.


src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -47,8 +47,6 @@ import { Tv2Logger } from '../tv2-logger'
 import { ActionFactory } from './action-factory'
 import { FrameTimeConverter } from '../helpers/frame-time-converter'
 
-const MINIMUM_DURATION_IN_MS: number = 1000
```
LindvedKrvang, line 50: Wouldn't this make a Mix8 transition almost impossible to see in the RundownView?

LindvedKrvang (approved):

MadsThy (approved):

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
KvelaGorrrrnio: It looks like this:
image

RasmusAlbrektsen (approved):

PR #266, opened 2024-10-25
SOF-2312: Remove unsynced content on rundown deactivaition and reset.

KvelaGorrrrnio:
The PR changes the Rundown entity, so it removes unsynced content on `reset`. The logic for resetting the rundown is put into `Rundown.reset`, and `Rundown.deactivate` now calls `Rundown.reset`. So hereby `deactivate` also removes unsynced content.

Note: For the web client to behave correctly, https://github.com/tv2/sofie-web-client/pull/279 should be used as well.


src/model/entities/rundown.ts
```diff
@@ -726,9 +717,25 @@ export class Rundown extends BasicRundown {
...
-    isRundownInRehearsalBeforeResetting ? this.enterRehearsal() : this.activate()
+    this.clearActiveCursor()
+    this.clearNextCursor()
+    this.infinitePieces = new Map()
+    this.previousPart = undefined
+    this.persistentState = undefined
+
+    this.resetSegments()
+    this.resetHistory()
+    this.infinitePieces = new Map()
```
MadsThy: I'm guessing there is a timing issue or something that means it is needed to overwrite this.infinitePieces two times in the same function? 😅

MadsThy (approved):

RasmusAlbrektsen (approved):

src/model/entities/rundown.ts
LindvedKrvang: There shouldn't be.
This looks like I didn't see that there already was a reset of the `infinitePieces` when I added the second line.

LindvedKrvang: Atleast, there is nothing happening between the two "resets" that modifies the infinite Pieces

src/model/entities/rundown.ts
```diff
@@ -136,6 +136,7 @@ export class Rundown extends BasicRundown {
   }
 
   private resetSegments(): void {
+    this.segments = this.segments.filter(segment => !segment.isUnsynced())
```
LindvedKrvang, line 139: You have now given `resetSegments` an extra responsibility since it now also removes unsynced Segments.
I would like that we didn't mix those two responsibilities.
Or, at the very least, update the name of the method to match what it does.

KvelaGorrrrnio: I will say that it is a matter of definition. Resetting a segment is returning it to its ingest state - since an unsynced segment is detached from its ingest state, then it is reset to nothing, hence the removal.

LindvedKrvang: I agree with that argument _if_ it happened inside of Segment.
I would expect `resetSegments` to iterate of all the Segments and reset them. Not _remove_ them.

KvelaGorrrrnio: I still think that it is one responsibility that `Rundown.resetSegments` facilitates. But I have split it up as requested.

src/model/entities/rundown.ts
KvelaGorrrrnio: Nice catch! I have removed the second clear statement for infinite pieces.

PR #268, opened 2024-10-28
SOF-2306 Unplanned Breaker Transition Effect Actions are blocking Take for too long

LindvedKrvang:
No longer convert casparCgPreRollDuration to miliseconds because they already are miliseconds.

When calculating the `blockTakeDuration` for a BreakerTransitionEffectAction, we would convert the `casparCgPreRollDuration` from frames to milliseconds. Only problem is, that `casparCgPreRollDuration` already is in milliseconds, so we would get a way higher number of milliseconds for the block duration.


src/model/entities/test/entity-test-factory.ts
```diff
@@ -199,4 +200,18 @@ export class EntityTestFactory {
       ...actionManifest,
     }
   }
+
+  public static createPieceAction(action?: Partial<PieceAction>): PieceAction {
+    return {
+      id: 'randomActionId',
+      rundownId: 'randomRundownId',
+      name: 'randomActionName',
```
MadsThy, line 208: Random names just called random 🤣 Really? 🤣

src/model/entities/test/entity-test-factory.ts
```diff
@@ -199,4 +200,18 @@ export class EntityTestFactory {
...
+      rank: Math.floor(Math.random() * 100),
+      type: PieceActionType.INSERT_PIECE_AS_ON_AIR,
+      data: {
+        pieceInterface: this.createPieceInterface()
```
MadsThy, line 212: If the function is only running because it gets called inside itself, what will then trigger the function in the first place? 😄  And how do you prevent it from running amok in recursive hell?

MadsThy (approved):

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -448,7 +448,7 @@ export class Tv2TransitionEffectActionFactory extends ActionFactory {
 
   private createPartInTransitionForBreakerTransitionEffect(breakerActionMetadata: Tv2BreakerTransitionEffectActionMetadata): InTransition {
     return {
-      blockTakeDuration: this.frameTimeConverter.convertFramesToMilliseconds(breakerActionMetadata.breaker.durationInFrames + breakerActionMetadata.casparCgPreRollDuration),
+      blockTakeDuration: this.frameTimeConverter.convertFramesToMilliseconds(breakerActionMetadata.breaker.durationInFrames) + breakerActionMetadata.casparCgPreRollDuration,
```
KvelaGorrrrnio, line 451: To help in other places and future use cases, it would be nice that the unit was added to the attribute. Also, the `casparCg` prefix seems unnecessary, since it could be used for something else than CasparCG. So I suggest that `Tv2BreakerTransitionEffectActionMetadata.casparCgPreRollDuration` is renamed to `.preRollDurationInMs` or `.prerollDurationInMs`.

KvelaGorrrrnio (approved):

src/blueprints/tv2/action-factories/tv2-transition-effect-action-factory.ts
RasmusAlbrektsen: It seems we are familiar with that way of spelling it at TV2:
https://tv2reklame.dk/tv2dk-apps/specifikationer/prerollbumper/

RasmusAlbrektsen (approved):

src/model/entities/test/entity-test-factory.ts
KvelaGorrrrnio: There is no recursion to see here :D The method is `createPieceAction` and it is `createPieceInterface` that is called inside it.

PR #271, opened 2024-11-01
SOF-2317 - Ingest Updates Lost When Deleting Next Segment

RasmusAlbrektsen:
This PR fixes an issue where we would fail trying to get the next cursor whilst trying to create the lookahead timeline objects. We would also try to emit `setNextEvent` even though we had nothing in the rundown. 

Also included: 
The tests can no longer abstract away the segments so these have been included. Also a test that covers the issue addressed in this PR has been added.


KvelaGorrrrnio (approved):

LindvedKrvang (approved):

src/business-logic/services/test/superfly-timeline-builder.spec.ts
```diff
@@ -1,20 +1,22 @@
...
-import { TransitionType } from '../../../model/enums/transition-type'
-import { PieceLifespan } from '../../../model/enums/piece-lifespan'
-import { ObjectCloner } from '../interfaces/object-cloner'
-import { anything, instance, mock, when } from '@typestrong/ts-mockito'
-import { Studio } from '../../../model/entities/studio'
-import { StudioLayer } from '../../../model/value-objects/studio-layer'
-import { LookaheadMode } from '../../../model/enums/lookahead-mode'
-import { LastPartInRundownException } from '../../../model/exceptions/last-part-in-rundown-exception'
-import { EntityTestFactory } from '../../../model/entities/test/entity-test-factory'
+import {TimelineBuilder} from '../interfaces/timeline-builder'
```
MadsThy: Something is wrong with your formatter i think 😄 It has removed all the spaces in the imports here

MadsThy (approved):

src/business-logic/services/test/superfly-timeline-builder.spec.ts
RasmusAlbrektsen: ............... I'll fix it.....

PR #274, opened 2024-11-05
Fix  Rundown activation breaking when no valid segments in rundown

RasmusAlbrektsen:
Since we threw an error during our initialization of a rundown from inactive to active state, we never update the frontend with it's activation. Since the behaviour for the user is that we do not see the rundown active, it has not been made impossible. 

Tests will be added, i just quickly created this PR so we could build it for production-test purposes.


src/model/entities/rundown.ts
```diff
@@ -88,21 +88,27 @@ export class Rundown extends BasicRundown {
 
   public activate(): void {
     if (this.isActive()) {
-      throw new AlreadyActivatedException('Can\'t activate Rundown since it is already activated')
+      throw new AlreadyActivatedException('Can\'t activate Rundown since it is already activated.')
```
MadsThy, line 91:
```suggestion
      throw new AlreadyActivatedException(`Can't activate Rundown since it is already activated.`)
```

src/model/entities/rundown.ts
```diff
@@ -88,21 +88,27 @@ export class Rundown extends BasicRundown {
...
     }
     if (this.mode === RundownMode.REHEARSAL) {
       this.mode = RundownMode.ACTIVE
       return
     }
+    if (!this.doesValidSegmentExistInRundown()) {
+      throw new UnsupportedOperationException('Can\'t activate a Rundown that has no valid Segment.')
```
MadsThy:
```suggestion
      throw new UnsupportedOperationException(`Can't activate a Rundown that has no valid Segment.`)
```

src/model/entities/rundown.ts
```diff
@@ -88,21 +88,27 @@ export class Rundown extends BasicRundown {
...
+    }
     this.initializeRundown(RundownMode.ACTIVE)
   }
 
   public enterRehearsal() :void {
     if (this.isActive()) {
-      throw new AlreadyActivatedException('Can\'t set Rundown to rehearsal since it is already activated')
+      throw new AlreadyActivatedException('Can\'t set Rundown to rehearsal since it is already activated.')
```
MadsThy, line 102:
```suggestion
      throw new AlreadyActivatedException(`Can't set Rundown to rehearsal since it is already activated.`)
```

src/model/entities/rundown.ts
```diff
@@ -88,21 +88,27 @@ export class Rundown extends BasicRundown {
...
     }
     if (this.getMode() === RundownMode.REHEARSAL) {
-      throw new AlreadyRehearsalException('Can\'t set Rundown to rehearsal since it is already in rehearsal')
+      throw new AlreadyRehearsalException('Can\'t set Rundown to rehearsal since it is already in rehearsal.')
```
MadsThy, line 105:
```suggestion
      throw new AlreadyRehearsalException(`Can't set Rundown to rehearsal since it is already in rehearsal.`)
```

src/model/entities/rundown.ts
```diff
@@ -88,21 +88,27 @@ export class Rundown extends BasicRundown {
...
+    }
+    if (!this.doesValidSegmentExistInRundown()) {
+      throw new UnsupportedOperationException('Can\'t set a Rundown to rehearsal that has no valid Segment.')
```
MadsThy:
```suggestion
      throw new UnsupportedOperationException(`Can't set a Rundown to rehearsal that has no valid Segment.`)
```

MadsThy (approved):

src/model/entities/rundown.ts
KvelaGorrrrnio: I don't think this is one of the use cases where template strings are preferred, since we don't have multiline or dynamic content in the messages :)

src/model/entities/rundown.ts
```diff
@@ -112,6 +118,10 @@ export class Rundown extends BasicRundown {
     this.setFirstSegmentAndPartNextCursor()
   }
 
+  private doesValidSegmentExistInRundown(): boolean {
```
LindvedKrvang: Ideally, the method should be placed _after_ it's first use (which is way further down in this case). Right now it's placed before the method it wants to "guard against", but there is no direct link between them.

src/model/entities/rundown.ts
```diff
@@ -730,7 +740,7 @@ export class Rundown extends BasicRundown {
     this.resetSegments()
     this.resetHistory()
 
-    if (this.mode !== RundownMode.INACTIVE) {
+    if (this.mode !== RundownMode.INACTIVE && this.doesValidSegmentExistInRundown()) {
       this.setFirstSegmentAndPartNextCursor()
```
LindvedKrvang: Instead of making the check outside of `setFirstSegmentAndPartNextCursor` then consider having the `doesValidSegmentExistInRundown` check as a guard clause inside of `setFirstSegmentAndPartNextCursor`.
```
private setFirstSegmentAndPartNextCursor(): void {
  if (!this.doesValidSegmentExistInRundown()) {
    return
  }
  ...
}
```

With the current implementation, you would need to _also_ make the check _everytime_ you want to call `setFirstSegmentAndPartNextCursor`

LindvedKrvang (approved):

src/model/entities/rundown.ts
RasmusAlbrektsen: Good point, it has been moved below `setFirstSegmentAndPartNextCursor`, which is the new point of use.

src/model/entities/rundown.ts
RasmusAlbrektsen: A very good suggestion. After observing it happen during test, it seems that we should allow the rundown to activate anyway.

src/model/entities/rundown.ts
RasmusAlbrektsen: I have refrained from doing this for now :-)

MadsThy (approved):

src/model/entities/test/rundown.spec.ts
```diff
@@ -3204,6 +3220,30 @@ describe(Rundown.name, () => {
     })
   })
 
+  it('resets when the last remaining on air segment is unsynced', () => {
```
KvelaGorrrrnio: `when the last remaining on air segment is unsynced` sounds like a scenario that could be a `describe` :)

KvelaGorrrrnio (approved):

src/model/entities/test/rundown.spec.ts
RasmusAlbrektsen: I have moved the scenarios into describes instead :-)

PR #275, opened 2024-11-05
SOF-2325: Ingest parts added to an empty segment.

KvelaGorrrrnio:
Ingested parts not in the current segment are now included when parts are updated on a segment.


MadsThy (approved):

src/business-logic/services/test/ingested-entity-to-entity-mapper.spec.ts
```diff
@@ -3,8 +3,56 @@ import { Part } from '../../../model/entities/part'
...
+  describe(IngestedEntityToEntityMapper.prototype.updateSegmentWithIngestedSegment, () => {
+    describe('when ingested segment has new parts', () => {
+      it('adds the new parts', () => {
+        const segmentId: string = 'segment-id'
+        const partId: string = 'part-id'
+        const part: Part = EntityTestFactory.createPart({ id: partId, segmentId })
+        const segment: Segment = EntityTestFactory.createSegment({ id: segmentId, parts: [part] })
+
+        const ingestedPart: IngestedPart = EntityTestFactory.createIngestedPart({ id: partId, segmentId })
+        const newPartId: string = 'new-part-id'
```
LindvedKrvang: Minor thing: If you want to save a few lines, then `partId` and `newPartId` does not need to be local variables. They can be given directly to the factory and then accessed through the returned Part object.
```
const part: Part = EntityTestFactory.createPart({ id: 'part-id', segmentId })
...
expect(result.getParts()).toEqual([ expect.objectContaining({ id: part.getId() }), ...])
```

src/business-logic/services/test/ingested-entity-to-entity-mapper.spec.ts
```diff
@@ -3,8 +3,56 @@ import { Part } from '../../../model/entities/part'
...
+        const ingestedSegment: IngestedSegment = EntityTestFactory.createIngestedSegment({ id: segmentId, ingestedParts: [ingestedPart, newIngestedPart] })
+
+        const testee: IngestedEntityToEntityMapper = new IngestedEntityToEntityMapper()
+
+        const result: Segment = testee.updateSegmentWithIngestedSegment(segment, ingestedSegment)
+
+        expect(result.getParts()).toEqual([
+          expect.objectContaining({ id: partId }),
+          expect.objectContaining({ id: newPartId }),
+        ])
```
LindvedKrvang: Minor thing: When testing that something has been added (or removed for that matter) I personnally like to have an assertions that verifies that the thing I want to add _isn't_ already there before executing the method under test.
This is mainly to prevent brainfarts like:
```
it('test something', () => {
   const cars: Car[] =[new Car('Skoda'), new Car('Audi')]
  
   const testee: CarFactory: new CarFactory()
   testee.addAudi(cars)

   expect(cars).toContain(/* Assertion to check an Audi is in the array */)
}

export class CarFactory {
 
  public addAudi(cars: Car[]): void {
     // Method accidentally left empty after a refactor.
  }
}
```
Above test still parse even though the method under test is empty because someone left an Audi car in the mock to begin with.

src/business-logic/services/test/ingested-entity-to-entity-mapper.spec.ts
```diff
@@ -3,8 +3,56 @@ import { Part } from '../../../model/entities/part'
...
+      })
+
+      it('preserves the rank ordering', () => {
```
LindvedKrvang, line 36: I feel that in order to test rank order you would need more than one test :p 
It's quite easy (on accident) to change the order of the Parts in the test-setup which could lead to a wrong implementation in the future where all test parses.

Imagine this timeline:
- Test is written today
- 3 months from now someone refactors the test-setup and accidentally swap `[ingestedPart, newIngestedpart]` to `[newIngestedPart, ingestedPart]`. Test still parse because the method under test implementation is unchanged.
- 6 months from that, someone refactors the method under test where they accidentally mess up the ranking (it simply doesn't sort anymore). The test still parse because the test-setup ranking was "correct" from the beginning and there is only one test to test the ranking, but the implementation no longer ranks the Parts.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/business-logic/services/test/ingested-entity-to-entity-mapper.spec.ts
KvelaGorrrrnio: The id constants are removed and uses the ids on the objects.

src/business-logic/services/test/ingested-entity-to-entity-mapper.spec.ts
KvelaGorrrrnio: I have added assertion of initial data structure to avoid brain farts 💨

src/business-logic/services/test/ingested-entity-to-entity-mapper.spec.ts
KvelaGorrrrnio: I have updated the test to be more robust, by creating a larger set of parts and new parts, with intertwined ranks.

PR #276, opened 2024-11-06
SOF-2324 Don't insert source in inputs not confiured in SplitScreen

LindvedKrvang:
Only allow to insert sources into inputs in SplitScreens that have a configuration for the input.

When inserting a Source into an Input in a SplitScreen, we now only allow to insert the Source if the Input is enabled in the SplitScreen.

For this to work, it does require that the SplitScreens have been configured correctly.

Added the callback `isActionAllowedToMutatePiece` so that we can prevent the Action from mutating the Piece if it is not allowed (for example input isn't configured for SplitScreen).


MadsThy (approved):

src/model/entities/action.ts
```diff
@@ -64,6 +64,7 @@ export interface MutateActionWithPieceMethods {
   type: MutateActionType.PIECE
   updateActionWithPiece: (action: Action, piece: Piece) => Action
   piecePredicate: (piece: Piece) => boolean
+  isActionAllowedToMutatePiece: (action: Action, piece: Piece) => boolean
```
KvelaGorrrrnio, line 67: Just a thought. Since it is only a subset of actions where this is necessary, then this could be optional, where it defaults to allowing the mutation.

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

PR #277, opened 2024-11-07
SOF-2329: Split screen actions should use input to box mappings

KvelaGorrrrnio:
Uses DVE inputs mapping field from settings to determine which split screen box to update when inserting sources into a split screen.
The PR also removes isActionAllowedToMutatePiece in favor of guards in the mutate methods to allow for proper client responses via meaningful exceptions:

image


MadsThy (approved):

RasmusAlbrektsen (approved):

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
```diff
@@ -369,14 +379,16 @@ export class Tv2SplitScreenActionFactory extends ActionFactory {
...
-      return false
-    }
+  private throwExceptionForUnavailableSplitScreenBoxInput(index: number, templateName: string): never {
+    throw new Tv2UnavailableOperationException(`Input ${index + 1} is not configured for the split screen '${templateName}'.`)
+  }
 
-    const insertSourceInputMetadata: Tv2SplitScreenInsertSourceInputMetadata = action.metadata as Tv2SplitScreenInsertSourceInputMetadata
-    return pieceMetadata.splitScreen.boxes[insertSourceInputMetadata.inputIndex].enabled
+  private getBoxIndex(inputIndex: number, splitScreenConfig: NonNullable<Tv2PieceMetadata['config']>): number {
+    return (splitScreenConfig.DVEInputs || '1:INP1;2:INP2;3:INP3;4:INP4')
```
LindvedKrvang: I think it would be nice with a constant for the default input mappings.

LindvedKrvang (approved):

src/blueprints/tv2/action-factories/tv2-split-screen-action-factory.ts
KvelaGorrrrnio: It is now in a constant :+1:

PR #280, opened 2024-11-13
Fix clear OverlayPilot graphics on clear graphics

LindvedKrvang:
For some reason, pilot overlay graphics wasn't being included when as a source layer for graphics.
When determining whether we should find infinite Pieces, we now no longer include infinite Pieces which has been stopped i.e. their `executedAt + duration < Date.now()`. 

Also in this PR:
- No longer stop Pieces that are already stopped, when stopping Pieces on a layer


src/model/entities/rundown.ts
```diff
@@ -424,11 +425,18 @@ export class Rundown extends BasicRundown {
     this.setInfinitePieces(layersWithPieces)
   }
 
+  private hasPieceEnded(piece: Piece): boolean {
+    return !!piece.getDuration() && piece.getExecutedAt() + piece.getDuration()! < Date.now()
+  }
```
KvelaGorrrrnio: This could be a method on `Piece`, then the duplication here and in segment.ts can be avoided.

KvelaGorrrrnio (approved):

MadsThy (approved):

src/model/entities/rundown.ts
LindvedKrvang: That's a good comment. Maybe I should use it in this PR (Spoiler alert: I did)

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

PR #282, opened 2024-11-21
Sof 2361/ovl graphics isnt cleared properly

LindvedKrvang:
Clearing OVL Pilot Graphics actual clears the graphics.

It's possible to set 8 different combinations of the lifespan of an OVL Pilot Graphics in iNews. 6 out of these 8 didn't have correct clear behaviour when executing our ClearGraphicsAction. This was due to when checking if a Piece had ended we checked if the ExecutedAt plus Duration was `less than` the current time and not `less than or equal to` the current time. Adding the `<=` seems to have fixed the clear behaviour.

Also in this PR:
- The combination: Unplanned with a preplanned duration, did not remove the graphic after the duration had passed. This is because the duration value was stored in the database as `adLibPiece.enable.duration` and not `adLibPiece.expectedDuration` which is how the duration of our other OVL graphics are stored... We now check both values when fetching the expected duration for an OVL graphic.
- OVL Pilot Graphics Action now gets a duration set from the expected duration.


RasmusAlbrektsen (approved):

src/data-access/repositories/mongo/mongo-ad-lib-piece-repository.ts
```diff
@@ -51,11 +54,23 @@ export class MongoAdLibPieceRepository extends BaseMongoRepository<MongoAdLibPie
...
+  private getExpectedDuration(adLibPiece: MongoAdLibPiece): number | undefined {
+    if (adLibPiece.expectedDuration) {
+      return adLibPiece.expectedDuration
+    }
+
+    if (adLibPiece.enable?.duration) {
+      return adLibPiece.enable.duration
+    }
+
+    return undefined
```
MadsThy, line 74: Could this be a more simplified version? 😄 

```suggestion
    return adLibPiece.expectedDuration ?? adLibPiece.enable?.duration ?? undefined
```

MadsThy (approved):

src/data-access/repositories/mongo/mongo-ad-lib-piece-repository.ts
KvelaGorrrrnio: If only it was undefined values we checked against, then even more simplified version would be:
```typescript
return adLibPiece.expectedDuration ?? adLibPiece.enable?.duration
```
But I think that the reason it is done as if-statements is that we also want to check for `0`.
So if this should be simplified, then one could write:
```typescript
return adLibPiece.expectedDuration || adLibPiece.enable?.duration || undefined
```

src/model/entities/piece.ts
```diff
@@ -164,6 +164,6 @@ export class Piece {
   }
 
   public hasEnded(): boolean {
-    return !!this.duration && this.executedAt + this.duration < Date.now()
+    return !!this.duration && this.executedAt + this.duration <= Date.now()
```
KvelaGorrrrnio: Suggestion: From a semantic perspective, if a duration is not set, then the duration at the moment of observation is infinite. And a piece can only have ended if it was on air at some point, which the executedAt is an indicator for. So an alternative way of expressing this could be as:
```typescript
public hasEnded(): boolean {
  if (!this.executedAt) {
    return false
  }
  const durationInMs: number = this.duration ? this.duration : Infinity
  return this.executedAt + durationInMs <= Date.now()
}
```

src/model/entities/test/piece.spec.ts
```diff
@@ -163,4 +163,41 @@ describe(Piece.name, () => {
...
+          expect(testee.hasEnded()).toBeTruthy()
+        })
+      })
+
+      describe('executedAt + duration is equal to now', () => {
+        it('has ended', () => {
+          const testee: Piece = EntityTestFactory.createPiece({ executedAt: 250, duration: 250 })
+          expect(testee.hasEnded()).toBeTruthy()
+        })
+      })
```
KvelaGorrrrnio, line 193: Since the issue was a off-by-one, it would be nice to have that tested (e.g. fo duration `249`, `250` and `251`.

src/model/entities/piece.ts
KvelaGorrrrnio: I am a bit curious about why `<=` fixes it. Am I right in saying that the issue here was that when we insert a clear piece, the graphics pieces are stopped, but the duration set by stopping it added with the executed at timestamp, would give the current time so it would be included in the timeline either way?

LindvedKrvang: I'm not entirely sure I follow your question, but the gist of this solution is:

The time that passes between the duration being set (when stopping the Piece) and checking if the Piece has ended, was/is so short that `Date.now()` returns the same value. So the Piece is stopped, which means it has ended, but our check for if it has ended would look like this: `this.executedAt + this.duration < Date.now() <=> 200 + 300 < 500` which returns false and the Piece would be included in the Timeline again.

The flow (dumped down a lot) looks like this:
```
insertClearGraphics() {
  const pieceToStop // Code to find the Piece to stop.
  pieceToStop.stop() // Date.now() is equal to 1000. executedAt is equal to 250. This makes duration equeal to 750.

  ... // more stuff happening

  updateInfinitePieces()
  buildTimeline()
}

updateInfinitePieces() {
  const oldInfinitePiecesToCheckIfStillShouldBeIncluded // This is the same Piece as the Piece we stopped
  if (oldInfinitePiecesToCheckIfStillShouldBeIncluded.hasEnded()) { // Date.now() is still 1000. That makes 250 + 750 < 1000 false when .hasEnded() should have returned true, which includes our graphics when it shouldn't.
     // Include it when we build the Timeline
  }
}
```

src/data-access/repositories/mongo/mongo-ad-lib-piece-repository.ts
LindvedKrvang: "simplified" is quite relative.
Yes, you make the code "simpler" by having fewer lines, but you make the individual lines more complex and harder to understand, because so much more is happening on a given line.

Personally, I don't think everything should be a one-liner. Especially when the one-liner can return more-than-two outcomes (which is the case here).
The code also becomes more difficult to debug using a debugger, because a single step in the debugger steps over so many things (greatly reducing one of the strongest features of the debugger).

But this all comes down to where the line between verbosity and "as few lines as possible" is drawn. For me, that's useally if the the single line returns more-than-two outcomes.

Sidenote: Verbosity is not a bad thing :)

src/model/entities/piece.ts
LindvedKrvang: It's been updated.

src/model/entities/test/piece.spec.ts
LindvedKrvang: Agreed. I have added them :)

src/data-access/repositories/mongo/mongo-ad-lib-piece-repository.ts
KvelaGorrrrnio: From the conversation, the issue I see is a misunderstanding of what is being checked. Whether that be the conditional not being clear enough (e.g. `adLibPiece.expectedDuration` is checked for both `null | undefined` and and if it is greater than `0`) or if it is an issue in how we model the duration values (e.g. `0` could be used instead of `null` or `undefined`) or both.

src/model/entities/test/piece.spec.ts
```diff
@@ -163,4 +163,55 @@ describe(Piece.name, () => {
...
+        const testee: Piece = EntityTestFactory.createPiece({ duration: undefined })
+        expect(testee.hasEnded()).toBeFalsy()
+      })
+    })
+
+    describe('the Piece has a duration', () => {
+      beforeEach(() => {
+        const now: number = 500
+        jest.useFakeTimers({ now })
+      })
```
KvelaGorrrrnio, line 179: Just a note: Since the fake timer is not being advanced, then running this _before each_ is not necessary. It could be executed directly in the `describe`.
And if a timestamp for now was given in as an argument to `hasEnded`, then the need for fake timers would disappear and the logic becomes more reusable :)

KvelaGorrrrnio (approved):

MadsThy (approved):

PR #285, opened 2024-11-21
SOF-2351 Inserting a Part OnAir no longer always sets the Part immediately next to the OnAir Part as next

LindvedKrvang:
When inserting a Part as OnAir we now keep the original next Part as next, even if this Part is somewhere else in the Rundown


RasmusAlbrektsen (approved):

MadsThy (approved):

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -276,13 +277,16 @@ export class RundownTimelineService implements RundownService {
...
     const unplannedNextPartToKeepAsNextPart: Part | undefined = !rundown.getNextPart().isPlanned ? rundown.getNextPart() : undefined
+    const nextCursor: RundownCursor | undefined = rundown.getNextCursor()
 
     rundown.insertPartAsNext(part)
     rundown.takeNext()
     rundown.getActivePart().setEndState(this.getEndStateForActivePart(rundown))
 
     if (unplannedNextPartToKeepAsNextPart) {
       rundown.insertPartAsNext(unplannedNextPartToKeepAsNextPart)
+    } else if (nextCursor) {
```
KvelaGorrrrnio, line 288: The `else` can be omitted, by moving the if statement to a new line :)

KvelaGorrrrnio (approved):

src/business-logic/services/rundown-timeline-service.ts
LindvedKrvang: The `else` is quite deliberate.
If you move the "if" to a new line, you are going to set the next twice if `unplannedNextPartToKeepAsNextPart` is not undefined.

PR #287, opened 2024-11-25
SOF-2353: Expose piece to action relation on pieces

KvelaGorrrrnio:
pieces now have a an optional `createdFromActionId`, which is set for inserted pieces.
The PR also ensures that setting Piece.executedAt when putting a part on air, takes the piece start offset into account.


RasmusAlbrektsen (approved):

src/presentation/dtos/piece-dto.ts
```diff
@@ -4,22 +4,28 @@ import { PieceLifespan } from '../../model/enums/piece-lifespan'
 export class PieceDto {
   public readonly id: string
   public readonly partId: string
+  public readonly rundownId: string
   public readonly name: string
   public readonly start: number
   public readonly duration?: number
+  public readonly executedAt: number
```
MadsThy, line 11: I don't think i understand the domain enough to know this, but I just wanted to ask if ´executedAt´ is obligatory? :) I just thought that since duration is optional, then maybe this would be optional too :)

MadsThy (approved):

src/presentation/dtos/piece-dto.ts
KvelaGorrrrnio: Very good observation. duration could be obligatory as well with a default value of 0. For now we keep it as is to minimize the changes up to release - but it is something that should be changed afterwards :+1:

LindvedKrvang: If you make duration mandatory with a default value of 0, then I'm 95% sure some logic elsewhere will break. FYI.

src/model/entities/piece.ts
```diff
@@ -166,11 +169,11 @@ export class Piece {
     this.timelineObjects.push(...timelineObjects)
   }
 
-  public hasEnded(): boolean {
-    if (this.executedAt === undefined) {
+  public hasEnded(timestamp: number): boolean {
```
LindvedKrvang, line 172: What's the reasoning for moving the Date check outside of the Piece?
It's a yes or no if the Piece has ended and that's not affected by external factors, so the method shouldn't be dependent on them.
All you have done is force yourself to write more code elsewhere

PR #291, opened 2024-12-05
Updates @tv2media/logger to version 2.0.3 and hides punycode deprecation warning when running in Docker.

KvelaGorrrrnio:
- The logger library is updated, such that JSON formatted logs includes a simplified ISO-8601 timestamp.
- The Dockerfile is updated to hide deprecations, since the deprecation warning makes our ELK stack unable to process logs for the given day.

Depending on the duration to the release, this commit could be beneficial to put on staging/master as a hotfix.


RasmusAlbrektsen (approved):
In my opinion this should be on the master branch (post merging staging into it) so we avoid logging issues on deploy days.

MadsThy (approved):

LindvedKrvang (approved):

PR #295, opened 2024-12-11
Add missing attributes required on PieceDto

LindvedKrvang:
Piece has gotten a new attribute `rundownId` which is required by the frontend validation. This attribute wasn't being included in the DTO, so the frontend never got the value.
The same goes for the existing attribute `executedAt`. The frontend now requires this attribute, but the backend never exposes it.


MadsThy (approved):

LindvedKrvang:
Closing the pullrequest because all I was missing was to update my develop branch

PR #298, opened 2025-01-10
SOF-2425 Prefix breakers with Effect and removing transition suffix.

vushu:
As the title suggests, also tested manually, it works.


RasmusAlbrektsen (approved):
Looks good!

LindvedKrvang (approved):

PR #305, opened 2025-02-07
Added take mode to the rundown

vushu:
I have added takeMode to rundown, tested with postman also. 

image

The change persists.

Screenshot 2025-02-07 at 13 02 02


src/business-logic/services/ingested-entity-to-entity-mapper.ts
```diff
@@ -1,14 +1,15 @@
...
+import {IngestedPart} from '../../model/entities/ingested-part'
+import {Part} from '../../model/entities/part'
+import {PartTimings} from '../../model/value-objects/part-timings'
+import {IngestedSegment} from '../../model/entities/ingested-segment'
+import {Segment} from '../../model/entities/segment'
+import {IngestedRundown} from '../../model/entities/ingested-rundown'
+import {Rundown, RundownAlreadyActiveProperties} from '../../model/entities/rundown'
+import {UnsupportedOperationException} from '../../model/exceptions/unsupported-operation-exception'
+import {RundownMode} from '../../model/enums/rundown-mode'
+import {TakeMode} from '../../model/enums/take-mode'
```
LindvedKrvang: We would like to keep the spaces around the brackets ;)

src/business-logic/services/test/ingest-rundown-synchronizer.spec.ts
```diff
@@ -1,10 +1,11 @@
...
-import { IngestedRundown } from '../../../model/entities/ingested-rundown'
-import { RundownMode } from '../../../model/enums/rundown-mode'
+import {IngestRundownSynchronizer, RundownSynchronizeResult} from '../ingest-rundown-synchronizer'
+import {IngestedEntityToEntityMapper} from '../ingested-entity-to-entity-mapper'
+import {EntityChangeDetector} from '../entity-change-detector'
+import {Rundown} from '../../../model/entities/rundown'
+import {EntityTestFactory} from '../../../model/entities/test/entity-test-factory'
+import {IngestedRundown} from '../../../model/entities/ingested-rundown'
+import {RundownMode} from '../../../model/enums/rundown-mode'
+import {TakeMode} from '../../../model/enums/take-mode'
```
LindvedKrvang: We would also like to keep these spaces ;)

src/business-logic/services/test/ingest-rundown-synchronizer.spec.ts
```diff
@@ -51,6 +52,17 @@ describe(IngestRundownSynchronizer.name, () => {
         const result: RundownSynchronizeResult = testee.synchronizeRundown(rundown, ingestedRundown)
 
         expect(result.updatedRundown?.getMode()).toBe(RundownMode.ACTIVE)
+        expect(result.updatedRundown?.getTakeMode()).toBe(TakeMode.STANDARD)
```
LindvedKrvang: We would like to keep our test to only test _one_ thing.
This test tests to see if we return with the correct RundownMode. This is different than TakeMode. The expectation for the TakeMode shouldn't be in this test. Instead, if you want to test for _this_ scenario, you should make a separate test for it.

Imagine you are looking at a test report and see a failing test called `Returns a rundown with the same rundown mode` - then you think: "Yes, something is wrong with the RundownMode. Lets get to work" and you start debugging that. After an hour you realize it actaully have nothing to do with the RundownMode at all, but someone added an unrelated assertion in the test.

Ideally, we should be able to look at the name of the failing test and then have a pretty good idea of what is wrong.

src/business-logic/services/test/ingest-rundown-synchronizer.spec.ts
```diff
@@ -51,6 +52,17 @@ describe(IngestRundownSynchronizer.name, () => {
...
+      })
+
+      it('returns a rundown with the same take mode', () => {
```
LindvedKrvang: This test is put under the describe `when the Rundown mode has changed`. The RundownMode doesn't have anything to do with the TakeMode and the TakeMode is never affected by the RundownMode (In fact, ingest is currently not even able to change the TakeMode, because it doesn't even know that it exists).

I think the interesting scenario to test here is when a Rundown is updated throug ingest, that it keeps whatever TakeMode it had and that the TakeMode isn't reset.

src/presentation/controllers/rundown-controller.ts
```diff
@@ -79,6 +80,22 @@ export class RundownController extends BaseController {
     }
   }
 
+  @PutRequest('/:rundownId/takeMode/:takeMode')
+  public async takeMode(request: Request, response: Response): Promise<void> {
+    try {
+      const rundownId: string = request.params.rundownId
+      const takeMode: TakeMode = TakeMode[request.params.takeMode as keyof typeof TakeMode]
+      if (takeMode === undefined) {
+        response.send(this.httpResponseFormatter.formatFailResponse(`Rundown "${rundownId}" failed to set it's Take Mode, since "${request.params.takeMode}" isn't a valid input.` ))
```
LindvedKrvang: Perhaps we should also log if this happens?

LindvedKrvang (approved):
Nice work :)
The approve assumes that you will take a look at the comments for the tests.

src/business-logic/services/test/ingest-rundown-synchronizer.spec.ts
vushu: I think I have to set my IDE correctly. :p

src/business-logic/services/test/ingest-rundown-synchronizer.spec.ts
vushu: I see, good point.

src/business-logic/services/test/ingest-rundown-synchronizer.spec.ts
vushu: Good point!

src/model/entities/rundown.ts
```diff
@@ -28,6 +28,7 @@ import { InvalidPartException } from '../exceptions/invalid-part-exception'
 import { SetNextDirection } from '../enums/set-next-direction'
 import { FirstPartInSegmentException } from '../exceptions/first-part-in-segment-exception'
 import { FirstSegmentInRundownException } from '../exceptions/first-segment-in-rundown-exception'
+import {TakeMode} from '../enums/take-mode'
```
LindvedKrvang: We would like spaces around the brackets :)

src/presentation/controllers/rundown-controller.ts
```diff
@@ -79,6 +83,24 @@ export class RundownController extends BaseController {
...
 
+  @PutRequest('/:rundownId/takeMode/:takeMode')
+  public async takeMode(request: Request, response: Response): Promise<void> {
+    try {
+      const rundownId: string = request.params.rundownId
+      const takeMode: TakeMode = TakeMode[request.params.takeMode as keyof typeof TakeMode]
+      if (takeMode === undefined) {
+        const errorMessage: string = `Rundown "${rundownId}" failed to set it's Take Mode, since "${request.params.takeMode}" isn't a valid input.`
+        this.logger.error(errorMessage)
+        response.send(this.httpResponseFormatter.formatFailResponse(errorMessage))
```
LindvedKrvang: This returns a status code 200, which isn't right when the request failed.
It should be a BadRequest since we are calling the endpoint with invalid parameters.

LindvedKrvang (changes requested):
Still good work.

The reason for the request changes is because, after a second look, I realized that you have yet to implement your task in it's entirety.
You are missing this section from your task description:
image

vushu:
@LindvedKrvang ah yeah thats right, I take a look at it :)

vushu:
I have added tests that check that we indeed will get a rundown updated event.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -42,6 +43,17 @@ export class RundownTimelineService implements RundownService {
...
 
+  public async setTakeMode(rundownId: string, takeMode: TakeMode): Promise<void> {
+    const rundown: Rundown = await this.rundownRepository.getRundown(rundownId)
+    if (takeMode === rundown.getTakeMode()) {
+      this.logger.debug(`Rundown: ${rundown.id} already has a takeMode of ${takeMode}.`)
+      return
+    }
+    rundown.setTakeMode(takeMode)
+    await this.saveRundown(rundown)
+    this.rundownEventEmitter.emitRundownUpdated(rundown)
```
LindvedKrvang: It would be nice if the `saveRundown(...)` was the very last call. Or the very least after we emit the event.

By emitting the event before saving the Rundown, we don't have to wait for the database transactions before we notify our clients and thus are our clients getting a faster response time :)

LindvedKrvang (approved):
The approve assumes you have seen the comment about the order of emitting and saving :)

src/business-logic/services/rundown-timeline-service.ts
vushu: I have done the change :), but wouldn't it be more certain to emit after save rundown?, so that we know for certain that we successfully saved the rundown? That's what I had in mind. in the case that saveRundown throws an error, we won't emit.

RasmusAlbrektsen (approved):

PR #308, opened 2025-02-11
Sof 2414/added recall logic to service

vushu:
Added recall logic to service.

creation of recall part I put in the service, so that the business logic stays within the service.

Still unsure about the `insertPartAsOnAir` so please guide me.

I could understand is that it's no trivial to filter relevant pieces, so for now I didn't filter it.


src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -27,6 +27,7 @@ import { TakeMode } from '../../model/enums/take-mode'
 
 export class RundownTimelineService implements RundownService {
   private readonly logger: Logger
+  private recallPart: Part | undefined
```
LindvedKrvang: We are in a "service". Services should be stateless. By adding an instance variable you have now added state to the service.
The service could be called to do a Take for Rundown A and then immediatly be called for Rundown B. The `recallPart` variable would not be from both Rundowns (it might not even be from any of them).

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -188,6 +193,49 @@ export class RundownTimelineService implements RundownService {
     }
   }
 
+  private shouldRecallPart(rundown: Rundown): void {
```
LindvedKrvang: You have prefixed your method with `should`. `should` and `is` implies questions and questions implies a yes or no answer i.e. methods named `shouldX` or `isX` should return a boolean, not void.

There is a mismatch in your method name and your implementation. Your method name indicates that you only check whether you should recall or not. Your implementation actually _does_ the recalling. These are two different things.

To separate the responsiblites, and make the naming accurate, you could do something like this:
```
private shouldRecallPart(rundown: Rundown): boolean {
  return rundown.getTakeMode() === TakeMode.RECALL
}

private recallPreviousPart(rundown: Rundown): void {
  const previousPart: Part | undefined = rundown.getPreviousPart()
  if (!previousPart) {
    return
  }
  const recalledPart: Part = previousPart.getRecallClone()
  rundown.insertPartAsNext(recalledPart)
}
```
The above snippet could then be used like:
```
if (this.shouldRecallPart(rundown)) {
  this.recallPreviousPart(rundown)
}
```

Notice the naming: 
The method that checks asks a question that gives an answer, but don't actually do anything. It just checks the current state. 
The method that actually do something has a name that indicates an action which will change the state. This is most often used by simply using verbs.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -188,6 +193,49 @@ export class RundownTimelineService implements RundownService {
...
+    }
+
+    const previousPart: Part | undefined = rundown.getPreviousPart()
+    if (previousPart) {
+      this.recallPart = this.createRecallPart(rundown.id, previousPart)
+      rundown.insertPartAsNext(this.recallPart)
+    }
+  }
+
+  private shouldEmitRecallPart(rundown: Rundown): void {
```
LindvedKrvang: Same comment about that this is named to indicate that it simply check the state, but the implementation actually alters the state, giving a mismatch between naming and implementation.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -188,6 +193,49 @@ export class RundownTimelineService implements RundownService {
...
+    if (this.recallPart) {
+      this.rundownEventEmitter.emitPartInsertedAsNextEvent(rundown, this.recallPart)
+    }
+  }
+
+  private createRecallPart(rundownId: string, part: Part): Part {
```
LindvedKrvang: The responsibility of the `rundown-timeline-service.ts`, which we are in, is to facilitate flows. 
This method actually "do something" which doesn't have anything to do with flow so it doesn't belong in this file.

Cloning/copying a Part (which this method does) is the responsibility of `Part.ts`. If you look in that file you will notice that there already is a `clon()` and `getUnsyncedCopy()` method in there.

Sidenote: `Part` already have the `rundownId` of which is belong, so you don't need to parse `rundownId` as a parameter. You can just get it from `part.getRundownId()`.

src/business-logic/services/rundown-timeline-service.ts
```diff
@@ -315,6 +363,8 @@ export class RundownTimelineService implements RundownService {
       rundown.setNextFromIds(nextCursor.segment.id, nextCursor.part.id, nextCursor.owner)
     }
 
+    this.shouldEmitRecallPart(rundown)
```
LindvedKrvang: By adding this method you potentially emit that you have inserted a new Part, but there is no code to actually insert the new Part. So you are potentially emitting that you have done something without having done anything.

The `unplannedNextPartToKeepAsNextPart` might need some explanation. Reach out if you need a deeper explanation.
Basically, we either need to do the `Recall` or the `unplannedNextPartToKeepAsNextPart`. I think that Recall takes priority.

src/model/entities/part.ts
```diff
@@ -63,7 +63,7 @@ export class Part {
   private segmentId: string
   private rank: number
 
-  private pieces: Piece[]
+  public pieces: Piece[]
```
LindvedKrvang: This should not be public. Making it public breaks encapsulation.
If you need the Pieces of the Part then use `part.getPieces()`.

src/business-logic/services/rundown-timeline-service.ts
vushu: Thats also what I thought, since I wasn't sure about insertPartAsOnAir logic.

src/business-logic/services/rundown-timeline-service.ts
vushu: I see that Part also may contain business logic, wasn't sure about that.

src/business-logic/services/rundown-timeline-service.ts
vushu: I wanted to make a single function, I understand you point.

vushu:
I believe I have done the necessary changes and this PR is ready for review,
I have made another task for filtering pieces out.
[SOF-2477](https://tv2cms.atlassian.net/browse/SOF-2477)

[SOF-2477]: https://tv2cms.atlassian.net/browse/SOF-2477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

LindvedKrvang (approved):
Looks good

RasmusAlbrektsen (approved):

PR #309, opened 2025-02-11
Added take mode recall and standard as actions

vushu:
I have added standard and recall as actions.


src/business-logic/services/execute-action-service.ts
```diff
@@ -177,6 +200,14 @@ export class ExecuteActionService implements ActionService {
         await this.rundownService.setNextFromDirection(rundownId, SetNextDirection.SEGMENT_BEFORE_NEXT_SEGMENT)
         break
       }
+      case SystemActionId.SET_TAKE_MODE_STANDARD: {
+        await this.rundownService.setTakeMode(rundownId,TakeMode.STANDARD)
```
LindvedKrvang: Would be nice with a space between the parameters when calling the function :)
Goes for line 208 as well.

LindvedKrvang (approved):
Looks good and works well :)

RasmusAlbrektsen (approved):

PR #311, opened 2025-02-13
Sof 2477/strip pieces when recalling

LindvedKrvang:
Only include the Pieces relevant for the Recall part of the Part. 
Pieces that are included are being copied, so they don't have the original reference in the heap.

Also in this PR:
- Refactored `Tv2PieceMetadata` to `PieceMetadata` and it is no longer optional on Piece. This change affects a lot of files making this PR big. The changes that actually implements SOF-2477 is in this commit: https://github.com/tv2/sofie-server/pull/311/commits/c2b82a7a450f7a61c4148471bb2816ad4339cc78


vushu (approved):
I see you are making the data less tv2 specific and making meta data an actual type.
LGTM.

vushu (approved):

RasmusAlbrektsen (approved):

PR #312, opened 2025-02-18
Added Macro CRUD

vushu:
I made the basic CRUD operation for macro.

image


src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -198,6 +199,12 @@ export interface MongoAction extends MongoId {
   argument?: ActionArgument
 }
 
+export interface MongoMacro extends MongoId {
```
LindvedKrvang: The entire idea of "MongoEntities" and the `mongo-entity-converter` is because some of our entities/data are being managed by Core and Core manages it in a way we don't agree with (weird aggregations of data. Naming all over the place etc).

The responsibility of the `mongo-entity-converter` is to map data back and forth between how Core wants it and how Alba wants it so that both Core and Alba can work with the data.

This means, that this converter (and by extension the "MongoEntities") are only needed for data managed by Core. `Macro` is an Alba only entity and therefore we don't need a `MongoMacro`.

I would like to see this MongoMacro removed again :)

LindvedKrvang (approved):
Looks good :)
My only comment is about the MongoMacro

src/data-access/repositories/mongo/mongo-entity-converter.ts
vushu: I have now removed MongoMacro, since the plan is phasing out core at some point, so it makes sense :)

LindvedKrvang (approved):

src/presentation/controllers/macro-controller.ts
```diff
@@ -0,0 +1,86 @@
...
+        id: macroDto.id,
+      }
+      await this.macroService.updateMacro(macroDto)
+      response.send(this.httpResponseFormatter.formatSuccessResponse(`Successfully updated Macro ${macro.id}`))
+    } catch (error) {
+      this.httpErrorHandler.handleError(response, error as Exception)
+    }
+  }
+
+
```
RasmusAlbrektsen, line 74: Two extra newlines here we should remove :-)

RasmusAlbrektsen (approved):

src/presentation/controllers/macro-controller.ts
vushu: I will remove it in next PR :D

PR #315, opened 2025-02-20
Sof 2486/will emit events on macro crud

vushu:
Added Macro events. This PR depends on https://github.com/tv2/sofie-server/pull/312


RasmusAlbrektsen (approved):

src/business-logic/services/test/macro-service-implementation.spec.ts
```diff
@@ -0,0 +1,55 @@
+import { MacroServiceImplementation } from '../macro-service-implementation'
+import { MacroRepository } from '../../../data-access/repositories/interfaces/macro-repository'
+import { instance, mock, verify, when } from '@typestrong/ts-mockito'
+import { MacroEventEmitter } from '../interfaces/macro-event-emitter'
+import { Macro } from '../../../model/entities/macro'
+
+describe(MacroServiceImplementation.name, () => {
+  describe(MacroServiceImplementation.prototype.createMacro.name, () => {
+    it('should emit macro created event', async () => {
+      const macro: Macro = { id: 'FakeMacroId', name: 'FakeMacro', operations: [] } as Macro
```
LindvedKrvang, line 10: You should update the `EntityTestFactory` with a `createMacro()` so that all you have to do here is to call:
```
const macro: Macro = EntityTestFactory.createMacro()
```

LindvedKrvang (approved):
Looks good :D 

My only comment is about that you should update and use `EntityTestFactory`.

src/business-logic/services/test/macro-service-implementation.spec.ts
vushu: I will remember to do this.

PR #317, opened 2025-02-21
SOF-2501/dsk-action-doesnt-apply-for-tricaster

RasmusAlbrektsen:
The Tricaster DSK on and off action timeline objects did not have the correct priority and was therefore overwritten by the applied graphics timeline objects that were currently on air. Therefore it actually worked so long as no graphic was shown (which of course was not that great).


LindvedKrvang (approved):

vushu (approved):
cool

PR #321, opened 2025-02-26
Added macro execution and tests

vushu:
We have discussed if an error occurs then we could notify the user, I took the liberty to make a new emit message called MacroOperationFailedEvent for this occasion.

Below a demonstration using postman, here I do a double take with 2seconds delay.

https://github.com/user-attachments/assets/e5da89b3-d03e-4e2b-9dd2-74a332b1e9e1


src/business-logic/facades/service-facade.ts
```diff
@@ -81,7 +81,7 @@ export class ServiceFacade {
   }
 
   public static createMacroService(): MacroService {
-    return new MacroServiceImplementation(EventEmitterFacade.createMacroEventEmitter(), RepositoryFacade.createMacroRepository())
+    return new MacroServiceImplementation(EventEmitterFacade.createMacroEventEmitter(), RepositoryFacade.createMacroRepository(), this.createActionService())
```
LindvedKrvang: To keep consistency in this file please use `ServiceFacade.createActionService()` instead of `this.createActionService()`

src/business-logic/services/interfaces/macro-event-emitter.ts
```diff
@@ -4,4 +4,5 @@ export interface MacroEventEmitter {
   emitMacroCreatedEvent(macro: Macro): void
   emitMacroUpdatedEvent(macro: Macro): void
   emitMacroDeletedEvent(macroId: string): void
+  emitMacroOperationFailedEvent(macro: Macro, operationIndex: number, message: string): void
```
LindvedKrvang: We have StatusMessages to emit information to the clients. Please use those. 
Adding a new emit event for every single thing that we wan't to communicate is going to give us tons of upkeep.

src/business-logic/services/macro-service-implementation.ts
```diff
@@ -31,4 +36,38 @@ export class MacroServiceImplementation implements MacroService {
...
+  }
+
+  private async executeOperationAtIndex(rundownId: string, macro: Macro, index: number): Promise<void> {
+    if (index >= macro.operations.length) {
+      return
+    }
+    try {
+      await this.executeOperation(rundownId, macro.operations[index])
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
+      setTimeout(() => this.executeOperationAtIndex(rundownId, macro, index + 1), macro.operations[index].delayNextOperationMs)
```
LindvedKrvang: We should try to avoid disabling eslint.
Instead of disabling it, you can add a `.catch()` to `executeOperationAtIndex`, so it could become
```
setTimeout(() => {
  this.executeOperationAtIndex(...).catch( () => // error handling here //)
}, ... )
```

src/business-logic/services/macro-service-implementation.ts
```diff
@@ -31,4 +36,38 @@ export class MacroServiceImplementation implements MacroService {
...
+    const macro: Macro = await this.macroRepository.getMacro(macroId)
+    await this.executeOperationAtIndex(rundownId, macro, 0)
+  }
+
+  private async executeOperationAtIndex(rundownId: string, macro: Macro, index: number): Promise<void> {
+    if (index >= macro.operations.length) {
+      return
+    }
+    try {
+      await this.executeOperation(rundownId, macro.operations[index])
```
LindvedKrvang: It would be nice that instead of keeping to look up the Operation then simply save it as a local variable:
```
const operation: Operation = macro.operations[index]
await this.executeOperation(rundownId, operation)
setTimeout(() => this.executeOperationAtIndex(rundownId, macro, index + 1), operation.delayNextOperationInMs)
```

src/model/entities/test/entity-test-factory.ts
```diff
@@ -225,4 +226,8 @@ export class EntityTestFactory {
       ...action
     }
   }
+
+  public static createMacro(operations?: Operation[]): Macro {
```
LindvedKrvang: Please create this the same way as the other methods in this file.

The idea of these methods is that you can get an entity without providing any data. If you then need some specific data for a given test, you can provide just that data.
That means, that it should be possible to provide nothing as an argument, or a single (or multiple) attributes.

This can be achieved by the spread operator and `Partial<T>`:
```
createmacro(macro?: Partial<Macro>): Macro {
  return {
    id: 'randomId',
    name: 'randomName',
    operations: [],
    ...macro
  }
```

To use this you can now call it like:
```
EntityTestFactory.createMacro()
EntityTestFactory.createMacro({ id: 'mySpecificId' })
EntityTestFactory.createMacro({ id: 'myOtherSpecificId', name: 'thisIsAName' })
EntityTestFactory.createMacro({ operations: [operationOne, operationTwo] })
```

src/presentation/controllers/macro-controller.ts
```diff
@@ -83,4 +83,16 @@ export class MacroController extends BaseController {
       this.httpErrorHandler.handleError(response, error as Exception)
     }
   }
+
+  @PutRequest('/execute/:macroId/rundown/:rundownId')
```
LindvedKrvang: We are following REST conventions. That means that entity names should be plural i.e. instead of `rundown` it should say `rundowns`.

LindvedKrvang (commented):
Good work :) 

There is some stuff that needs addressing.

I don't like where the path of `emitMacroOperationFailedEvent` is going as it will require a new event for every single tiny bit of information we want to send to our clients. Instead we should have a "central/single" system to send information to clients. We do have this today already through StatusMessages.

src/business-logic/services/macro-service-implementation.ts
vushu: I still get the same linter warning, after adding the catch, at the time I ended up disabling it, since I don't believe investing too much time is worth the effort I sadly had already spend too much time on it.

src/business-logic/services/interfaces/macro-event-emitter.ts
vushu: I tried looking at status message, I doesn't seem to cover the case where I wan't to warn the user with more info, example I want to tell the user that this exactly operation failed, by returning the operations and the index of the failed one, whereas status message only allows for string messages, I will use the status message for simplicities sake, then when we need more information we change it.

src/model/entities/test/entity-test-factory.ts
vushu: Forgot the spread operator :)

src/business-logic/services/macro-service-implementation.ts
RasmusAlbrektsen: Yeah it seems vexing to make it work the way it is setup right now, i haven't been able to solve this in a non time consuming manner.

src/business-logic/services/macro-service-implementation.ts
```diff
@@ -31,4 +36,38 @@ export class MacroServiceImplementation implements MacroService {
...
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
+      setTimeout(() => this.executeOperationAtIndex(rundownId, macro, index + 1), macro.operations[index].delayNextOperationMs)
+    }
+    catch (error: unknown) {
+      let errorMessage: string = 'Operation failed for unknown reason'
```
RasmusAlbrektsen: A personal an minor note, i like when we add a period after sentences like this. So:
`The operation failed for an unknown reason.`

RasmusAlbrektsen: If you feel like changing it, it would also apply for the exception you throw on line 70.

RasmusAlbrektsen (approved):
This approve is also dependent on the fixes that @LindvedKrvang  has asked for :-) 
I would like it if we didn't suppress the linter too.

src/business-logic/services/macro-service-implementation.ts
LindvedKrvang: Did you try the provided example? That should work

src/business-logic/services/macro-service-implementation.ts
```diff
@@ -31,4 +39,41 @@ export class MacroServiceImplementation implements MacroService {
...
+      {
+        this.executeOperationAtIndex(rundownId, macro, index + 1).catch((err: Error) => {throw err})
+      }, operation.delayNextOperationMs)
+    }
+    catch (error: unknown) {
+      let errorMessage: string = 'The operation failed for an unknown reason.'
+      if (error instanceof Exception) {
+        errorMessage = error.message
+      }
+      this.statusMessageEventEmitter.emitStatusMessageEvent({id: 'OperationFailed', message: errorMessage, statusCode: StatusCode.BAD, title: 'Operation Failed'})
```
LindvedKrvang: Given how the StatusMessages currently work, if you give it anything but an `INFO` status code, the StatusMessage is going to be persisted untill it receives an `INFO` StatusMessage with the same `id`.

What this means for this implementation is, if there is an error, then we are going to get a StatusMessage that never goes away. (It's not great, but we unfortunately need that behaviour for Device statusses untill we control the connection to our Devices...)

So if you want to use StatusMessage, you should give it the status code `INFO`.

Minor thing/perosnal preference: It's quite weird to see an `id` with an uppercased first character :p

src/business-logic/services/test/macro-service-implementation.spec.ts
```diff
@@ -1,55 +1,127 @@
...
 import { MacroEventEmitter } from '../interfaces/macro-event-emitter'
-import { Macro } from '../../../model/entities/macro'
+import { Macro, Operation, OperationType } from '../../../model/entities/macro'
+import { EntityTestFactory } from '../../../model/entities/test/entity-test-factory'
+import { ActionService } from '../interfaces/action-service'
+import { Exception } from '../../../model/exceptions/exception'
+import { ErrorCode } from '../../../model/enums/error-code'
+import {StatusMessageEventEmitter} from '../interfaces/status-message-event-emitter'
 
 describe(MacroServiceImplementation.name, () => {
```
LindvedKrvang, line 12: For all of your tests in this file you are setting up mock for all the dependencies of `MacroServiceImplementation`, even the dependencies that are not used/needed for the specific test.

We should try to keep our tests to only have relevant information for the specific test at hand. If not, the tests has a tendency to become bloated quickly, which makes them difficult to read/understand.

I will highly suggest that you create a form of `createTestee` method, that will by default set up all required dependencies as mock, unless a mock was provided as an argument. Take a look at our other test classes for inspiration.

And just to give another reason to implement this: Imagine (the very likely thing) that we add another dependency to `MacroServiceImplementation`. You would now have to update _every single test_ just to make the code compile. That's a very high maintenance burden.

src/business-logic/services/test/macro-service-implementation.spec.ts
```diff
@@ -1,55 +1,127 @@
...
+  })
+})
+
+describe(MacroServiceImplementation.prototype.executeMacro.name, () => {
+  it('should emit macro operation failed event', async () => {
+    const operations: Operation[] = [{
+      type: OperationType.ACTION,
+      delayNextOperationMs: 200,
+      actionId: 'MyActionId'
+    }]
```
LindvedKrvang: You should also create a `TestEntityFactory.createOperation()` method to help with your test setup :)

src/business-logic/services/test/macro-service-implementation.spec.ts
```diff
@@ -1,55 +1,127 @@
...
+    }, {
+      type: OperationType.ACTION,
+      delayNextOperationMs: 200,
+      actionId: 'TAKE'
+    }, {
+      type: OperationType.ACTION,
+      delayNextOperationMs: 0,
+      actionId: 'Something'
+    }]
+    const macro: Macro = EntityTestFactory.createMacro({operations: operations})
```
LindvedKrvang, line 88: Minor thing: If the attribute and the variable being assigned has the same name, you can suffice just to write the name once: `EntityTestFactory.createmacro({ operations })`.

src/presentation/enums/event-type.ts
```diff
@@ -50,7 +50,8 @@ export enum ActionTriggerEventType {
 export enum MacroEventType {
   MACRO_CREATED = 'MACRO_CREATED',
   MACRO_UPDATED = 'MACRO_UPDATED',
-  MACRO_DELETED = 'MACRO_DELETED'
+  MACRO_DELETED = 'MACRO_DELETED',
+  MACRO_OPERATION_FAILED = 'MACRO_OPERATION_FAILED'
```
LindvedKrvang: If I read the PR correctly, then this needs to be deleted?

src/presentation/value-objects/macro-event.ts
```diff
@@ -18,3 +18,10 @@ export interface MacroDeletedEvent extends TypedEvent {
   type: MacroEventType.MACRO_DELETED
   macroId: string
 }
+
+export interface MacroOperationFailedEvent extends TypedEvent {
+  type: MacroEventType.MACRO_OPERATION_FAILED
+  macro: MacroDto,
+  operationIndex: number,
+  message: string
+}
```
LindvedKrvang: If I read the PR correctly, then this needs to be deleted?

LindvedKrvang (commented):
Still good work :)
We are almost there for the approve

src/business-logic/services/macro-service-implementation.ts
vushu: There doesn't seem to be an Info but only , Good, Bad, Warning and Unknown ,which one is preferred? for now i use Good.

src/business-logic/services/macro-service-implementation.ts
vushu: I works now thanks.

RasmusAlbrektsen (approved):
Good work :-)

src/business-logic/services/macro-service-implementation.ts
LindvedKrvang: Ahh. I meant the `GOOD` status :)

LindvedKrvang (approved):

PR #322, opened 2025-03-03
Sof 2508/changed action triggers to be generic triggers

vushu:
I have not yet changed the db collection name to triggers and the controller endpoint to be triggers, I will do this later on in another PR, so that we can merge this into develop without breaking the frontend.


src/data-access/repositories/mongo/mongo-trigger-repository.ts
```diff
@@ -0,0 +1,73 @@
...
+      case TriggerType.ACTION: {
+        return this.createNewActionTrigger(triggerWithoutId)
+      }
+      case TriggerType.MACRO: {
+        return this.createNewMacroTrigger(triggerWithoutId)
+      }
+    }
+  }
+
+  private async createNewActionTrigger(triggerWithoutId:  Omit<Trigger, 'id'>): Promise<ActionTrigger> {
```
LindvedKrvang: Why the `new` filler word? That's implicit by `create`.

Same comment goes for Macro

LindvedKrvang: `createNewActionTrigger` and `createNewMacroTrigger` are "identical".

The only difference is that you cast them to different types, but you don't do anything that is either Macro nor Action specific. And since TypeScript compiles to JavaScript casting an object doesn't really "do" anything. Consider the following snippet:
```
interface Dog {
  bark: string
}

interface Cat {
  meow: string
}

type Animal = Dog | Cat

function printAnimal(animal: Animal): void {
  const dog: Dog = animal as Dog
  console.log(dog)
}

const cat: Cat = { meow: 'Random Meow' }
printAnimal(cat)
```
The output of the above snippet is going to be `{ meow: 'Random Meow' }` Even though we cast it to `Dog`, it doesn't mean that all attributes that are not on Dog are being stripped from the object. It just means that we can only _access_ the attributes on Dog.

The same applies for you methods here. And since you are not actually doing any type checking, this becomes valid and executable code:
image

src/model/entities/trigger.ts
```diff
@@ -0,0 +1,23 @@
+export type Trigger = ActionTrigger | MacroTrigger
+
+interface BaseTrigger {
+  id: string
+  type: TriggerType
+  data: unknown // Defined by whatever frontend that needs to have an ActionTrigger
```
LindvedKrvang: You should update the comment so it doesn't say `ActionTrigger` but `Trigger` instead

src/model/entities/trigger.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export interface ActionTrigger extends BaseTrigger {
+  type: TriggerType.ACTION
+  actionId: string
+  actionArguments: string | number
```
LindvedKrvang: `actionArguments` are optional so you should add a `?` to it.

src/presentation/controllers/trigger-controller.ts
```diff
@@ -0,0 +1,108 @@
...
+    private readonly httpResponseFormatter: HttpResponseFormatter
+  ) {
+    super()
+  }
+
+  @GetRequest()
+  public async getTriggers(_request: Request, response: Response): Promise<void> {
+    try {
+      const triggers: Trigger[] = await this.triggerService.getTriggers()
+      response.send(this.httpResponseFormatter.formatSuccessResponse(triggers.map(trigger => new ActionTriggerDto(trigger as ActionTrigger))))
```
LindvedKrvang: This should also support `MacroTrigger`.

Also: `trigger as ActionTrigger` is not a type check. You are basically telling TypeScript: "I know better, this is an ActionTrigger, so treat it like one". If it is not an ActionTrigger, then you risk a RuntimeError inside ActionTriggerDto

src/data-access/repositories/mongo/mongo-trigger-repository.ts
LindvedKrvang: To further demonstrate this, this:
image
Is without question saved as:
image

src/presentation/controllers/trigger-controller.ts
```diff
@@ -0,0 +1,108 @@
...
+        case TriggerType.ACTION:
+          trigger = this.mapToActionTrigger(triggerDto)
+          break
+        case TriggerType.MACRO:
+          trigger = this.mapToMacroTrigger(triggerDto)
+          break
+        default:
+          this.respondWithUnexpectedType(triggerDto, response)
+          return
+      }
```
LindvedKrvang: We need better type checking here. It currently accepts this:
image
Which it should not.

src/presentation/controllers/trigger-controller.ts
```diff
@@ -0,0 +1,108 @@
...
```
LindvedKrvang: This needs to be updated the same way the create endpoint also needs it

src/presentation/controllers/trigger-controller.ts
```diff
@@ -0,0 +1,108 @@
...
+      type: TriggerType.ACTION,
+    }
+  }
+
+  private mapToMacroTrigger(triggerDto: Partial<TriggerDto>): Trigger {
+    return {
+      ...triggerDto as MacroTriggerDto,
+      type: TriggerType.MACRO,
+    }
+  }
```
LindvedKrvang: As mentioned in my other comments, using `as` is not really mapping anything.
In fact, the `as` keyword should be minimized in use for production code as it is basically a way to tell TypeScript to ignore type checking, which is a major reason for using TypeScript.

src/presentation/controllers/trigger-controller.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  @DeleteRequest('/:actionTriggerId')
```
LindvedKrvang: It should say `/:triggerId`

src/presentation/controllers/trigger-controller.ts
```diff
@@ -0,0 +1,108 @@
...
+import { HttpErrorHandler } from '../interfaces/http-error-handler'
+import { Exception } from '../../model/exceptions/exception'
+import { Request, Response } from 'express'
+import { ActionTrigger, Trigger, TriggerType } from '../../model/entities/trigger'
+import { ActionTriggerDto, MacroTriggerDto, TriggerDto } from '../dtos/trigger-dto'
+import { TriggerService } from '../../business-logic/services/interfaces/trigger-service'
+import { HttpResponseFormatter } from '../interfaces/http-response-formatter'
+import { UnexpectedCaseException } from '../../model/exceptions/unexpected-case-exception'
+
+@RestController('/actionTriggers')
```
LindvedKrvang: This needs to be updated as part of this PR.
You will need to create a PR for the frontend that has this change to be merged at the same time as this PR.

src/data-access/repositories/mongo/mongo-trigger-repository.ts
```diff
@@ -0,0 +1,73 @@
+import { BaseMongoRepository } from './base-mongo-repository'
+import { TriggerRepository } from '../interfaces/trigger-repository'
+import { ActionTrigger, MacroTrigger, Trigger, TriggerType } from '../../../model/entities/trigger'
+import { MongoDatabase } from './mongo-database'
+import { UuidGenerator } from '../interfaces/uuid-generator'
+import { NotFoundException } from '../../../model/exceptions/not-found-exception'
+import { MongoId } from './mongo-entity-converter'
+
+const ACTION_TRIGGER_COLLECTION: string = 'actionTriggers'
```
LindvedKrvang: No reason not to update this now. It's part of this task.
Remember to have a migration that moves all the Triggers from the `actionTriggers` collection into the new collection.

src/presentation/enums/event-type.ts
```diff
@@ -41,10 +41,10 @@ export enum ActionEventType {
...
 }
 
-export enum ActionTriggerEventType {
-  ACTION_TRIGGER_CREATED = 'ACTION_TRIGGER_CREATED',
-  ACTION_TRIGGER_UPDATED = 'ACTION_TRIGGER_UPDATED',
-  ACTION_TRIGGER_DELETED = 'ACTION_TRIGGER_DELETED'
+export enum TriggerEventType {
+  TRIGGER_CREATED = 'TRIGGER_CREATED',
+  TRIGGER_UPDATED = 'TRIGGER_UPDATED',
+  TRIGGER_DELETED = 'TRIGGER_DELETED'
```
LindvedKrvang, line 47: These requires a PR that also have this change for the frontend.

src/presentation/interfaces/trigger-event-observer.ts
```diff
@@ -0,0 +1,5 @@
+import { TriggerEvent } from '../value-objects/trigger-event'
+
+export interface TriggerEventObserver {
+  subscribeToTriggerEvents(onTriggerEventCallback: (actionTriggerEvent: TriggerEvent) => void): void
```
LindvedKrvang: Update `actionTriggerEvent` to `triggerEvent`

src/presentation/services/event-builder.ts
```diff
@@ -318,27 +318,27 @@ export class EventBuilder implements RundownEventBuilder, ActionEventBuilder, Ac
...
 
-  public buildActionTriggerDeletedEvent(actionTriggerId: string): ActionTriggerDeletedEvent {
+  public buildTriggerDeletedEvent(actionTriggerId: string): TriggerDeletedEvent {
     return {
-      type: ActionTriggerEventType.ACTION_TRIGGER_DELETED,
+      type: TriggerEventType.TRIGGER_DELETED,
       timestamp: Date.now(),
-      actionTriggerId,
+      triggerId: actionTriggerId,
     }
```
LindvedKrvang: All of these are missing to have the parameter name updated.

LindvedKrvang (changes requested):
Good start, but the hard part has yet to be done. (Proper type checking between Action- and MacroTriggers)

There is no need to not change the collection name and the controller endpoints. You already broke the frontend with the changes to `TriggerEventType`.

I don't see any migrations that ensure that all current ActionTriggers now have a `type` attribute with `TriggerType.ACTION`. At the same time, I don't see any migration to move `actionArguments` from `data` on the ActionTriggers that needs this.

src/presentation/enums/event-type.ts
vushu: noted

vushu:
migrations are on the way.

vushu:
Migrations are done.

src/data-access/facades/repository-facade.ts
```diff
@@ -253,8 +253,8 @@ export class RepositoryFacade {
     return new MongoMacroRepository(MongoDatabase.getInstance(LoggerFacade.createLogger()), this.createUuidGenerator())
   }
 
-  public static createActionTriggerRepository(): ActionTriggerRepository {
-    return new MongoActionTriggerRepository(MongoDatabase.getInstance(LoggerFacade.createLogger()), this.createUuidGenerator())
+  public static createActionTriggerRepository(): TriggerRepository {
```
LindvedKrvang: You are returning a `MongoTriggerRepository`, so the name of this method needs to be updated to `createTriggerRepository` instead of `createActionTriggerRepository`.

src/data-access/migrations/mongo/mongo-migrations/20250305065128-add-action-arguments-for-action-triggers.ts
```diff
@@ -0,0 +1,19 @@
+import { Db } from 'mongodb'
+
+export async function up(db: Db): Promise<void> {
+  await db.collection('actionTriggers').updateMany(
+    { type: 'ACTION' },
+    {
+      $set: { actionArguments: 0 }
```
LindvedKrvang: This just blindly sets all `actionArguments` on all ActionTriggers to zero.
An ActionTrigger that does not require an ActionArgument should _not_ have a value for ActionArguments i.e. it should be undefined.

The purpose of this migration is to not break our already configured ActionTriggers that have an ActionArgument. By blindly setting all ActionArguments to zero, the ActionTriggers are still broken.

This migration needs to filter through all ActionTriggers data and _only_ update the ones that actually have ActionArguments. And it needs to update it to the value already saved as ActionArguments. Then it should delete the ActionArgument that is in the `data` blob.

LindvedKrvang: I can see you have created three separate migrations for what should essentially be one migration

src/presentation/dtos/trigger-dto.ts
```diff
@@ -0,0 +1,77 @@
...
+  public data: unknown
+
+  public static createTriggerDto(trigger: Trigger): TriggerDto {
+    switch (trigger.type) {
+      case TriggerType.ACTION:
+        return new ActionTriggerDto(trigger)
+      case TriggerType.MACRO:
+        return new MacroTriggerDto(trigger)
+      default:
+        throw new UnexpectedCaseException('','Unexpected Trigger type.')
```
LindvedKrvang: Would be nice to include what the unexpected TriggerType is i.e. parse the `trigger.type` variable to the exception.

src/presentation/facades/event-builder-facade.ts
```diff
@@ -18,7 +18,7 @@ export class EventBuilderFacade {
     return new EventBuilder()
   }
 
-  public static createActionTriggerEventBuilder(): ActionTriggerEventBuilder {
+  public static createActionTriggerEventBuilder(): TriggerEventBuilder {
```
LindvedKrvang: Name needs to be updated to not include `Action`

src/presentation/facades/event-emitter-facade.ts
```diff
@@ -34,8 +34,8 @@ export class EventEmitterFacade {
     return RundownEventService.getInstance(EventBuilderFacade.createRundownEventBuilder())
   }
 
-  public static createActionTriggerEventEmitter(): ActionTriggerEventEmitter {
-    return ActionTriggerEventService.getInstance(EventBuilderFacade.createActionTriggerEventBuilder())
+  public static createActionTriggerEventEmitter(): TriggerEventEmitter {
```
LindvedKrvang: Name needs to be updated to not include `Action`

src/presentation/facades/event-emitter-facade.ts
```diff
@@ -50,8 +50,8 @@ export class EventEmitterFacade {
     return ActionEventService.getInstance(EventBuilderFacade.createActionEventBuilder())
   }
 
-  public static createActionTriggerEventObserver(): ActionTriggerEventObserver {
-    return ActionTriggerEventService.getInstance(EventBuilderFacade.createActionTriggerEventBuilder())
+  public static createActionTriggerEventObserver(): TriggerEventObserver {
```
LindvedKrvang: Name needs to be updated to not include `Action`

LindvedKrvang (changes requested):
Looks good :)

I unfortunately have to request changes because you are blindly setting `actionArguments` to zero on _all_ ActionTriggers in your migration. `actionArguments` are optional and should only be set, if the ActionTrigger needs ActionArguments.

src/presentation/dtos/trigger-dto.ts
vushu: As we discussed offline, I will just remove this default case.

RasmusAlbrektsen (approved):
LGTM! :-D

LindvedKrvang (approved):
Awesome :D

PR #329, opened 2025-03-19
SOF-2542 Clone PieceMetadata to not have the same reference in the heap

LindvedKrvang:
When inserting a new source into a SplitScreen input we would update the source of the Piece metadata object directly on the original Piece. This meant that we had altered the metadata on the original Piece and the only way to revert this change is to reingest the Piece again, which is not good/often enough.

To fix this, we now clone the Piece metadata object so we have a different object in the heap. This works since we are using a "replace Piece" Action, so we would create a new Piece anyway. Now that piece simply gets its metadata from a object that isn't pointing to the same reference as the metadata of the original Piece.


vushu (approved):
Looks good, I know you are not a fan of helper functions, but for me it seems like this could just be called through a very simple helper function.

LindvedKrvang:
> Looks good, I know you are not a fan of helper functions, but for me it seems like this could just be called through a very simple helper function.

How do you suggest to access this helper function? By making it static?

vushu:
either that or just a free function.

RasmusAlbrektsen (approved):

LindvedKrvang:
> either that or just a free function.

What's a "free function"?

PR #330, opened 2025-03-20
fix: Creates a link in the dockerfile that allows us to still use the…

RasmusAlbrektsen:
To solve the issue of wanting a relative path in our migration config and still having it work in our docker container, we no link the folder at root level using the Dockerfile, so that migrations will work for a docker container whether run in production or locally in development mode both as a container and on a developer pc setup.


vushu (approved):
Lovely

LindvedKrvang (approved):

PR #336, opened 2025-04-09
Sof 2556/duplicate timeline object ids

LindvedKrvang:
When executing a TransitionEffectAction we would insert timelineObjects into the current Piece to "disable" any transition effects that might be on it.
These TimelineObjects never got cleared again and would just accummulate if the same Piece continuasly got transition effects.
Since the TimelineObjectFactories for Atem and TriCaster only made the ids "unique" to a randomness between 0 and 100, there was a high chance for two TimelineObjects to receive the same id. This is a no-go when resolving the Timeline which would make TSR break.

This PR introduces two things:
- Atem- and TriCasterTimelineObjectFactory now generate a "unique" id by post-fixing the current timestamp.
- Piece have had its TimelineObjects split into two arrays. One for the "original" timelineObjects created when the Piece was created. And one for all "non-planned" TimelineObjects i.e. the TimelineObjects that has been inserted throughout the execution of the Rundown. The "inserted" TimelineObjects are being cleared when we copy the Piece, so we no longer have a lot of TimelineObjects accumulating on Piece during multiple copies. A "gaurd-clause" has also been added to Piece when inserting TimelineObjects. If a TimelineObject with an indentical id already exists, we now throw an error.


src/model/entities/test/piece.spec.ts
```diff
@@ -278,4 +279,101 @@ describe(Piece.name, () => {
...
   })
+
+  describe(Piece.prototype.copy.name, () => {
+    it('gets "_COPY" post-fixed to its id', () => {
+      const copyPostFix: string = '_COPY'
+      const testee: Piece = EntityTestFactory.createPiece({ id: 'somePieceId' })
+      expect(testee.id).not.toContain(copyPostFix)
+
+      const result: Piece = testee.copy()
+      expect(result.id).toContain(copyPostFix)
```
vushu, line 290: this would still pass if you have an id:  '_COPYSomeID' which in this case is a prefix.

src/blueprints/tv2/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
```diff
@@ -151,7 +151,7 @@ export class Tv2AtemVideoMixerTimelineObjectFactory implements Tv2VideoMixerTime
 
   private createAtemMeTimelineObjectForLayer(id: string, layer: Tv2AtemLayer, enable: TimelineEnable, me: AtemMixEffectWithTransition | AtemMixEffectWithPreview, metadata?: TimelineObjectMetadata): AtemMixEffectTimelineObject {
     return {
-      id: `${id}_${Math.floor(Math.random() * 100)}`,
+      id: `${id}_${Date.now()}`,
```
vushu, line 154: Is there a possibility of the `id` argument beeing the same?
Consider `Date.now() + Math.random()`, since there may be a possibility of having same timestamp.

RasmusAlbrektsen (approved):

vushu (approved):

PR #337, opened 2025-04-09
Merge staging into master

LindvedKrvang:
When we released `1.1.0` we forgot to merge staging into master.
This PR rectifies that.
Note: There is some new stuff on staging that is not in the relased version, but it's some bugfixes we need to patch anyway.
We can merge it to master. Make a build and deploy on staging, do a thourough test and then deploy to production


RasmusAlbrektsen (approved):
LGTM

PR #339, opened 2025-04-10
exposing reset, activate, deactivate rundown

vushu:


LindvedKrvang:
What is the motive behind exposing `Activate` and `Deactivate`? I strongly suggest not to expose these two.

For Activate: How do we know which RundownId to send along the Activate action? So far, we fetch the RundownId from the Active Rundown, but the nature of `Activate` is that the Rundown we want to do something on isn't activated, so we don't know how to get the RundownId.

For Deactivate: This leaves the risk of pressing a wrong button and then "out-of-nowhere" you Rundown is now deactivated and the output the viewer is seing is now black... If you call the Action from anywhere that isn't through a dedicated flow in the WebClient, then you are not prompted with a confirmation dialog. (Same thing applies for the Reset Action, and since most of our Rundowns start with a klar-on-air, the viewer will now either see black or the still frame of the jingle)

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

PR #342, opened 2025-05-02
Sof 2311/fix spanningness issue

vushu:
This fix is to detect pieces that has changing life span.

These changes will affect the whole segment, so we will detect in the level of segments,
And will then notify the client that there is a segment update.

- We moved the lifespan check from the part level detection, to be instead handle in the segment level, thereby notifying a segment update

- We also added if a pieces that spans across the segment is removed we notify a segment update.


src/business-logic/services/entity-change-detector.ts
```diff
@@ -33,6 +35,21 @@ export class EntityChangeDetector {
       || segment.expectedDurationInMs !== ingestSegment.budgetDuration
       || segment.invalidity?.reason !== ingestSegment.invalidity?.reason
       || segment.definesShowStyleVariant !== ingestSegment.definesShowStyleVariant
+      || this.hasLifeSpanPieceChanges(segment, ingestSegment)
+  }
+
+  public hasLifeSpanPieceChanges(segment: Segment, ingestedSegment: IngestedSegment): boolean {
```
LindvedKrvang: Should be a private access modifier.

src/business-logic/services/entity-change-detector.ts
```diff
@@ -33,6 +35,21 @@ export class EntityChangeDetector {
...
+    const pieces: Piece[] = segment.getParts().flatMap(part => part.getPieces())
+    const ingestedPieces: IngestedPiece[] = ingestedSegment.ingestedParts.flatMap(part => part.ingestedPieces)
+    const differingPieces: Piece[] = pieces.filter(piece => !ingestedPieces.some(ingestedPiece => ingestedPiece.id === piece.id))
+
+    const hasNewSpanningPieces: boolean = differingPieces.some((piece) => piece.pieceLifespan !== PieceLifespan.WITHIN_PART)
```
LindvedKrvang: Just a note: 
The Lifespan of a Piece can be WithinPart, Spanning or Sticky.
Here you say you are checking if they are new _spanning_ Pieces, but in reality you check for if they aren't WithinPart. This creates a mismatch between name and implementation.

The term we use for Pieces that isn't WithinPart is InifinitePieces. (These Lifespans aren't great, but we have inherited them from Sofie...)

src/business-logic/services/entity-change-detector.ts
```diff
@@ -71,4 +87,8 @@ export class EntityChangeDetector {
       || this.serializeComplexTypeForComparison(ingestedPieceA.metadata) !== this.serializeComplexTypeForComparison(ingestPieceB.metadata)
       || this.serializeComplexTypeForComparison(ingestedPieceA.content) !== this.serializeComplexTypeForComparison(ingestPieceB.content)
   }
+
+  public doesIngestedPiecesDifferInLifeSpan(ingestedPieceA: Piece, ingestPieceB: IngestedPiece): boolean {
```
LindvedKrvang: Should be a private access modifier.

Sidenote:
In an effort to reduce the amount of scrolling up and down when reading a file, we try to keep the methods in the "order they are called". I.e. if a method is calling another method, that other method would be declared beneath the first method (where possible, offcourse).

They flow looks something like this:
```
methodA(): void {
  methodB()
}

methodB(): void {
  methodD()
}

methodD(): void {
  // ...
}

methodC(): void {
  //...
}
```

src/business-logic/services/entity-change-detector.ts
```diff
@@ -61,7 +78,6 @@ export class EntityChangeDetector {
   public doesIngestedPiecesDiffer(ingestedPieceA: IngestedPiece, ingestPieceB: IngestedPiece): boolean {
     return ingestedPieceA.name !== ingestPieceB.name
       || ingestedPieceA.layer !== ingestPieceB.layer
-      || ingestedPieceA.pieceLifespan !== ingestPieceB.pieceLifespan
```
LindvedKrvang: I think we should still check for Lifespan changes on a Part level.
What happens if we get a Lifespan change on a Part, but we don't receive any update to check if a Segment has changed?

LindvedKrvang (approved):

src/business-logic/services/entity-change-detector.ts
RasmusAlbrektsen: I agree to keep this here, so i've added it back. But i think we would always see a segment update, that is what we make sure of in this file. In any case, it has been added back :-)

src/business-logic/services/entity-change-detector.ts
RasmusAlbrektsen: I have changed the modifier as well as the order.

src/business-logic/services/entity-change-detector.ts
RasmusAlbrektsen: I have changed it to specifically look for spanning pieces instead :-)

src/business-logic/services/entity-change-detector.ts
RasmusAlbrektsen: Is now private!

src/business-logic/services/entity-change-detector.ts
```diff
@@ -1,12 +1,12 @@
...
-import { PieceLifespan } from '../../model/enums/piece-lifespan'
+import {Segment} from '../../model/entities/segment'
+import {IngestedSegment} from '../../model/entities/ingested-segment'
+import {Part} from '../../model/entities/part'
+import {IngestedPart} from '../../model/entities/ingested-part'
+import {IngestedPiece} from '../../model/entities/ingested-piece'
+import {IngestedRundown} from '../../model/entities/ingested-rundown'
+import {Rundown} from '../../model/entities/rundown'
+import {Piece} from '../../model/entities/piece'
+import {PieceLifespan} from '../../model/enums/piece-lifespan'
```
LindvedKrvang: @RasmusAlbrektsen You should take a look at your editor settings. We would like to keep these spaces ;)

LindvedKrvang (approved):

src/business-logic/services/entity-change-detector.ts
RasmusAlbrektsen: I guess i my editor relies on the linting from the project usually, because i only have this issue in this specific repo :-D

RasmusAlbrektsen (approved):

LindvedKrvang (approved):

PR #344, opened 2025-05-05
Sof 2571/keep track of and emit playout content

LindvedKrvang:
Note: Dependent on https://github.com/tv2/alba-server/pull/343

This PR sets up the scaffolding for emitting `PlayoutContentEvents`.

It also introduces the `PlayoutContentService` interface with a `PlayoutContentStateService` implementation that keeps track of `PlayoutContent` in both Program and Preview.
If there is any change in the Program or Preview PlayoutContents, then it will update and emit the changes.
If there are no changes, then no event will be emitted.


src/business-logic/services/playout-content-state-service.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (!this.instance) {
+      this.instance = new PlayoutContentStateService(playoutContentEventEmitter)
+    }
+    return this.instance
+  }
+
+  private programPlayoutContents: PlayoutContent[] = []
+  private previewPlayoutContents: PlayoutContent[] = []
+
+  constructor(private readonly playoutContentEventEmitter: PlayoutContentEventEmitter) {
```
LindvedKrvang, line 21: This should be a private constructor, but then I can't instantiate the class for the tests...
Inputs/suggestions are highly appreciated.

src/business-logic/services/playout-content-state-service.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  private areArraysEqual(arrayOne: unknown[], arrayTwo: unknown[]): boolean {
+    if (arrayOne.length !== arrayTwo.length) {
+      return false
+    }
+    const isAllEntriesOfArrayOneIsInArrayTwo: boolean = arrayOne.every(entry => arrayTwo.includes(entry))
+    const isAllEntriesOfArrayTwoIsInArrayOne: boolean = arrayTwo.every(entry => arrayOne.includes(entry))
+    return isAllEntriesOfArrayOneIsInArrayTwo && isAllEntriesOfArrayTwoIsInArrayOne
+  }
```
LindvedKrvang, line 73: This is a very basic implementation of an array equals method.
Inputs/suggestions for optimizations are highly appreciated.

src/business-logic/facades/service-facade.ts
```diff
@@ -190,4 +193,10 @@ export class ServiceFacade {
       EventEmitterFacade.createDeviceEventEmitter()
     )
   }
+
+  public static createPlayoutContentService(): PlayoutContentService {
+    return new PlayoutContentStateService(
```
LindvedKrvang, line 198: This should be `PlayoutContentStateService.getInstance()`.
This has been fixed in https://github.com/tv2/alba-server/pull/345

src/business-logic/services/playout-content-state-service.ts
vushu: I think it's okay to make this public, I don't see any another way to make this easy testable either.

vushu (approved):

src/business-logic/services/playout-content-state-service.ts
LindvedKrvang: But since it is a singleton, it shouldn't be public.
Letting it stay public allow people to do `new PlayoutContentStateService()` resulting in having more than one instance working on different state.

It didn't take longer than this very PR for someone (me) to make that mistake

vushu: ah I see 😄

src/business-logic/services/playout-content-state-service.ts
RasmusAlbrektsen: Do we care if the sequence of  elements is the same or only if the elements are the same?

RasmusAlbrektsen (approved):
Just a question in terms of the array equals implementation :-)

src/business-logic/services/playout-content-state-service.ts
LindvedKrvang: My guees would be we don't care about the sequence.
We basically just want to know id Camera 1 is on or not. Not whether it is the first or fith PlayoutContent.

PR #346, opened 2025-05-07
SOF-2579 Add PlayoutContent to Action metadata

LindvedKrvang:
Note: Dependent on https://github.com/tv2/alba-web-client/pull/374

Add PlayoutContent to Action metadata

Update: Action.metadata now also have a `OuputChannel` so the Action knows if it wants to be in PROGRAM, PREVIEW, or if it doesn't care about it.


RasmusAlbrektsen (approved):

vushu (approved):
We don't have to do any migrations right since this is in metadata?

LindvedKrvang:
> We don't have to do any migrations right since this is in metadata?

No, the Actions are continously generated when ingesting

PR #347, opened 2025-05-14
Sof 2395/fix reset for piece duration

vushu:
Cherry picked the commits from 
https://github.com/tv2/alba-server/compare/staging...SOF-2353/add-taken-off-air-timestamp-to-piece
Since this is what @KvelaGorrrrnio had done in this branch.

which fixed the issue.

**BEFORE** it doesn't fill out the 10 secs in the ui

https://github.com/user-attachments/assets/20e7054c-2ca5-4b51-bff2-405b1a3d99d3

**AFTER** it represents the piece with it's duraction of 10 secs

https://github.com/user-attachments/assets/93eff8aa-2ca1-40d5-8454-c00f4f7df3e7

**If I understand this correctly:**

The fix is to store a `takenOffAirTimestamp` to register when it was taken off.
This way we can use this timestamp to calculate the correct duration of the piece the we `getDuration`.


vushu:
I have removed a redundant line, fixed merge conflicts and re-tested.

src/model/entities/piece.ts
```diff
@@ -77,37 +81,48 @@ export class Piece {
...
+      this.resetExecution()
     }
     this.originalTimelineObjects = [...ingestedPiece.timelineObjects]
   }
 
-  public setExecutedAt(executedAt: number): void {
-    this.executedAt = executedAt
+  public putOnAir(putOnAirTimestamp: number): void {
+    if (this.executedAt) {
+      return
```
LindvedKrvang, line 98: I think we should log if we try to put something onAir that is supposedly already onAir.
I can't remember it breaks anything if we don't set it, but we should at least also log this case instead of only returning

src/model/entities/piece.ts
```diff
@@ -77,37 +81,48 @@ export class Piece {
...
       // Infinite Pieces might still be OnAir when their Part is reset, so we can't reset their "executedAt" here.
-      this.executedAt = 0
+      this.resetExecution()
     }
     this.originalTimelineObjects = [...ingestedPiece.timelineObjects]
   }
 
-  public setExecutedAt(executedAt: number): void {
-    this.executedAt = executedAt
+  public putOnAir(putOnAirTimestamp: number): void {
```
LindvedKrvang, line 96: We are trying get away from the "onAir" semantic in the backend.

Besides this method sets the timestamp where the Piece is supposedly being executed. This isn't necessarily "now". `putOnAir` indicates that it happens right now and if that's the case then we wouldn't need the timestamp as a parameter.

The execution of the Piece depends on the pre- and post-roll of both the Piece and its corresponding Part, which could be either now, in the past or in the future.

LindvedKrvang (approved):
Looks good.

Only thing is that I'm personally not a fan of the semantic change going from `setExecutedAt` to `putOnAir`. I have tried explaining it a bit further in one of the comments.

RasmusAlbrektsen (approved):

vushu:
Fair observations. I think, I will merge it in now as is, I think this needs to be discussed further.

src/model/entities/piece.ts
RasmusAlbrektsen: I think it was intended to be `put` in past tense, so as to say: "this is the timestamp where we put this piece on air".

PR #349, opened 2025-05-19
Sof 2581/misc lights in skaarhoj

LindvedKrvang:
Introduce `baselinePieces`. These are Pieces that will be inserted into the infinite Pieces if there are no other Pieces that takes priority on the layer.
The principle is the same as baseline TimelineObjects. We need them as Pieces because we need to acces the PlayoutContent of the baseline TimelineObjects.

Also in this PR:
- Refactor to introduce `TimelineObjectFactoryProvider`. `ActionFactoryProvider` no longer also instantiate the TimlineObjectFactories. This is so we can use TimelineObjectFactories outside of the ActionFactoryProvider.
- Introduce MutateActionMethods for mutating an Action with PlayoutContent. This will take a predicate that found a PlayoutContent by first looking in the ProgramPlayoutContents, then PreviewPlayoutContents and then allow the Action to be mutated by it.
- Introduce Action to toggle DownstreamKeyers.
- Persist PlayoutContentState and read it on startup


vushu (approved):
LGTM

src/blueprints/tv2/action-factories/tv2-video-mixer-configuration-action-factory.ts
```diff
@@ -22,10 +30,35 @@ export class Tv2VideoMixerConfigurationActionFactory extends ActionFactory {
   public createVideoMixerActions(blueprintConfiguration: Tv2BlueprintConfiguration): Tv2Action[] {
     return [
       ...this.createDownstreamKeyerOffActions(blueprintConfiguration),
-      ...this.createDownstreamKeyerOnActions(blueprintConfiguration)
+      ...this.createDownstreamKeyerOnActions(blueprintConfiguration),
+      ...this.createEmptyDownstreamKeyerToggleActions(blueprintConfiguration)
     ]
   }
 
+  public isVideoMixerAction(action: Tv2Action): boolean {
```
RasmusAlbrektsen, line 38: Is this perhaps not a name fitting of the functionality? From the method signature it seems to assert if we have a video mixer related action but we only check whether the `PlayoutContentType` is of type `DOWNSTREAM_KEYER`.

RasmusAlbrektsen (approved):
Nice changes!

src/blueprints/tv2/action-factories/tv2-video-mixer-configuration-action-factory.ts
LindvedKrvang: It follows the "contract" (which isn't really defined as a contract) that the XActionFactory has a "isXAction" to be used when getting mutated Actions.
Right now this is only the DOWNSTREAM_KEYER that needs that.

It could be argued (as you point out) that the current implementation is focussed more on how it is used instead of what it is supposed to do :)

LindvedKrvang: On the other hand: We have no other videoMixerActions so we don't have any other things to put here

PR #351, opened 2025-05-21
Added logic to handle partless segments when they are on air

vushu:
Added behavior to handle partless segments, which will now be marked as unsynced segments if they are on air.
as described in [SOF-2172](https://tv2cms.atlassian.net/browse/SOF-2172)

https://github.com/user-attachments/assets/bcd8a8da-cddf-4b53-afe3-36e0774690a3

[SOF-2172]: https://tv2cms.atlassian.net/browse/SOF-2172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

In result by this change we will enforce the behavior of unsynced segment meaning this will happen:

https://github.com/user-attachments/assets/8036a5e4-6506-446f-91bd-9787c16b91bd

So next time you reinsert the parts, it will be seen as a new segment


vushu:
There are some issues that I need to iron out, closing this for now.

vushu:
Reopened after talking about some behavioral changes with the team, which we agreed upon to be expected behavior.

src/business-logic/services/ingest-rundown-synchronizer.ts
```diff
@@ -30,11 +30,12 @@ export class IngestRundownSynchronizer {
   ) {}
 
   public async synchronizeRundown(rundown: Rundown, ingestedRundown: IngestedRundown): Promise<RundownSynchronizeResult> {
+    const partLessIngestedSegmentIds: ReadonlySet<string> = this.getPartlessSegmentIds(ingestedRundown.ingestedSegments)
```
RasmusAlbrektsen: Should it perhaps be `partlessIngestedSegmentIds`? Just like you have `getPartlessSegmentIds` :-)

RasmusAlbrektsen: This recurs a few times in the PR, so remember to change it everywhere :-)

RasmusAlbrektsen (approved):

src/business-logic/services/ingest-rundown-synchronizer.ts
vushu: Ah you are right :d

LindvedKrvang (approved):

PR #352, opened 2025-05-26
SOF-1964 Able to connect to devices.

LindvedKrvang:
We are now able to connect to Devices.
If a connection is lost, it will try to reconnect until a new connection is established.

This PR contains a refactor of the interface previously known as `Device`. It is now called `DeviceConfiguration`.
A new class `Device` has been introduced. This class is responsible for maintaining the connection to the physical (or virtual) device.

`Devices` are the active connections and are therefore not saved in the database. That's what the `DeviceConfiguration` is for.

When connecting to a device we instantiate a `Device` from a `DeviceConfiguration`.
When deleting a `DeviceConfiguration`, we disconnect its corresponding `Device` and then discard the `Device`.
When updating a `DeviceConfiguration` we disconnect its corresponding `Device`, discard the `Device`, and creates a new `Device` with the updated `DeviceConfiguration`.
The reason for "discarding" the `Device` and creating a new one is to keep it immutable.

The interface `CoreDeviceConfiguration` has been introduced.
This interface is so we don't have to pollute our `DeviceConfiguration` with the information required to make the current Device handling system from Core working.

The REST endpoints for "device handling" has been modified.
The POST, PUT, and DELETE endpoints are in context of `DeviceConfigurations`. (They are unchanged).
The GET endpoints now returns `DeviceDtos`.

Note: A lot of files have had a name change. That's why there are so many deleted/new files.


src/model/entities/device-configuration.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// "Core" devices are in their own flow, so they should not be included in the DeviceConfiguration type
+export interface CoreDeviceConfiguration {
+  id: string
+  name: string
+  type: DeviceType
+  isConnected: boolean
+  statusCode: StatusCode
+  statusMessage: string
+}
```
KvelaGorrrrnio, line 31: Could it be an idea to move this to a separate file to make it clear if we accidentally add dependencies between Core and non-Core types?

src/model/entities/devices/device.ts
```diff
@@ -0,0 +1,117 @@
+import { DeviceType } from '../../enums/device-type'
+import { StatusCode } from '../../enums/status-code'
+import { DeviceConfiguration } from '../device-configuration'
+import { StatusMessageEventEmitter } from '../../../business-logic/services/interfaces/status-message-event-emitter'
+import { StatusMessage } from '../status-message'
+
+const RECONNECT_TIMEOUT_IN_MS: number = 10 * 1000
```
KvelaGorrrrnio, line 7: `10 * 1000` is `10_000` ;)

src/model/entities/devices/device.ts
```diff
@@ -0,0 +1,117 @@
...
+      this.updateStatus(StatusCode.WARNING, 'Disconnected')
+    }
+    this.isConnected = false
+    this.reconnect()
+  }
+
+  private reconnect(): void {
+    if (this.timeoutIdentifier) {
+      clearTimeout(this.timeoutIdentifier)
+    }
```
KvelaGorrrrnio, line 85: The if-statement is unnecssary since `clearTimeout` is idempotent :)
Same goes for the `disconnect` method.

src/model/entities/devices/device.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.updateStatus(StatusCode.WARNING, 'Connecting...')
+    this.connectToDevice(
+      () => this.onDeviceConnectedCallback(),
+      (errorMessage: string) => this.onDeviceErrorCallback(errorMessage),
+      () => this.onDeviceDisconnectedCallback()
+    )
+  }
+
+  private updateStatus(statusCode: StatusCode, statusMessage: string): void {
```
KvelaGorrrrnio, line 49: Should calling this multiple times with the same inputs update and emit on every call? Or should it only change if the state changes?

src/presentation/dtos/device-dto.ts
```diff
@@ -1,56 +1,15 @@
...
   constructor(device: Device) {
-    this.type = device.type
-    this.id = device.id
-    this.name = device.name
-    this.isConnected = device.isConnected
-    this.statusCode = device.statusCode
-    this.statusMessage = device.statusMessage
+    this.statusCode = device.getStatusCode()
+    this.statusMessage = device.getStatusMessage()
+    this.configuration = device.getConfiguration()
```
KvelaGorrrrnio, line 13: We should be aware going forward that this is likely to expose secrets like passwords, API keys or other secrets.

src/business-logic/services/device-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+    const device: Device = this.createDevice(createdDeviceConfiguration)
+    this.deviceEventEmitter.emitDeviceCreatedEvent(device)
+    device.connect()
+  }
+
+  public async update(deviceConfiguration: DeviceConfiguration): Promise<void> {
+    await this.deviceConfigurationRepository.update(deviceConfiguration)
+    this.disconnectedAndRemoveDevice(deviceConfiguration.id)
+    const device: Device = this.createDevice(deviceConfiguration)
+    this.deviceEventEmitter.emitDeviceUpdatedEvent(device)
```
KvelaGorrrrnio, line 72: Why are we emitting an event here? Won't `this.createDevice` produce an event as well?

src/model/entities/devices/telemetrics-device.ts
```diff
@@ -0,0 +1,33 @@
+import { StatusCode } from '../../enums/status-code'
+import { Device } from './device'
+import { TelemetricsDeviceConfiguration } from '../device-configuration'
+import { StatusMessageEventEmitter } from '../../../business-logic/services/interfaces/status-message-event-emitter'
+
+// TODO: This is not a final implementation. This is a "placeholder" until we make the real implementation.
+export class TelemetricsDevice extends Device {
```
KvelaGorrrrnio, line 7: As far as I can see, there is no reason to have a `TelemetricsDevice` class in the Alba server. We are only going to connect to playout devices from the playout gateway :)
Has the team talked about how we will differentiate between "gateways" and playout devices?

src/model/entities/devices/i-news-device.ts
```diff
@@ -0,0 +1,77 @@
+import { StatusCode } from '../../enums/status-code'
+import WebSocket, { CloseEvent, ErrorEvent, Event, MessageEvent } from 'ws'
+import { Device } from './device'
+import { INewsDeviceConfiguration } from '../device-configuration'
+import { StatusMessageEventEmitter } from '../../../business-logic/services/interfaces/status-message-event-emitter'
+
+const INEWS_HOST: string = process.env.INEWS_HOST ?? 'ws://localhost:3008'
+
+export class INewsDevice extends Device {
```
KvelaGorrrrnio, line 9: Shouldn't this be `InewsDevice` as per the styleguide?
Same for `INewsDeviceConfiguration`.

src/model/entities/devices/i-news-device.ts
```diff
@@ -0,0 +1,77 @@
+import { StatusCode } from '../../enums/status-code'
+import WebSocket, { CloseEvent, ErrorEvent, Event, MessageEvent } from 'ws'
+import { Device } from './device'
+import { INewsDeviceConfiguration } from '../device-configuration'
+import { StatusMessageEventEmitter } from '../../../business-logic/services/interfaces/status-message-event-emitter'
+
+const INEWS_HOST: string = process.env.INEWS_HOST ?? 'ws://localhost:3008'
```
KvelaGorrrrnio, line 7: It seems a bit odd that we have an iNews device configuration, but not all configurations for it are defined in it.
It would be interesting to explore whether it should be a device, and how the configuration should be managed.

src/model/entities/devices/i-news-device.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export class INewsDevice extends Device {
+
+  private socket: WebSocket
+
+  constructor(
+    private readonly iNewsDeviceConfiguration: INewsDeviceConfiguration,
+    statuesMessageEventEmitter: StatusMessageEventEmitter,
```
KvelaGorrrrnio, line 15: `statues` should be `status` :)

src/model/entities/devices/i-news-device.ts
```diff
@@ -0,0 +1,77 @@
...
+        return StatusCode.UNKNOWN
+      }
+    }
+  }
+
+  protected connectToDevice(
+    onConnectedCallback: () => void,
+    onErrorCallback: (errorMessage: string) => void,
+    onDisconnectedCallback: () => void
+  ): void {
```
KvelaGorrrrnio, line 45: Just a thought: Could it make sense to return an observable instead?
There might be some complications like timing depending on how the observable is implemented.

src/data-access/repositories/mongo/mongo-device-changed-listener.ts
```diff
@@ -1,19 +1,19 @@
...
 import { ChangeStream, ChangeStreamDeleteDocument, ChangeStreamDocument, ChangeStreamOptions } from 'mongodb'
 import { MongoChangeEvent } from './mongo-enums'
 import { Logger } from '../../../logger/logger'
-import { Device } from '../../../model/entities/device'
+import { CoreDeviceConfiguration } from '../../../model/entities/device-configuration'
 
 const DEVICE_COLLECTION_NAME: string = 'peripheralDevices'
 
-export class MongoDeviceChangedListener extends BaseMongoRepository<MongoDevice> implements DataChangedListener<Device> {
+export class MongoDeviceChangedListener extends BaseMongoRepository<MongoDeviceConfiguration> implements DataChangedListener<CoreDeviceConfiguration> {
```
KvelaGorrrrnio, line 12: Suggestion: Changing the class name to `MongoCoreDeviceConfigurationChangedListener` to reflect that this only listens for changes to the Sofie Core collection.

src/data-access/facades/repository-facade.ts
```diff
@@ -287,24 +287,24 @@ export class RepositoryFacade {
     return new MongoSystemInformationRepository(MongoDatabase.getInstance(LoggerFacade.createLogger()), new MongoEntityConverter(LoggerFacade.createLogger()))
   }
 
-  public static createDeviceDataChangedListener(): DataChangedListener<Device> {
+  public static createDeviceConfigurationDataChangedListener(): DataChangedListener<CoreDeviceConfiguration> {
```
KvelaGorrrrnio, line 290: Suggestion: By adding `Core` in between `create` and `DeviceConfiguration` we emphasize that this is only for listening for DeviceConfigurations of Sofie Core.

src/data-access/repositories/mongo/mongo-core-device-configuration-repository.ts
```diff
@@ -0,0 +1,41 @@
+import { BaseMongoRepository } from './base-mongo-repository'
+import { MongoDatabase } from './mongo-database'
+import { MongoDeviceConfiguration, MongoEntityConverter, MongoId } from './mongo-entity-converter'
+import { DeviceConfigurationRepository } from '../interfaces/device-configuration-repository'
+import { DeviceConfiguration } from '../../../model/entities/device-configuration'
+import { UnsupportedOperationException } from '../../../model/exceptions/unsupported-operation-exception'
+
+const DEVICE_COLLECTION_NAME: string = 'peripheralDevices'
+
+export class MongoCoreDeviceConfigurationRepository extends BaseMongoRepository<MongoId> implements DeviceConfigurationRepository {
```
KvelaGorrrrnio, line 10: This class does not implement the interface properly. Either the contract should be changed for the class or the `getDeviceConfigurations` method should map the data to `DeviceConfiguration`s :)

src/business-logic/services/device-changed-service.ts
```diff
@@ -53,14 +53,14 @@ export class DeviceChangedService implements DataChangeService {
   }
 
   private async updateStatusMessageFromCurrentDeviceStatus(): Promise<void> {
-    const devices: Device[] = await this.deviceRepository.getDevices()
+    const devices: CoreDeviceConfiguration[] = await this.deviceRepository.getDeviceConfigurations() as unknown as CoreDeviceConfiguration[]
```
KvelaGorrrrnio, line 56: Perhaps a `coreDeviceRepository` should be used instead to avoid making assumptions on the data structure?

KvelaGorrrrnio (commented):
I am not quite sure what the "device" term should encompass - it seems a bit vague/broad. I am 26/33 files deep, but I will continue tomorrow 😁

src/business-logic/services/device-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  private createDevice(deviceConfiguration: DeviceConfiguration): Device {
+    const device: Device = this.deviceFactory.createDevice(deviceConfiguration, (device: Device) => this.deviceEventEmitter.emitDeviceUpdatedEvent(device))
+    this.devices.push(device)
+    return device
+  }
+
+  public getDevices(): Device[] {
+    return this.devices
+  }
```
KvelaGorrrrnio, line 51: A mutable reference is returned here. To make it a bit more "safe" we could spread the list or mark the return type as `readonly Device[]`.

src/business-logic/services/device-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  private disconnectedAndRemoveDevice(deviceConfigurationId: string): void {
+    const deviceIndex: number = this.devices.findIndex(device => device.getId() === deviceConfigurationId)
+    if (deviceIndex === -1) {
+      return
+    }
+    // This should always return an array with exactly one entry.
+    const deletedDevices: Device[] = this.devices.splice(deviceIndex, 1)
+    deletedDevices[0].disconnect()
```
KvelaGorrrrnio, line 83: Suggestion: By moving the indexing up after the splice, the syntax might be a bit closer to what the semantics entail. By doing this the reason for the comment to exist is now confined to a single line instead of two and the comment can be moved to that specific line to avoid comment drifting. Also we should probably use the same terminology for the action performed (remove vs delete). E.g.
```typescript
const removedDevice: Device = this.devices.splice(deviceIndex, 1)[0]! // This should always return an array with exactly one entry.
removedDevice.disconnect()
```

src/business-logic/services/device-manager.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  public async update(deviceConfiguration: DeviceConfiguration): Promise<void> {
+    await this.deviceConfigurationRepository.update(deviceConfiguration)
+    this.disconnectedAndRemoveDevice(deviceConfiguration.id)
+    const device: Device = this.createDevice(deviceConfiguration)
+    this.deviceEventEmitter.emitDeviceUpdatedEvent(device)
+    device.connect()
+  }
+
+  private disconnectedAndRemoveDevice(deviceConfigurationId: string): void {
```
KvelaGorrrrnio, line 76: Since this is an action to perform, we could change `disconnected` to `disconnect`, such that the method name is `disconnectAndRemoveDevice`.

src/business-logic/services/test/device-changed-service.spec.ts
```diff
@@ -1,19 +1,19 @@
...
+import { DeviceConfigurationRepository } from '../../../data-access/repositories/interfaces/device-configuration-repository'
+import { EntityTestFactory } from '../../../model/entities/test/entity-test-factory'
 
 const DEVICE_STATUS_MESSAGE_PREFIX: string = 'DEVICE_'
 
 describe(DeviceChangedService.name, () => {
   it('calls the StatusMessageService with a StatusMessage that has the same id as the Device', () => {
-    const device: Device = EntityTestFactory.createDevice({id: 'deviceId'})
-    const deviceDataChangedListener: DataChangedListener<Device> = createDeviceChangedListenerMock(device)
+    const device: CoreDeviceConfiguration = EntityTestFactory.createCoreDeviceConfiguration({id: 'deviceId'})
```
KvelaGorrrrnio, line 15: Just a note: Since `device` has been given a new meaning, that should probably be reflected in variable names :)

src/business-logic/services/test/device-manager.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+  )
+}
+
+function createDeviceFactory(deviceConfigurations: DeviceConfiguration[]): DeviceFactory {
+  const deviceFactory: DeviceFactory = mock(DeviceFactory)
+  deviceConfigurations.forEach(deviceConfiguration => {
+    when(deviceFactory.createDevice(deviceConfiguration, anything())).thenReturn(instance(mock(Device)))
+  })
+  return deviceFactory
+}
```
KvelaGorrrrnio, line 233: Why is deviceConfigurations passed in? Wouldn't returning an instance of a mocked device for any given configuration be the same?
Is there a use-case where we don't want to have a device when given a device configuration?

src/data-access/repositories/mongo/mongo-device-configuration-repository.ts
```diff
@@ -0,0 +1,64 @@
...
+    super(mongoDatabase)
+  }
+
+  protected getCollectionName(): string {
+    return DEVICE_COLLECTION_NAME
+  }
+
+  public async getDeviceConfigurations(): Promise<DeviceConfiguration[]> {
+    this.assertDatabaseConnection(MongoDeviceConfigurationRepository.prototype.getDeviceConfigurations.name)
+    return await this.getCollection().find<DeviceConfiguration>({}).toArray()
```
KvelaGorrrrnio, line 23: The `await` can be removed :)

src/data-access/repositories/mongo/mongo-device-configuration-repository.ts
```diff
@@ -0,0 +1,64 @@
...
+    return deviceConfiguration
+  }
+
+  public async create(deviceConfiguration: DeviceConfiguration): Promise<DeviceConfiguration> {
+    this.assertDatabaseConnection(MongoDeviceConfigurationRepository.prototype.create.name)
+    const deviceConfigurationWithId: DeviceConfiguration = {
+      ...deviceConfiguration,
+      id: this.uuidGenerator.generateUuid(),
+    }
+    await this.getCollection().updateOne({ id: deviceConfigurationWithId.id }, { $set: deviceConfigurationWithId }, { upsert: true })
```
KvelaGorrrrnio, line 41: Why not use insertOne? I assume that it is to avoid having two documents with the same `id`.
An alternative, if we are afraid that the UUID generator won't create UUIDs, is to use assert that it doesn't exist before inserting.

LindvedKrvang:
Closing the PR since it is no longer valid

PR #354, opened 2025-05-28
SOF-1966 Delete Device setup as it is now redundant

LindvedKrvang:
Delete Device handling since we are no longer going to need this.


vushu (approved):
LGTM

RasmusAlbrektsen (approved):

PR #355, opened 2025-05-30
SOF-2191 Replace references for 'Sofie' with 'Alba'

LindvedKrvang:
Replace references of 'Sofie' with 'Alba'.

Note: The package called `sofie-stack` and its associated content is still named 'Sofie'. This is because it starts Sofie-Core, the iNew-gateway and the playout-gateway i.e. the old Sofie stack.

Disclaimer: I don't know if the scripts for migration is broken. It may be necessary to do a new database dump with the correct name.


RasmusAlbrektsen (approved):

.github/workflows/create-merge-pull-requests.yml
```diff
@@ -13,8 +13,8 @@ jobs:
         - uses: actions/checkout@v3
         - name: Set Git config
           run: |
-            git config --local user.email "Sofie-Automation@users.noreply.github.com"
-            git config --local user.name "GitHub Actions # Sofie-Automation"
+            git config --local user.email "Alba-Automation@users.noreply.github.com"
```
KvelaGorrrrnio, line 16: Shouldn't this be the same across repos? :) I think it is simply `alba@users.noreply.github.com` in alba-tv2-server.

KvelaGorrrrnio (approved):

.github/workflows/create-merge-pull-requests.yml
LindvedKrvang: I guess. 
I am not well versed in how we want our setup to be

PR #360, opened 2025-06-10
SOF-2593: Move files into new structure

KvelaGorrrrnio:
The PR aims to combine a set of self-explanatory commits that move files into their respective context modules and DDD-based submodules.

Noticeable changes:
- `sofie-ingest` module is added to contain current database change listeners and synchronization along with media update subscription.
- `action-system` that contains everything related to actions (macros, actions and triggers).
- `blueprints` that contains the current blueprint implementation.
- Adds TODOs for obvious violations of the architecture and boy scout notes.
- There is a module called `leftovers` where files that needs to be split up or changed is put. This is mostly services like facades and event bus. These will be handled right away in a separate PR.


vushu (approved):
Cool LGTM.

RasmusAlbrektsen (approved):

PR #361, opened 2025-06-11
SOF-2593: Split event type

KvelaGorrrrnio:
_**Note:** Is based on #360._
The PR splits up the `EventType` type and enumerations into multiple files, so they can be moved to the modules that own them.
In doing so `IngestEventType` has been merged into `RundownEventType` except for the media related types, which are extracted into `MediaEventType`.
The PR also moves the action event types from the domain module to the application module as they had been misplaced.

PR #362, opened 2025-06-12
SOF-2593: Split event facades

KvelaGorrrrnio:
_**Note:** This is based on #361._
The EventBuilder, EventBuilderFacade and EventEmitterFacade has been split up into separate classes based which application modules own the corresponding events.


vushu (approved):

RasmusAlbrektsen (approved):
Every new file doesn't have bracket spaces on import, also all new imports in existing files. We should really add this rule, since we've aimed to keep this until now.

KvelaGorrrrnio:
> Every new file doesn't have bracket spaces on import, also all new imports in existing files. We should really add this rule, since we've aimed to keep this until now.

When doing this, it would be great if we make an eslint configuration package that can be shared across our repositories for a streamlined linting setup.

PR #363, opened 2025-06-13
SOF-2593: Use composition root for dependency injection setup

KvelaGorrrrnio:
_**Note:** Is based on #362, but makes the changes to facades obsolete as they are removed._

The PR is a bit more messy than I would have liked - so sorry for that. The main change in this PR is removing the facades used for program composition (setting up the dependency injection) and introduces the object graph composition in the index.ts file as a [composition root](https://freecontent.manning.com/dependency-injection-in-net-2nd-edition-understanding-the-composition-root/).

The main reasoning behind moving to a composition root for object graph construction is that it centralizes the composition, and allows for pairing an interface with an implementation from another module, which the facade-approach doesn't (without getting cyclic dependencies). It is a change in mentality, where one sees modules as libraries or components that are available when composing a program, rather than part of the program structure.

In theory using a composition root can save a bit of memory as for complex graphs we can reuse objects instead of creating once per class depending on it.

A downside of doing the object graph composition in a single file is that it can be quite huge and if not managed properly become unmanageable.

_**Note:** The AuditLog decorator is assigned a separate instance of the ConsoleLogger as it is not a part of the dependency injection structure._

_**Note:** There still files that need to be split up (e.g. MongoEntityConverter), but that will be handled in the task SOF-2592 that implements the fitness functions for ensuring correct dependency usage across modules and submodules._

In addition this PR also:
- The `WebSocketEventServer.startServer` method is made async, to be able to sequence and time the startup process of the application.
- Dissolves the Tv2Logger, and uses the Logger interface from the cross-cutting-concerns module.
- Moves data change services subscription setup calls from the constructor to the `initialize` method, to the construction synchronous and separate construction for initialization. 
- Dissolves singleton pattern usage (`getInstance`), and the single instance requirement is handled by the composition root in index.ts.


src/cross-cutting-concerns/application/decorators/audit-log-decorator.ts
```diff
@@ -9,7 +9,7 @@ interface IncomingRequest {
 }
 
 export function AuditLog(): MethodDecorator {
-  const logger: Logger = LoggerFacade.createLogger()
+  const logger: Logger = new ConsoleLogger()
```
LindvedKrvang, line 12: What is the reason for not injecting a Logger instance here, but instead make its own instantiation?

LindvedKrvang (approved):
Wonderful to see the singletons dying.

There are a lot of spaces missing in the import statements.

src/blueprints/domain/services/action-factories/tv2-action-factory-provider.ts
```diff
@@ -39,10 +38,16 @@ export class Tv2ActionFactoryProvider {
...
+
   constructor(
     private readonly configurationMapper: Tv2ConfigurationMapper,
-    private readonly timelineObjectFactoryProvider: TimelineObjectFactoryProvider
-  ) { }
+    private readonly timelineObjectFactoryProvider: TimelineObjectFactoryProvider,
+    private readonly objectCloner: ObjectCloner,
+    logger: Logger
+  ) {
+    this.logger = logger.tag(this.constructor.name)
```
vushu, line 49: Does this create a new logger instance?

vushu (approved):

src/cross-cutting-concerns/application/decorators/audit-log-decorator.ts
KvelaGorrrrnio: Because this is a static method. The PR description has a note on it:
> Note: The AuditLog decorator is assigned a separate instance of the ConsoleLogger as it is not a part of the dependency injection structure.

src/blueprints/domain/services/action-factories/tv2-action-factory-provider.ts
KvelaGorrrrnio: It wraps the logger instance in a context object (that conforms to the logger interface) that has the new the tag metadata :-)

PR #364, opened 2025-06-16
SOF-2564: Allow multiple Infinite Pieces on the same layer

LindvedKrvang:
This PR allow multiple Infinite Pieces to be on the same layer.
If there are more than one Infinite Piece on the same layer, the "old" Infinite Pieces will be "takenOffAir" when the "current" Infinite Piece is executed plus whatever delay the active Part has.

Note: There should only ever be more than one Infinite Piece on the same layer if we are going to a new Part that has an Infinite Piece for the layer _and_ the Part wants to delay its execution of its Pieces. 
This should be a rare instance, so for the most of the time, we will still only have one Infinite Piece per layer.

Also in this PR:
- Rename the `SuperflyTimelineBuilder` to `SuperTimelineBuilder` as it is based on the "super-timeline", not the "super-fly-timeline".


src/data-access/repositories/mongo/mongo-entity-converter.ts
```diff
@@ -265,8 +265,17 @@ export class MongoEntityConverter {
...
+  private mapToInfinitePieceMap(infinitePieces: Piece[]): Map<string, Piece[]> {
+    const map: Map<string, Piece[]> = new Map()
+    infinitePieces.forEach(piece => {
+      const piecesForLayer: Piece[] | undefined = map.get(piece.layer)
+      if (!piecesForLayer) {
+        map.set(piece.layer, [piece])
+        return
+      }
+      piecesForLayer.push(piece)
+    })
```
KvelaGorrrrnio: Suggestion: Nullish coalescing can be used to make the `pieceForLayer` variable non-nullable, which would remove the need for the if statement:
```typescript
  infinitePieces.forEach( piece => {
      const piecesForLayer: Piece[] = map.get(piece.layer) ?? []
      map.set(piece.layer, [...piecesForLayer, piece])
  })
```

src/model/entities/rundown.ts
```diff
@@ -412,31 +412,25 @@ export class Rundown extends BasicRundown {
...
-    this.resetOutlivedInfinitePieces(Array.from(layersWithPieces.values()))
+    this.resetOldInfinitePieces(layersWithPieces)
     this.setInfinitePieces(layersWithPieces)
   }
 
-  private findOldInfinitePiecesNotOnLayers(layers: Set<string>): Piece[] {
-    return Array.from(this.infinitePieces.values()).filter((oldPiece) => !layers.has(oldPiece.layer))
+  private findOldNotOutlivedInfinitePieces(): Piece[] {
+    return Array.from(this.infinitePieces.values()).flatMap((oldList: Piece[]) => oldList).filter(oldPiece => !this.isPieceOutlived(oldPiece))
   }
```
KvelaGorrrrnio: Suggestion: `flat` can be used instead of `flatMap` as there is no functionality applied to the `oldList`.
Also there is nothing in the method that indicates whether it is a new or old piece - that is currently determined where the method is called.
```typescript
  private findNotOutlivedInfinitePieces(): Piece[] {
    return Array.from(this.infinitePieces.values()).flat().filter(piece => !this.isPieceOutlived(piece))
  }
```

I find the the `NotOutlived` part a bit weirdly phrased, but I don't have a better suggestion :D

src/model/entities/rundown.ts
```diff
@@ -470,26 +464,41 @@ export class Rundown extends BasicRundown {
     }
   }
 
-  private resetOutlivedInfinitePieces(piecesThatHasNotBeenOutlived: Piece[]): void {
-    const pieceIdsThatHasNotBeenOutlived: string[] = piecesThatHasNotBeenOutlived.map((piece) => piece.id)
-    Array.from(this.infinitePieces.values())
-      .filter((piece) => !pieceIdsThatHasNotBeenOutlived.includes(piece.id))
-      .forEach((piece) => piece.resetExecution())
+  private resetOldInfinitePieces(newInfinitePieces: Map<string, Piece[]>): void {
+    this.infinitePieces.forEach((oldList: Piece[], layer: string) => {
```
KvelaGorrrrnio: `oldList` is not very descriptive and could be renamed to something like `infinitePieces` or `oldInfinitePieces`. Depending on how one interprets the term _old_, one could argue that these are not old - they are the current infinite pieces. So perhaps `infinitePieces` or `currentInfinitePieces`? If that is too broad, then perhaps `infinitePiecesOnLayer`?

src/model/entities/rundown.ts
```diff
@@ -470,26 +464,41 @@ export class Rundown extends BasicRundown {
...
+      const newList: Piece[] | undefined = newInfinitePieces.get(layer)
+      if (!newList) {
+        oldList.forEach(piece => piece.resetExecution())
+        return
+      }
+
+      oldList.filter(piece => newList.every(p => p.id !== piece.id))
+        .forEach(piece => piece.resetExecution())
+    })
   }
```
KvelaGorrrrnio: Suggestion: With nullish coalescing, we can get rid of the nullable type for the `newList` variable and remove the if statement:
```typescript
  private resetInfinitePiecesNoLongerOnLayers(newInfinitePieces: Map<string, Piece[]>): void {
    this.infinitePieces.forEach((currentInfinitePiecesOnLayer: Piece[], layer: string) => {
      const newInfinitePiecesOnLayer: Piece[] = newInfinitePieces.get(layer) ?? []

      currentInfinitePiecesOnLayer
        .filter(piece => newInfinitePiecesOnLayer.every(p => p.id !== piece.id))
        .forEach(piece => piece.resetExecution())
    })
  }
```

src/model/entities/rundown.ts
```diff
@@ -470,26 +464,41 @@ export class Rundown extends BasicRundown {
...
+      if (!newList) {
+        layersWithPieces.set(piece.layer, [piece])
+        return
+      }
+
+      if (this.getActivePart().getTimings().delayStartOfPiecesDuration === 0) {
+        return
+      }
+
+      if (piece.id === newList[0].id) {
```
KvelaGorrrrnio: What does `newList[0]` refer to? Is that the oldest element?

KvelaGorrrrnio (commented):
Overall looks good. Disclaimer: I am not through the entire PR yet :-)

src/model/entities/rundown.ts
RasmusAlbrektsen: `Ongoing` ?

RasmusAlbrektsen (approved):

src/model/entities/rundown.ts
```diff
@@ -888,18 +898,18 @@ export class Rundown extends BasicRundown {
...
     infinitePiecesFromPart.map(piece => {
       piece.markAsUnsynced()
       return piece.getUnsyncedCopy()
-    }).forEach(unsyncedPiece => this.infinitePieces.set(unsyncedPiece.layer, unsyncedPiece))
+    }).forEach(unsyncedPiece => this.infinitePieces.set(unsyncedPiece.layer, [unsyncedPiece]))
   }
 
   public getInfinitePieces(): Piece[] {
-    return Array.from(this.infinitePieces.values())
+    return Array.from(this.infinitePieces.values()).flatMap(array => array)
```
KvelaGorrrrnio: `.flat()`. can be used here instead:
```typescript
    return Array.from(this.infinitePieces.values()).flat()
```

src/data-access/repositories/mongo/mongo-entity-converter.ts
LindvedKrvang: Updated

src/model/entities/rundown.ts
LindvedKrvang: Updated.

src/model/entities/rundown.ts
LindvedKrvang: Updated

src/model/entities/rundown.ts
LindvedKrvang: Updated

src/model/entities/rundown.ts
LindvedKrvang: It is supposed to be the "newest" or the InfinitePiece that is supposed t obe OnAir.

I can see that I put the newly added Pieces at the end of the list, which does indicates that `newList([0])` is the oldest.
I have updated so the newly added Pieces is put at the start of the list.

src/business-logic/facades/service-facade.ts
```diff
@@ -2,7 +2,7 @@ import { RundownService } from '../services/interfaces/rundown-service'
 import { RundownTimelineService } from '../services/rundown-timeline-service'
 import { RepositoryFacade } from '../../data-access/facades/repository-facade'
 import { TimelineBuilder } from '../services/interfaces/timeline-builder'
-import { SuperflyTimelineBuilder } from '../services/superfly-timeline-builder'
+import { SuperTimelineBuilder } from '../services/super-timeline-builder'
```
vushu, line 5: What is a supertimeline?

vushu (approved):

src/business-logic/facades/service-facade.ts
KvelaGorrrrnio: It is a library that models the timeline data structure and resolves it based on the start, end and layer of a timeline objects, along with how they relate to other timeline objects.

src/model/entities/rundown.ts
```diff
@@ -966,7 +972,7 @@ export class Rundown extends BasicRundown {
     this.assertActive(this.stopActivePiecesOnLayers.name)
     const piecesToStop: Piece[] = [
       ...this.getActiveCursor()?.part.getPieces().filter(piece => layers.includes(piece.layer) && !piece.hasEnded(Date.now())) ?? [],
-      ...layers.map(layer => this.infinitePieces.get(layer)).filter((piece): piece is Piece => !!piece)
+      ...layers.map(layer => this.infinitePieces.get(layer)).flatMap(pieceList => pieceList).filter((piece): piece is Piece => !!piece)
```
KvelaGorrrrnio: `Array.flat()` can be used here as well.

src/model/entities/test/piece.spec.ts
```diff
@@ -19,6 +19,14 @@ describe(Piece.name, () => {
       const result: number = testee.getExecutedAt()
       expect(result).toEqual(now)
     })
+
+    it('resets takenOffAirTimestamp', () => {
+      const testee: Piece = new Piece({ takenOffAirTimestamp: 1000 } as PieceInterface)
```
KvelaGorrrrnio, line 24: The `EntityTestFactory` can be used to create a valid `PieceInterface`, making the test less fragile :)

src/business-logic/facades/service-facade.ts
RasmusAlbrektsen: It's a timeline, but... like... super!

src/model/entities/rundown.ts
KvelaGorrrrnio: Okay. I still think that if I wait 6 months and get back to this line, then I would still have no clue what order `newList` have - we should also avoid using types in names :-) I see this is a good case for using a domain object that encapsulates the concept and invariants of the infinite piece layers. The method also seems to have an increased number of responsibilities as it now alters the `layersWithPieces` and calls `piece.takeOffAir`.

One could add the concept of an `InfinitePiecesLayer`. I have tried to draft a class for it in the following code snippet. It will reduce the complexity of the rundown entity; reduce error rate when working with the infinite pieces;  it will be easier to test the logic on a infinite piece layer; and it will be easier to name things. e.g. `newList` could be `newInfinitePiecesLayer` and the condition can be modelled as `piece.id === newestInfinitePiece.id` (where `const newestInfinitePiece = newInfinitePiecesLayer.getNewestInfinitePiece()`).

```typescript
class InfinitePieceLayer {
  private readonly infinitePieces: Piece[]

  constructor(infinitePieces: readonly Piece[]) {
   // We could add invariant assertion here that the list has a minimum length of 1 if that is a desired invariant to enforce.
  // It would allow for `getNewestInfinitePiece` to always return a Piece.
    this.infinitePieces = [...infinitePieces]
  }

  public addInfinitePiece(infinitePiece: Piece[]): InfinitePieceLayer {
    return new InfinitePieceLayer(this.infinitePieces.concat(infinitePiece))
  }

  public getPiecesNoLongerPresent(infinitePieceLayer: InfinitePieceLayer) : readonly Piece[] {
    this.infinitePieces.filter(piece => !infinitePieceLayer.hasInfinitePiece(piece))
  }

  public hasInfinitePiece(infinitePiece: Piece): boolean {
    this.infinitePieces.some(piece => piece.id === infinitePiece.id)
  }

  public getNewestInfinitePiece(): Piece | undefined {
    return this.infinitePieces.at(-1)
  }

  public getOngoingInfinitePieces(currentTimestamp: number, onAirSegment: Segment): readonly Piece[] {
    return this.infinitePieces.filter(piece => !this.isInfinitePieceOutlived(piece, currentTimestamp, onAirSegment))
  }

  private isInfinitePieceOutlived(piece: Piece, currentTimestamp: number, onAirSegment: Segment): boolean {
    if (piece.hasEnded(currentTimestamp)) {
      return true
    }

    switch (piece.pieceLifespan) {
      case PieceLifespan.WITHIN_PART: {
          // Not an infinite, so we don't care about it and just mark it as outlived.
          return true
        }
        // Once taken, the Piece acts like STICKY_UNTIL_RUNDOWN_CHANGE, so it has same rules about being outlived.
      case PieceLifespan.START_SPANNING_SEGMENT_THEN_STICKY_RUNDOWN:
      case PieceLifespan.STICKY_UNTIL_RUNDOWN_CHANGE: {
          // Since we are in the context of a Rundown then the Piece will never be able to leave the Rundown, so the Piece is NOT outlived.
          return false
        }
      case PieceLifespan.STICKY_UNTIL_SEGMENT_CHANGE: {
          // If the Piece belongs to the active Segment, then the Piece is NOT outlived.
          return !onAirSegment.doesPieceBelongToSegment(piece)
        }
      case PieceLifespan.SPANNING_UNTIL_RUNDOWN_END:
      case PieceLifespan.SPANNING_UNTIL_SEGMENT_END: {
          // We always mark SPANNING as outlived because even if it isn't we need to check if there is another SPANNING Piece between this Piece and the active Part.
          return true
        }
      default: {
          ExhaustiveCaseChecker.assertAllCases(piece.pieceLifespan, 'piece lifespan')
        }
      }
  }
}
```

This can be further extended to abstract away the `Map<string, InfinitePiecesLayer>` if we want to enforce invariants between layers or e.g. if we want to be able to remove all pieces from a layer.

I understand if this is too big change for the bug task, but then I think that we should make a tech-debt task for it. And it would be nice if `newList` was renamed to something more descriptive and that `newList[0]` is extracted out into a constant like `newestInfinitePiece`.

src/model/entities/rundown.ts
```diff
@@ -470,26 +464,37 @@ export class Rundown extends BasicRundown {
...
-        )
+      const newList: Piece[] | undefined = layersWithPieces.get(piece.layer)
+      if (!newList) {
+        layersWithPieces.set(piece.layer, [piece])
+        return
+      }
+
+      if (this.getActivePart().getTimings().delayStartOfPiecesDuration === 0) {
+        return
+      }
```
KvelaGorrrrnio, line 486: Can you explain what this checks for and why we want to check for it? :)

src/model/entities/rundown.ts
```diff
@@ -470,26 +464,37 @@ export class Rundown extends BasicRundown {
...
+
+      if (piece.id === newList[0].id) {
+        return
       }
-      layersWithPieces.set(piece.layer, piece)
+      piece.takeOffAir(newList[0].getExecutedAt() + this.getActivePart().getTimings().delayStartOfPiecesDuration)
```
KvelaGorrrrnio: The case we reach here is if the piece is an old infinite piece on the layer or if it isn't on the layer. Should we be able to have duplicate references in an infinite piece layer?

Also could you explain what the duration exactly is and why it is taken off air here?

Edit: I reread the PR description and it partially answered this 😄

src/model/entities/rundown.ts
```diff
@@ -470,26 +464,37 @@ export class Rundown extends BasicRundown {
...
+    this.infinitePieces.forEach((currentInfinitePieces: Piece[], layer: string) => {
+      const newInfinitePiecesList: Piece[] = newInfinitePieces.get(layer) ?? []
+
+      currentInfinitePieces.filter(piece => newInfinitePiecesList.every(p => p.id !== piece.id))
+        .forEach(piece => piece.resetExecution())
+    })
   }
 
-  private addPiecesToLayers(pieces: Piece[], layersWithPieces: Map<string, Piece>): Map<string, Piece> {
+  private addPiecesToLayers(pieces: Piece[], layersWithPieces: Map<string, Piece[]>): Map<string, Piece[]> {
```
KvelaGorrrrnio, line 476: This also takes pieces off air - just a FYI 😄

src/model/entities/rundown.ts
```diff
@@ -516,23 +521,24 @@ export class Rundown extends BasicRundown {
...
+        layersWithPieces.set(baselinePiece.layer, [baselinePiece])
       })
   }
 
-  private setInfinitePieces(layersWithPieces: Map<string, Piece>): void {
+  private setInfinitePieces(layersWithPieces: Map<string, Piece[]>): void {
     this.infinitePieces = new Map()
-    layersWithPieces.forEach((piece: Piece, layer: string) => {
-      if (piece.pieceLifespan === PieceLifespan.WITHIN_PART) {
+    layersWithPieces.forEach((pieceList: Piece[], layer: string) => {
```
KvelaGorrrrnio: We should avoid using types in names :) `pieceList` can be renamed to `pieces` or `piecesOnLayer`.

src/model/entities/rundown.ts
```diff
@@ -888,18 +894,18 @@ export class Rundown extends BasicRundown {
   }
 
   private markInfinitePiecesFromPartUnsynced(partId: string): void {
-    const infinitePiecesFromPart: Piece[] = [...this.infinitePieces.values()].filter(piece => piece.getPartId() === partId)
+    const infinitePiecesFromPart: Piece[] = this.getInfinitePieces().flatMap(list => list).filter(piece => piece.getPartId() === partId)
```
KvelaGorrrrnio: `Array.flat()` can be used here as well.

src/model/entities/rundown.ts
```diff
@@ -412,31 +412,25 @@ export class Rundown extends BasicRundown {
...
 
     this.addBaselinePiecesNotOnLayers(layersWithPieces)
-    this.resetOutlivedInfinitePieces(Array.from(layersWithPieces.values()))
+    this.resetOldInfinitePieces(layersWithPieces)
     this.setInfinitePieces(layersWithPieces)
   }
 
-  private findOldInfinitePiecesNotOnLayers(layers: Set<string>): Piece[] {
-    return Array.from(this.infinitePieces.values()).filter((oldPiece) => !layers.has(oldPiece.layer))
+  private findOldNotOutlivedInfinitePieces(): Piece[] {
```
KvelaGorrrrnio: There is nothing in this method that favors having `Old` in the name - as far as I see :)

KvelaGorrrrnio (commented):
I am not approving as I don't fully understand what is happening and why :)

src/model/entities/rundown.ts
LindvedKrvang: The whole reason for having more than one Infinite Piece per layer is because some Parts wants to delay the start of their Pieces, which means we need to keep the old Pieces.
If the Part doesn't want to delay the start, then there simply is no reason for us to do all the logic/calculations for figuring out if we should keep the old Pieces.

This check prevents us doing unnecessary work.

KvelaGorrrrnio: But can't `piece` in this context be a piece from the on air part?

src/model/entities/test/piece.spec.ts
LindvedKrvang: The `EntityTestFactory` and `EntityMockFactory` are meant to be used for the entities that are _not_ being tested i.e. the data relevant for the test.
They are not meant to be instantiating the actual testee.

src/model/entities/rundown.ts
LindvedKrvang: If we allow duplicate Pieces on the same layer, then we need to update the TimelineBuilder to exclude duplicate Pieces.
If not, then it will try to put two TimelineObjects with the same id on the Timeline at the same Time which will crash TSR.

By `duration` I assume you refer to `delayStartOfPiecesDuration`?
This is the small 'gap' between when the Part is being put on Air untill the Pieces are being put on Air.
It is this gap that is the cause of this bug task. In the old implementation we would simply have "nothing" in this gap.

If we don't take the old Piece off Air then it has no reference to when it should end. We got around that before since we simply discarded the Piece if there was another Piece on the layer.
Since we now need to "extend" the duration of the old Piece into the new Part until the new Piece starts, we need to tell the old Piece to stop (i.e. be taken of Air) at the time the new Piece starts (i.e. the new Piece.getExecutedAt + the delay from the Part).
If we don't tell the old Piece to stop then it will still be an Infinite Piece and just continue carrying on.

src/model/entities/rundown.ts
LindvedKrvang: I don't think we can split does two things unfortunately....
At execution, we are only allowed to have one TimelineObject on the Timeline for a layer at a given time. If we add a new Piece to a layer, then we need to take the old Piece of Air when the new Piece starts.

src/model/entities/rundown.ts
LindvedKrvang: I see your point about the usage of "new". I have updated it to be called `piecesOnLayer`.

I don't have the entire overview of your proposed class.
It would be nice make `rundown.ts` smaller, however, I don't think I am going to venture into that endeavour now.

src/model/entities/rundown.ts
LindvedKrvang: It's been mapped away.

src/model/entities/rundown.ts
LindvedKrvang: Updated

src/model/entities/rundown.ts
LindvedKrvang: I agree. I think a lot of the naming are leftovers from my initial implementation where I still had the original implementation and needed to be able keep track of the differences :) 

It's been updated.

src/model/entities/rundown.ts
LindvedKrvang: The iron has been applied.

src/model/entities/rundown.ts
LindvedKrvang: You are correct. It's been updated.

src/model/entities/test/piece.spec.ts
KvelaGorrrrnio: Which is why I suggested `PieceInterface` and not `Piece` :D

KvelaGorrrrnio (approved):
Looks good. I have created a Kaizen task for extracting infinite piece logic from rundown entity.

PR #365, opened 2025-06-17
SOF-2593: Configure shared eslint setup

KvelaGorrrrnio:
_**Note:** Is based on #363._

Adds configuration that uses the eslint configuration defined in https://github.com/tv2/automation-graphics-eslint-config and applies the rules. The rules are up to debate. This PR aims to setup the shared config, not iron out all the minor details in rules. That should be a continuous process.

_**Note:** There is also a PR for using the shared configuration for the Alba TV 2 server at https://github.com/tv2/alba-tv2-server/pull/129._


RasmusAlbrektsen (approved):

src/blueprints/domain/services/action-factories/tv2-video-clip-action-factory.spec.ts
```diff
@@ -71,10 +71,10 @@ describe(Tv2VideoClipActionFactory.name, () => {
...
 
 function createTestee(params?: {
-  actionManifestMapper?: Tv2ActionManifestMapper,
-  logger?: Logger,
-  videoMixerTimelineObjectFactory?: Tv2VideoMixerTimelineObjectFactory,
-  audioMixerTimelineObjectFactory?: Tv2AudioMixerTimelineObjectFactory,
+  actionManifestMapper?: Tv2ActionManifestMapper
+  logger?: Logger
+  videoMixerTimelineObjectFactory?: Tv2VideoMixerTimelineObjectFactory
+  audioMixerTimelineObjectFactory?: Tv2AudioMixerTimelineObjectFactory
```
LindvedKrvang, line 77: I know this PR is not for discussing the actual eslint rules.
So this comment is just my vote to say I think we should keep the trailing commas in this case :)

LindvedKrvang (approved):

src/action-system/application/dtos/trigger-dto.ts
```diff
@@ -33,7 +33,7 @@ export class ActionTriggerDto extends TriggerDto {
   public readonly actionId: string
   public readonly actionArguments?: string | number
 
-  constructor(trigger: ActionTrigger) {
+  public constructor(trigger: ActionTrigger) {
```
vushu, line 36: Aren't constructors implicitly public?, so it's for explicitness?

vushu (approved):

src/action-system/application/dtos/trigger-dto.ts
KvelaGorrrrnio: They are indeed, but so are all other methods and for them we make an explicit distinction between private and public 😄

src/blueprints/domain/services/action-factories/tv2-video-clip-action-factory.spec.ts
KvelaGorrrrnio: Noted :D I don't have a strong preference for whether we use trailing commas for literal object types. For interfaces I would prefer them not having trailing commas.

PR #366, opened 2025-06-18
Sof 2585/connect to inew gatewawy

LindvedKrvang:
Note: Dependent on https://github.com/tv2/alba-server/pull/363

This PR makes Alba connect to our new INewsGateway. 
To "configure" the gateway we need to set environment variables of which there are currently two:
- One for the host address.
- One as a feature flag, so we can disable the feature in production until the entire ingest flow is done.

This PR also introduces a "ReconnectingWebsocket" and a barebone "HealthStatus system".
- For the "HealthStatus system" I didn't really know how much (or little) we need, so hence the barebone.
- The ReconnectingWebsocket will only emit a new HealthStatus if its status has actually changed.


src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Socket } from '../../../cross-cutting-concerns/infrastructure/interfaces/socket'
+import {
+  IngestHealthStatusEventEmitter
+} from '../../../rundown-ingest/application/interfaces/ingest-health-status-event-emitter'
+import { IngestHealthStatus } from '../../../rundown-ingest/application/enum/ingest-health-status'
+
+const HOST: string = process.env.INEWS_GATEWAY_HOST ?? 'ws://localhost:3008'
+const FEATURE_FLAG: boolean = process.env.DISABLE_INEWS_GATEWAY === 'true'
+
+export class INewsGatewayConnector implements GatewayConnector {
```
KvelaGorrrrnio, line 11: I would like this to use `Inews` instead of `INews` as per our [styleguide](https://tv2cms.atlassian.net/wiki/spaces/SOF/pages/4968120327/Styleguide#Naming-conventions).
It will also make the filename be `inews-gateway-connector` instead of `i-news-gateway-connector`.

src/cross-cutting-concerns/infrastructure/interfaces/socket.ts
```diff
@@ -0,0 +1,10 @@
+export interface Socket {
+  connect(
+    connectionString: string,
+    onConnected: () => void,
+    onData: (data: unknown) => void,
+    onError: () => void,
+    onClose: (isClosedByError: boolean) => void
+  ): void
```
KvelaGorrrrnio: This is a quite complex method signature and it can be hard to remember the order of the methods. I will suggest to move these out to separate methods, so that a consumer of this contract can opt in for the methods that they need. E.g. `subscribeToData(onDataCallback: (data: unknown) => void): void`.

Ideally, the connection string should not be given when connecting as it will exhibit a detail related to the implementation. If the connection url/host-ip should be given, then this contract should be changed to `WebSocket`. We might need to differentiate between `ws://` and `wss://`.

src/cross-cutting-concerns/infrastructure/services/reconnecting-web-socket.ts
```diff
@@ -0,0 +1,81 @@
...
+    this.onConnected = onConnected
+    this.onData = onData
+    this.onError = onError
+    this.onClose = onClose
+
+    this.connectToNewSocket()
+  }
+
+  private connectToNewSocket(): void {
+    this.webSocket = new WebSocket(this.connectionString)
```
KvelaGorrrrnio: This should probably dispose of the old WebSocket connection before overwriting it to avoid phantom connection handles :)

src/cross-cutting-concerns/infrastructure/services/reconnecting-web-socket.ts
```diff
@@ -0,0 +1,81 @@
...
+    })
+
+    this.webSocket.addEventListener('error', (event: ErrorEvent) => {
+      this.logger.data(event).error(`Error from WebSocket listening on: ${this.connectionString}`)
+      this.onError()
+    })
+
+    this.webSocket.addEventListener('close', (event: CloseEvent) => {
+      this.logger.debug(`WebSocket listening on ${this.connectionString} was closed`)
+      this.onClose(event.code !== WEB_SOCKET_NORMAL_CLOSURE_CODE)
```
KvelaGorrrrnio: If the interface is changed to  `WebSocket`, could it makes sense to give the closure code with the onClose instead of a boolean value?

src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
```diff
@@ -0,0 +1,56 @@
...
+        }
+        this.updateHealthStatus(IngestHealthStatus.UNKNOWN)
+      }
+    )
+  }
+
+  private getConnectionString(): string {
+    const queues: string = ''
+    return `${HOST}/?queues=${queues}`
+  }
```
KvelaGorrrrnio, line 34: Suggestion: I assume that the responsibility of this method is to build the connection string. In that case we should probably externalize which queue ids to subscribe to by making it parameter of type `string[]`. Then the caller can simply pass in the list of queues and the URL will be formatted correctly.

```typescript
  private getConnectionString(queueIds: readonly string[]): string {
    return `${HOST}/?queues=${queueIds.join(',')}`
  }
```

RasmusAlbrektsen: I'd prefer this as well :-)

RasmusAlbrektsen (approved):
Good stuff!

src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
```diff
@@ -0,0 +1,56 @@
...
+const FEATURE_FLAG: boolean = process.env.DISABLE_INEWS_GATEWAY === 'true'
+
+export class INewsGatewayConnector implements GatewayConnector {
+  private healthStatus: IngestHealthStatus = IngestHealthStatus.UNKNOWN
+
+  public constructor(private readonly socket: Socket, private readonly ingestHealthStatusEventEmitter: IngestHealthStatusEventEmitter) {
+  }
+
+  public connect(): void {
+    if (FEATURE_FLAG) {
```
vushu: I wasn't sure I understod what `FEATURE_FLAG,` indicates. 
By looking at the implementation I know, maybe consider calling it `DISABLE_FEATURE`. or something alike.

src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
```diff
@@ -0,0 +1,56 @@
...
+      () => {
+        this.updateHealthStatus(IngestHealthStatus.GOOD)
+      },
+      data => this.onData(data),
+      () => {
+        this.updateHealthStatus(IngestHealthStatus.BAD)
+      },
+      (isClosedByError: boolean) => {
+        if (isClosedByError) {
+          return
```
vushu: Consider logging that it was closed by error.

vushu (approved):
lgtm

src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
LindvedKrvang: This current method implementation is just a placeholder until AG-391 does the real implementation.

src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
LindvedKrvang: The error is already logged in `reconnecting-web-socket`, so logging it here would be logging it twice.
The counter-argument to that is that the reason we aren't logging here is because of an implementation detail somewhere else, which also isn't great

src/cross-cutting-concerns/infrastructure/services/reconnecting-web-socket.ts
LindvedKrvang: I guess it could. It depends on if that's how we want the interface?

src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
LindvedKrvang: I'm going to rename it in: https://github.com/tv2/alba-server/pull/369
This is mainly so I can do the renaming at once instead of over several PRs :)

src/cross-cutting-concerns/infrastructure/services/reconnecting-web-socket.ts
LindvedKrvang: Added `this.webSocket?.close()` before creating a new Socket :)

src/cross-cutting-concerns/infrastructure/interfaces/socket.ts
LindvedKrvang: I'm not entirely sure if I have understood your comment correctly. So correct me if I am wrong ;)

I have moved the different "onX" parameters out to serparate methods.
The `connect()` method still takes a "connectionString" or "ipAddress" but the WebSocket prefix is now added within the `ReconnectingWebSocket`.

src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
LindvedKrvang: I have updated the name of the variable

src/tv2-inews-ingest/infrastructure/services/i-news-gateway-connector.ts
LindvedKrvang: It's been renamed in: https://github.com/tv2/alba-server/pull/369/commits/b6c4be66051d353494c13ba46317319c0ab57164

Note: We now have both the `inews` and `iNews` spelling in the codebase as the BlueprintConfiguration spells it as `iNews`

KvelaGorrrrnio: I will create a task for streamlining it :)

src/cross-cutting-concerns/infrastructure/interfaces/socket.ts
KvelaGorrrrnio: Thanks! I am still not entirely sure whether we should hide the fact that it is a WebSocket or not. We can circle back to it at a later point :)

src/cross-cutting-concerns/infrastructure/services/reconnecting-web-socket.ts
KvelaGorrrrnio: As we talked about on teams - a good approach would be to use a custom closure code to check for whether we should reconnect or not.

src/cross-cutting-concerns/infrastructure/services/reconnecting-web-socket.ts
KvelaGorrrrnio: Let us keep the boolean for now as we don't have any use cases that warrants the closure code.

KvelaGorrrrnio (approved):

PR #367, opened 2025-06-19
SOF-2592: Add fitness functions for module dependency management

KvelaGorrrrnio:
This PR setups an initial configuration for [dependency-cruiser](https://www.npmjs.com/package/dependency-cruiser/) for checking inter-module dependency violations.
This PR only adds rules for:
- General dependency management (e.g. disallowing cyclic dependencies, only depending production dependencies, etc.)
- Domain modules may only depend on domain modules.
- Application modules may only depend on application and domain modules.

The PR also solves the low hanging violations, like:
- Add [body-parser](https://www.npmjs.com/package/body-parser) as an explicit dependency.
- Moving migrations out in the root folder.
- Splits up cyclic dependencies for `TimelineObjectMetadata`.
- Splits `Tv2StringHashConverter` (and renames it) into application interface and infrastructure implementation.

_**Note:** Some of the rules have been modified to accommodate our current dependencies and structure, as solving the violations would be too great a task in this PR._

The rest of the rules we need to uphold our architecture will be made in separate PRs to keep the review scope manageable.


vushu (approved):
This is really nice :)

src/blueprints/domain/services/action-factories/tv2-graphics-action-factory.spec.ts
```diff
@@ -254,3 +254,8 @@ function createConfiguredBlueprintConfiguration(): Tv2BlueprintConfiguration {
     }
   })
 }
+function createMockOfStringHashGenerator(): StringHashGenerator {
```
LindvedKrvang, line 258: Missing space between functions :)

LindvedKrvang (approved):

src/blueprints/domain/services/action-factories/tv2-graphics-action-factory.spec.ts
KvelaGorrrrnio: It is now spaced out 😛

PR #370, opened 2025-06-25
SOF-2592: Add fitness for cross cutting concerns module

KvelaGorrrrnio:
>[!Note]
> The PR is based on the changes for #374.

Firstly, sorry for another big PR 🙏 

This PR adds a fitness function for ensuring that the cross-cutting-concerns module does not depend on other modules. The reason being that it is a special module where shared logic and concepts live and thus should not create transient dependencies for all the other modules.

The rule had some violations which led to the changes:

1. **Exceptions**
  - The base class for exceptions are moved to the cross-cutting-concerns-module.
  - The other exceptions has been moved to the modules that own the concept that thet describe.

2. **Mongo entity conversion**
  - The `MongoEntityConverter` has been split up and the individual converter methods are moved to the modules that own the entities that they convert from and to.

3. **Media**
  - All `Media` related code is moved to the sofie-ingest module as it builds on the media concept used in Sofie. We are most likely going to use a separate structure in Alba for asset management, where "playout items" and "media" is handled by the same asset system.

5. **Sofie data listening**
  - The `DataChangeService` and related code is moved to the sofie-ingest module as they all pertain the Sofie database.

6. **Entity test generation**
  - The `EntityTestFactory` is still placed in the rundown-execution module with a todo on splitting it up. This is not done in this PR as it has a scope where a separate PR is necessary to ensure that tests are not broken.
  - The `StatusMessage` part of the factory is moved into a `StatusMessageBuilder` as it was the only entity that the cross-cutting-concerns module is responsible for; and moving the `EntityTestFactory` to the cross-cutting-concerns, would it depend on all other modules.

7. **WebSocket event server uses only one event observer**
  - The `WebSocketEventServer` had ties to all other modules that produces events, so dependency direction is flipped, by introducing the event bus `TypedEventBus` that the individual event services can use to publish to the event server. This means that for most of the event services the observer interface is not necessary and are deleted. There is still one thing missing in that the WebSocketEventServer should be in the application layer. So it needs to be separated into transport and event logic. This will happen in a separate PR.

```mermaid
graph LR

typedEventServer["
  TypedEventServer
  [Service]

  Collects typed events and sends them the the websocket event server.
"]
typedEventEmitter["
  TypedEventEmitter
  [Interface]
"]
typedEventObserver["
  TypedEventObserver
  [Interface]
"]
typedEventBus["
    TypedEventBus
    [Service]
"]
rundownEventService["
    RundownEventService
    [Service]
    Builds and emits rundown events.
"]
statusMessageEventService["
    StatusMessageEventService
    [Service]
    
    Builds and emits status message events.
"]
XyzEventService["
    XyzEventService
    [Service]

    Builds and emits xyz events.
"]

typedEventObserver --> |"implemented by"| typedEventBus
typedEventEmitter --> |"implemented by"| typedEventBus



rundownEventService --> |"emits rundown events to"| typedEventEmitter
statusMessageEventService --> |"emits status message events to"| typedEventEmitter
XyzEventService --> |"emits xyz events to"| typedEventEmitter

typedEventServer --> |"subscribes to typed events from"| typedEventObserver
```


src/cross-cutting-concerns/domain/services/status-message-builder.ts
```diff
@@ -0,0 +1,39 @@
...
+      title: '',
+    })
+  }
+
+  private constructor(statusMessage: StatusMessage) {
+    this.statusMessage = statusMessage
+  }
+
+  public withStatusCode(statusCode: StatusCode): StatusMessageBuilder {
+    return new StatusMessageBuilder({ ...this.statusMessage, statusCode })
```
vushu, line 21: Ah yeah using { ... } to return "self" I see :)

src/cross-cutting-concerns/infrastructure/services/web-socket-event-server.ts
```diff
@@ -86,33 +61,7 @@ export class WebSocketEventServer implements EventServer {
...
-    this.statusMessageEventObserver.subscribeToStatusMessageEvents((statusMessageEvent: StatusMessageEvent) => {
-      webSocket.send(JSON.stringify(statusMessageEvent))
-    })
-    this.deviceEventObserver.subscribeToDeviceEvents((deviceEvent: DeviceEvent) => {
-      webSocket.send(JSON.stringify(deviceEvent))
-    })
-    this.playoutContentEventObserver.subscribeToPlayoutContentEvents((playoutContentEvent: PlayoutContentEvent) => {
-      webSocket.send(JSON.stringify(playoutContentEvent))
-    })
+    this.typedEventObserver.subscribeToTypedEvents((typedEvent: TypedEvent) => webSocket.send(JSON.stringify(typedEvent)))
```
vushu, line 64: Nice that events now can be put into a single `TypedEvent`.

src/rundown-execution/domain/exceptions/on-air-exception.ts
```diff
@@ -1,5 +1,5 @@
-import { ErrorCode } from '../enums/error-code'
-import { Exception } from './exception'
+import { ErrorCode } from '../../../cross-cutting-concerns/domain/enums/error-code'
+import { Exception } from '../../../cross-cutting-concerns/domain/exceptions/exception'
```
vushu, line 2: So these exception are domain meaning in terms of automation and not according to bounded context?

vushu (approved):
Very cool, consider making a highlevel diagram that reflects how the modules interact with eachother using the event bus server. This would be a good supplement for better understanding atleast for me :D

src/rundown-execution/domain/entities/test/entity-mock-factory.ts
```diff
@@ -9,6 +9,7 @@ import { PartTimings } from '../../value-objects/part-timings'
 import { RundownMode } from '../../enums/rundown-mode'
 import { PlayoutContentType } from '../../enums/playout-content-type'
 
+// TODO: Get rid of it.
```
LindvedKrvang, line 12: I disagree. Sometimes you do want mocks over actual instances.

LindvedKrvang (approved):

RasmusAlbrektsen (approved):

src/rundown-execution/domain/entities/test/entity-mock-factory.ts
KvelaGorrrrnio: And I disagree with that :) Entities should not be mocked as they contain domain and business logic which is easily corrupted with mocks, so mocking entities create fragile tests and covers up regressions.

src/rundown-execution/domain/exceptions/on-air-exception.ts
KvelaGorrrrnio: The `OnAirException` is a domain exception related to rundown execution. The `Exception` is a base class for all exceptions through all layers and is placed in the cross-cutting concerns domain module as it then is accessible from all other modules. It can be discussed whether that is the right approach :)
And I would like to remove the `ErrorCode` (in a separate PR) as it is redundant code for the exceptions.

PR #374, opened 2025-06-27
AG-391 Listen for configured iNews queues data

LindvedKrvang:
Note: Dependent on https://github.com/tv2/alba-server/pull/369

We now listen for ingest changes for the configured iNews queues.
When we update the iNewsIngestConfiguration we also update which queues we are listening for to match the configuration.

Also in this PR:
- Fixed bug in ReconnectingWebSocket that would have multiple "loops" opened when reconnecting.
- Renamed `Socket` to `WebSocket` to be more specific in what a consumer can expect from the interface.


RasmusAlbrektsen (approved):

KvelaGorrrrnio (approved):
Looks good :rocket:

vushu (approved):

PR #375, opened 2025-07-01
refactor: Renames ObjectCloner.clone to DeepObjectCloner.deepClone.

KvelaGorrrrnio:
Whether an object is shallowly or deeply cloned is important for the user of the contract, which is why this PR renames `ObjectCloner.clone` to `DeepObjectCloner.deepClone`.
It also enables one class to implement both a shallowClone and deepClone method, when the need arises.


vushu (approved):
Deeply well done!

RasmusAlbrektsen (approved):
That was deep..

PR #376, opened 2025-07-02
refactor: Adds Sofie prefix to the IngestService interface and implementation in sofie-ingest module.

KvelaGorrrrnio:
This is done to avoid confusion in regards to the `IngestService` interfaces that are both defined in the modules tv2-inews-ingest and sofie-ingest.


vushu (approved):
Easy to ingest!

RasmusAlbrektsen (approved):

PR #380, opened 2025-07-10
AG-529: Include infinite timeline object groups in media player assignment.

KvelaGorrrrnio:
By assigning media players to infinite timeline object groups, we are able to have video clips running without being in program. This is required to support keeping a video clip alive during a VOSS session.


RasmusAlbrektsen (approved):

vushu (approved):
Cool

PR #386, opened 2025-07-15
AG-544: Separate package proposal for migrations

KvelaGorrrrnio:
This PR is a proposal for AG-544 for moving the migrations out into a separate package. It makes running migrations independently executable from alba-server, which aids in running migrations when the system is down.
It also simplifies the root package with less dependencies and scripts.

By separating the steps for deployment, we can in case of failure automatically rollback the migration and restart the old version of Alba server.

The proposal publishes the migrations under the same version tag as the alba-server image gets.


vushu:
I see the appeal and it's fine by me, but there are tradeoffs by doing this, namely that we need to access another package to implement migrations, also would you also like to move [tv2 server migrations](https://github.com/tv2/alba-tv2-server/blob/master/src/data-access/migrations/mongo/mongo-migrations/20250219123329-set-is-disabled-to-panel-configurations.ts) into this package? Or would you make another package handling this, since this is tv2 specific migrations.

KvelaGorrrrnio:
@vushu Can you elaborate on what you mean by accessing another package to implement migrations?
My recommendation would be to use a separate package for the alba-tv2-server migrations as it is a separate service. The two services do not share a database, they only share the MongoDB cluster. But they might use different database clusters in the future - who knows.

By having the migrations as a separate package, we avoid references to production code and types in the migrations.

vushu:
@KvelaGorrrrnio I think I misunderstood, I thought you would make a new repo also, as seen this is still within this repo :)

vushu (approved):

RasmusAlbrektsen (approved):

PR #387, opened 2025-07-17
AG-544: WebSocketEventServer sends response from event handlers.

KvelaGorrrrnio:
NTP events were received but no responses were sent back.

The PR also adds loading of the rundown clock in the baseline.


vushu (approved):

vushu (approved):

RasmusAlbrektsen (approved):

src/rundown-execution/domain/services/blueprint-timeline-builder.ts
```diff
@@ -4,15 +4,49 @@ import { Configuration } from '../entities/configuration'
...
 
 export class BlueprintTimelineBuilder implements TimelineBuilder {
   public constructor(
     private readonly timelineBuilder: TimelineBuilder,
     private readonly blueprint: Blueprint
   ) {}
 
-  public getBaseTimeline(): Timeline {
-    return this.timelineBuilder.getBaseTimeline()
+  public getBaseTimeline(configuration: Configuration): Timeline {
```
LindvedKrvang, line 19: This implementation should ideally be in the `superfly-timelinebuilder`

KvelaGorrrrnio: Can you elaborate on why?
I see this as more of a blueprints task as loading the rundown clock from the web client is very much specific to our use cases - not a generic functionality of Alba. It might be that this is more fit to live directly in the blueprints module.

PR #394, opened 2025-07-31
AG-568 - Mix under breaker actions

RasmusAlbrektsen:
Until now we only allowed mixing under planned breakers, and since this is never the use case for how they are employed we needed this to work for our actions as well.


src/blueprints/domain/services/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -441,27 +441,37 @@ export class Tv2TransitionEffectActionFactory extends ActionFactory {
...
+    const cutToSourceTimelineObject: Tv2BlueprintTimelineObject = this.videoMixerTimelineObjectFactory.createProgramTimelineObject(videoMixerInputSource, videoMixerTimelineEnable, { mediaPlayerSession })
+    const transitionToSourceTimelineObjects: Tv2BlueprintTimelineObject[] = this.videoMixerTimelineObjectFactory.createMixTransitionEffectTimelineObjects(sourceInput, 4, { mediaPlayerSession }, start)
+
+    const mixEffectTimelineObjects: Tv2BlueprintTimelineObject[] = this.doesBreakerHaveAlphaForEntireDuration(breaker) ? transitionToSourceTimelineObjects : [cutToSourceTimelineObject]
 
     return [
-      this.videoMixerTimelineObjectFactory.createProgramTimelineObject(videoMixerInputSource, videoMixerTimelineEnable, { mediaPlayerSession }),
-      this.videoMixerTimelineObjectFactory.createCleanFeedTimelineObject(videoMixerInputSource, videoMixerTimelineEnable, { mediaPlayerSession }),
+      ...mixEffectTimelineObjects,
+      this.videoMixerTimelineObjectFactory.createCleanFeedTimelineObject(sourceInput, videoMixerTimelineEnable, { mediaPlayerSession }),
```
LindvedKrvang: Isn't cleanfeed also supposed to show the breaker?

src/blueprints/domain/services/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
```diff
@@ -314,11 +314,11 @@ export class Tv2AtemVideoMixerTimelineObjectFactory implements Tv2VideoMixerTime
     ]
   }
 
-  private createTransitionEffectTimelineObject(layer: Tv2AtemLayer, meContent: AtemMixEffectWithTransition, metadata?: TimelineObjectMetadata): AtemMixEffectTimelineObject {
+  private createTransitionEffectTimelineObject(layer: Tv2AtemLayer, meContent: AtemMixEffectWithTransition, metadata?: TimelineObjectMetadata, start?: number): AtemMixEffectTimelineObject {
```
LindvedKrvang, line 317: This method and the method below (createMixTransitionEffectTimelineObjects) is starting to have too many parameters.
We shouldn't necessarily do something about it now, but 3+ parameters is usally a bad omen.

LindvedKrvang (commented):
Didn't approve yet since I would like an answer to the cleanfeed question.
I believe that we are supposed to show the breaker on the cleanfeed and this looks like it removes the breaker from it?

src/blueprints/domain/services/action-factories/tv2-transition-effect-action-factory.ts
```diff
@@ -441,27 +441,37 @@ export class Tv2TransitionEffectActionFactory extends ActionFactory {
...
+      start,
       duration: this.frameTimeConverter.convertFramesToMilliseconds(breaker.durationInFrames - breaker.startAlpha - breaker.endAlpha) + casparCgPreRollDuration
     }
 
     const videoMixerInputSource: number = breakerActionMetadata.downstreamKeyer.videoMixerFillSource
-    const fileName: string = this.assetPathHelper.joinAssetToFolder(breakerActionMetadata.breaker.fileName, breakerActionMetadata.breakerFolder)
+    const fileName: string = this.assetPathHelper.joinAssetToFolder(breaker.fileName, breakerActionMetadata.breakerFolder)
+
+    const cutToSourceTimelineObject: Tv2BlueprintTimelineObject = this.videoMixerTimelineObjectFactory.createProgramTimelineObject(videoMixerInputSource, videoMixerTimelineEnable, { mediaPlayerSession })
+    const transitionToSourceTimelineObjects: Tv2BlueprintTimelineObject[] = this.videoMixerTimelineObjectFactory.createMixTransitionEffectTimelineObjects(sourceInput, 4, { mediaPlayerSession }, start)
```
KvelaGorrrrnio: Why is `4` frames used here - is it due to the specific breaker?
We could use something like `TEMPORARY_HARDCODED_MIX_UNDER_TRANSITION_DURATION_IN_MS`.

KvelaGorrrrnio (approved):
I'll piggyback on @LindvedKrvang's comment on clean feed - otherwise seems good :) We could consider passing in an Enable instead of start to make it more flexible?

src/blueprints/domain/services/action-factories/tv2-transition-effect-action-factory.ts
RasmusAlbrektsen: Good point, i'll extract to a constant! 
It is the decided upon hardcoded mix frames that we coordinated with BEDE.

src/blueprints/domain/services/action-factories/tv2-transition-effect-action-factory.ts
RasmusAlbrektsen: I've changed it back to use the `videoMixerInputSource` :-)

src/blueprints/domain/services/timeline-object-factories/tv2-atem-video-mixer-timeline-object-factory.ts
RasmusAlbrektsen: Good point!

LindvedKrvang (approved):

PR #395, opened 2025-08-08
AG-571: Wipe to server issues

RasmusAlbrektsen:
This PR changes the program timeline object to have a calculated start rather than 0.


src/rundown-execution/domain/services/superfly-timeline-builder.spec.ts
```diff
@@ -3380,7 +3380,7 @@ describe(SuperflyTimelineBuilder.name, () => {
                 expect(lookaheadTimelineObject.enable.start).toBe(0)
               })
 
-              it('sets the end to be when the active group starts', () => {
+              it('sets the end to be tied to the active group', () => {
```
KvelaGorrrrnio: `tied to` might be a bit vague - I at least am a bit confused about what it means.
Can we be more specific on what it is tied to? :-)

RasmusAlbrektsen: No longer relevant :-O

KvelaGorrrrnio (approved):

LindvedKrvang (approved):

PR #398, opened 2025-08-14
AG-629: Events for Actions, Health statuses and iNews Ingest Configurations are emitted.

KvelaGorrrrnio:
They used the "old" way of emitting events, where the WebSocket server should subscribe to the events (which it didn't). Now they use the new way of publishing events to the `TypedEventBus`.


RasmusAlbrektsen (approved):
I'm not sure why there are changes to the lock-file but otherwise looks good :-D

KvelaGorrrrnio:
> I'm not sure why there are changes to the lock-file but otherwise looks good :-D 

It is probably just patch updates that are semantically compatible with the version targets we have in the package.json.

LindvedKrvang (approved):

PR #407, opened 2025-08-29
AG-322 Allow InfinitePieceGroups to be created for Pieces that belongs to the ActivePart, but currently isn't on it

LindvedKrvang:
If we had an unplanned Infinite Piece on a Part and went back to take the Part again, the Infinite Pieces would disappear while we were on the Part and reappear after we left the Part again.

We have until now assume that if the Infinite Piece belonged to the active Part, that it would also be included in the Pieces of the Active Part. Hence we wouldn't create an Infinite Piece group for it.
In this case, the unplanned Piece would be added to the Infinite Pieces of the Rundown. When we would "re-take" the Part, we first reset it. This means that the unplanned Piece is removed from the Part, so our assumption of the Piece always being on the Part is no longer true. Since the Piece is still included in the Infinite Pieces of the Rundown, the Piece would reappear after we left the Part again. 

This PR checks if an Infinite Piece is not included in the Active Part - if it is not, we will create an Infinite Piece group for it. If it is included in the Part, we don't create an Infinite Piece group since the Piece will be included in the Active Group.

Also in this PR:
- `ActionArguments` are allowed to be optional in the REST endpoint.


src/rundown-execution/domain/services/super-timeline-builder.ts
```diff
@@ -44,7 +44,7 @@ const LOOKAHEAD_PRIORITY: number = 0.1
 const BASELINE_PRIORITY: number = 0
 const LOW_PRIORITY: number = -1
 
-export class SuperflyTimelineBuilder implements TimelineBuilder {
+export class SuperTimelineBuilder implements TimelineBuilder {
```
KvelaGorrrrnio: Why is it renamed? The package is called [superfly-timeline](https://www.npmjs.com/package/superfly-timeline).

KvelaGorrrrnio (commented):
Functionality looks good, but I don't see why `SuperflyTimelineBuilder` is renamed :-)

src/rundown-execution/domain/services/super-timeline-builder.ts
LindvedKrvang: Because it is in fact not called `Superfly-timeline`, but `supertimeline`. https://github.com/SuperFlyTV/supertimeline

This is is something that SuperFly themselves has pointed out to me long ago.

LindvedKrvang:
> Functionality looks good, but I don't see why `SuperflyTimelineBuilder` is renamed :-)

To accurately describe what it is doing

KvelaGorrrrnio (approved):

KvelaGorrrrnio (approved):

rana-tv2 (approved):

PR #420, opened 2025-09-05
AG-705: [Band-aid] Allow infinite pieces with `executedAt` set to 0

KvelaGorrrrnio:
This PR comments out the check for whether we include infinite pieces with `executedAt: 0`. It does not fix the root cause, but avoids the user getting stuck in a flow.
A follow-up on this should be performed to fix the root-cause and re-enable the commented out check. The band-aid solution is chosen due to a restricted timeframe for getting Gallery F ready for producing "Skyggesiden".


rana-tv2 (approved):

LindvedKrvang (approved):
We should do a thourough testing to ensure that our theory of this not being too impactfull is true.

PR #421, opened 2025-09-08
fix: Adds Video MixMinus timeline class to program timeline objects on remote and split screen action pieces.

KvelaGorrrrnio:
Adds the `overridden_on_mix_minus` class to remote and DVE timeline objects.
We should at a later point in time look into if we can do this without classes.


LindvedKrvang (approved):
Maybe move the string value out in an Enum, so it is not a magic string

rana-tv2 (approved):

PR #424, opened 2025-09-15
AG-319 Update persisting Audio logic

LindvedKrvang:
This PR changes how the logic for persisted audio works.

It used to be, that in order to persist audio, the previous Part would need to "want to persist audio" and the new Part would need to "accept persisted audio".
If the audio had to keep being persisted, then subsequent Parts would all need to "want to be persist audio" and "accept audio".
This meant if LIVE 1 wanted to persist audio and LIVE 2 and LIVE 3 only accepts audio, then going LIVE 1 -> LIVE 2 -> LIVE 3 would only persist the audio from LIVE 1 on the LIVE 2, not on the LIVE 3.

This PR changes the logic so that audio will keep being persisted until a Part is taken that does not "accept persisted audio" regardless of if intermediate Parts "wants to persist" their own audio.
So in the example above, the audio from LIVE 1 would still be persisted when LIVE 3 is taken, but there would be no audio from LIVE 2.

Also in this PR:
- Removed a lot of old logic about finding the persisted audio. These were legacy from how Blueprints originally did the logic.
- Removed "legacy" interface attributes that was only used by old Blueprints.


src/blueprints/domain/services/tv2-end-state-for-part-service.ts
```diff
@@ -60,32 +46,16 @@ export class Tv2EndStateForPartService implements BlueprintGetEndStateForPart {
     return endState
   }
 
-  private getEmptyTv2RundownPersistentState(): Tv2RundownPersistentState {
-    return {
-      activeMediaPlayerSessions: [],
-      isNewSegment: false
-    }
-  }
+  private getAudioLayersToPePersisted(part: Part, time: number, previousPart?: Part): string[] {
```
KvelaGorrrrnio: I assume that `PePersisted` is a typo :D

KvelaGorrrrnio (approved):
Good job :rocket:

src/blueprints/domain/services/tv2-end-state-for-part-service.ts
LindvedKrvang: image

LindvedKrvang: Fixed

rana-tv2 (approved):

PR #426, opened 2025-09-24
AG-711 Prune outlived infinite Pieces when setting a new next part

LindvedKrvang:
"Prune" outlived InfinitePieces when we set a new Part as Next.


src/rundown-execution/domain/entities/piece.ts
```diff
@@ -115,11 +112,23 @@ export class Piece {
...
+      // Infinite Pieces might still be OnAir when their Part is reset, so we can't reset them here.
+      return
+    }
+
     this.executedAt = 0
     this.takenOffAirTimestamp = 0
   }
 
+  private isActiveInfinitePiece(): boolean {
+    if (this.pieceLifespan === PieceLifespan.WITHIN_PART) {
```
RasmusAlbrektsen: I'm probably just confused, but why is it that an infinite piece cannot be active when it has a lifespan of `WITHIN_PART`?

RasmusAlbrektsen (approved):
Looks good, just one question :-)

src/rundown-execution/domain/entities/test/rundown.spec.ts
```diff
@@ -722,7 +722,7 @@ describe(Rundown.name, () => {
           expect(layersToIgnore.has(layer)).toBeTruthy()
         })
 
-        it('sets executedAt to zero for the Piece no longer being an infinite', () => {
+        it('takes the Piece off Air for the Piece no longer being an infinite', () => {
```
KvelaGorrrrnio, line 725: I am not sure I understand this. Are the first and second `Piece` the same piece or different pieces?

src/rundown-execution/domain/entities/piece.ts
```diff
@@ -115,11 +112,23 @@ export class Piece {
...
+      return false
+    }
+    return this.isStarted() && !this.hasEnded(Date.now())
```
KvelaGorrrrnio: You can use `isInfinitePiece` so that this becomes:
```typescript
return this.isInfinitePiece() && this.isStarted && !this.hasEnded(Date.now)
```
Note: I would prefer that we don't use `Date.now` in our domain objects, as it creates temporal coupling to the logic, which conflicts with [principle 9 of our style guide](https://tv2cms.atlassian.net/wiki/spaces/AG/pages/5436080156/Code+Style+Guide#9.-Separate-actions-from-calculations).

src/rundown-execution/domain/entities/piece.ts
```diff
@@ -209,8 +218,12 @@ export class Piece {
     this.insertedTimelineObjects.push(...timelineObjects)
   }
 
+  public isStarted(): boolean {
+    return this.executedAt > 0
+  }
```
KvelaGorrrrnio: `isStarted` holds some ambiguity. A stopped piece will from the semantics of the name might not be true, but the implementation will return true.
It can be as is for now - but I think we should look at terms and concepts around state and transitions for our entities :)

KvelaGorrrrnio (approved):
Nice work :rocket:

src/rundown-execution/domain/entities/piece.ts
LindvedKrvang: Because if it has a lifespan of `WITHIN_PART`, then it is not an infinite Piece ;)

rana-tv2 (approved):

src/rundown-execution/domain/entities/piece.ts
LindvedKrvang: Do you want it to be `hasStarted()` instead? For a Piece to be stopped, it would have needed to have been started

PR #433, opened 2025-10-14
AG-747 Looking for MediaPlayerSessions in use now also check the infinite groups

LindvedKrvang:
When looking for MediaPlayerSessions in use, we would only look in the active group of the Timeline.
With the introduction of VOSS we now have MediaPlayerSessions running from infinite Pieces.

Since we didn't check the infinite groups of the Timeline we had an egdecase where a MediaPlayerSession in use would switch MediaPlayer if it wasn't already assigned to the "first" MediaPlayer.
This is because we wouldn't find any MediaPlayerSessions in use, but we could see we would need a MediaPlayer, so we would assign one. The logic for assigning MediaPlayers is simply to always assign the "first" MediaPlayer if available (the order of MediaPlayers only change if the configuration is updated).

This PR now also check the infinite groups to see if they have any MediaPlayerSessions in use, so we don't reassign the MediaPlayer if that is the case.


rana-tv2 (approved):

rana-tv2:
nice work  👍

RasmusAlbrektsen (approved):
Good catch!

src/blueprints/domain/services/tv2-on-timeline-generate-service.spec.ts
```diff
@@ -453,6 +453,38 @@ describe(Tv2OnTimelineGenerateService.name, () => {
             expect(result.activeMediaPlayerSessions[1].mediaPlayer.id).toBe('1')
           })
         })
+
+        describe('there is an infinite group that wants to continue using the MediaPlayer', () => {
+          const mediaPlayerIds: string[] = ['1', '2']
+          const configuration: Configuration = {} as Configuration
```
KvelaGorrrrnio, line 459: We have a `ConfigurationTestFactory` class that can be used for creating a valid configuration:
```typescript
const configuration: Configuration = ConfigurationTestFactory.createConfiguration()
```

KvelaGorrrrnio (approved):

PR #435, opened 2025-10-27
AG-797: Use track 3+4 for audio bed actions

KvelaGorrrrnio:
CasparCG 2.2+ changes how we reroute audio tracks. It is changed from a configuration-based approach to a command-based approach.
This works with planned audio beds - but support for the command-based wasn't implemented for audio bed actions.

## Change description

This change adds the `useAudioFilterSyntax` from Sofie, that is used in Sofie for determining whether we should use the configuration-based or command-based approach. It is the command-based approach that uses the audio filter syntax.

## Checklist

- [x] The changes are documented.
- [x] Core logic is covered by tests.


rana-tv2 (approved):
Looks good. I am not sure where the `AUDIO_BED_AUDIO_FILTER` string literal value come from, would like to hear if there is any documentation that you can refer me to 😄

KvelaGorrrrnio:
> Looks good. I am not sure where the `AUDIO_BED_AUDIO_FILTER` string literal value come from, would like to hear if there is any documentation that you can refer me to 😄

@rana-tv2 The audio filter syntax is FFmpeg arguments that CasparCG passes through to FFmpeg. The `pan` filter that we use is for mixing audio channels/tracks: https://ffmpeg.org/ffmpeg-filters.html#pan-1

PR #437, opened 2025-10-30
AG-799: Increase log detail for rundown induced exceptions

KvelaGorrrrnio:
We have observed in production inconsistency in the rundown state, that rendered the producer unable to produce the show without resetting the rundown. When debugging, we were unable to determine what had happened.
This change tries to provide better and more details in the rundown entity aggregate structure.

## Change description

The most important change is 3e67d947265798412d880d8da7168e139a861ba7, where we log the part names and ids when we can't find the next off air part.

The rest of the changes are normalization of wording and detail (having both names and ids for entities in exception messages and using lowercase entity names).

## Checklist

- [x] The changes are documented.
- [x] Core logic is covered by tests.


src/rundown-execution/domain/entities/segment.ts
```diff
@@ -101,7 +101,7 @@ export class Segment {
     if (!this.invalidity) {
       return
     }
-    throw new InvalidSegmentException(`Unable to do "${operationName}", since segment "${this.name}" is invalid.`)
+    throw new InvalidSegmentException(`Unable to do "${operationName}", since segment "${this.name}" with id is invalid.`)
```
rana-tv2:
```suggestion
    throw new InvalidSegmentException(`Unable to do "${operationName}", since segment with id:  "${this.name}"  is invalid.`)
```

src/rundown-execution/domain/entities/segment.ts
```diff
@@ -150,19 +150,19 @@ export class Segment {
   public findNextPartNotOnAir(fromPart: Part): Part {
     const fromPartIndex: number = this.parts.findIndex(part => part.id === fromPart.id)
     if (fromPartIndex === -1) {
-      throw new NotFoundException('Part does not exist in Segment')
+      throw new NotFoundException(`Part '${fromPart.name}' with id '${fromPart.id}' does not exist in segment '${this.name}' with id '${this.id}'. Segment contains the following parts: ${this.parts.map(part => `${part.name} (${part.id})`).join(', ')}.`)
     }
     const nextPart: Part | undefined = this.parts.slice(fromPartIndex + 1).find(part => !part.invalidity && !part.isOnAir())
     if (!nextPart) {
-      throw new LastPartInSegmentException(`The part "${fromPart.name}" with id "${fromPart.id}" is the last part in the segment "${this.name}" with id "${this.id}".`)
+      throw new LastPartInSegmentException(`The part '${fromPart.name}' with id "${fromPart.id}" is the last part in the segment "${this.name}" with id "${this.id}".`)
```
rana-tv2: why is it only `${formPart.name}` is changed here and the rest are still in double quotes?

rana-tv2 (approved):
Looks good :)

src/rundown-execution/domain/entities/segment.ts
KvelaGorrrrnio: Nicely spotted :) I have updated it in 49072188.

src/rundown-execution/domain/entities/segment.ts
KvelaGorrrrnio: I had a blind spot for this line - thanks for pointing it out. I have changed it in c1449d1a.

RasmusAlbrektsen (approved):

PR #445, opened 2025-11-18
AG-745 PoC Mix to Jingle

LindvedKrvang:
This PR changes the `PieceLayer` of our TransitionPieces from our Actions to not be on the `JINGLE` layer.
They are now instead on their own `TRANSITION` layer.

The reason for this is if a producer mistankenly tried to Mix to a Jingle, the Jingle would disappear. This is because the Jingle and Transition Piece shared the same PieceLayer.
By having a separate PieceLayer for our Transition Pieces, they can now co-exist.

We still can't do an actual Mix to the Jingle, but that is due to how our setup actually plays jingles.
This PR just ensures that no "black frames" suddenly appear if the producer accidentially tries to mix to the jingle.


RasmusAlbrektsen (approved):

KvelaGorrrrnio (commented):
The proposal looks fine :) I would very much appreciate if this was documented at https://tv2cms.atlassian.net/wiki/spaces/AG/pages/5465899044/Mix+to+Jingle-issue, such that we have it for reference if something is affected by the change.
The pros and cons for each option you have looked at (and that was previously discussed) should be documented along with the chosen option and why.

In addition, the change in layer should also be applied to blueprints for consistency.

LindvedKrvang:
> The proposal looks fine :) I would very much appreciate if this was documented at https://tv2cms.atlassian.net/wiki/spaces/AG/pages/5465899044/Mix+to+Jingle-issue, such that we have it for reference if something is affected by the change. The pros and cons for each option you have looked at (and that was previously discussed) should be documented along with the chosen option and why.
> 
> In addition, the change in layer should also be applied to blueprints for consistency.

@KvelaGorrrrnio So what's stopping this PR from being merged? ;)

rana-tv2 (approved):

PR #447, opened 2025-11-20
Fix: Use native fetch API over got

KvelaGorrrrnio:
## Context

The Docker image deployed, after mutliple dependabot PR merges (including got upgrade) merged into it, is unable to start due to ESM import conflicts for Got. The initial approach was to downgrade to the previous got version - but did not work. 
To get a working application, it was decided to try to upgrade node and use the native fetch API.

## Change description

Node version is bumped to 24.11 (the current active version); `got` and `@types/got` are removed; and the native fetch API is used for the HttpService implementation.

## Checklist

- [X] The changes are documented.
- [X] Core logic is covered by tests.


LindvedKrvang (approved):
Amazing to see Dependency Inversion in action :D

rana-tv2 (approved):

RasmusAlbrektsen (approved):

